PR #108, opened 2018-06-11
node::stress_object: ignore WARNING messages from cassandra-stress

vladzcloudius:
Ignore messages printed by cassandra-stress to stderr that start with
"WARNING:". These messages are not critical (e.g. "WARNING: An illegal reflective access operation has occurred") and the tool may complete the test just fine despite the problem that caused this kind
of a message.


bentsi:
this PR is more than 6 months old and no consensus reached. Let's reopen if we need this functionality.

PR #116, opened 2018-10-31
Comment commitlog_sync_batch_window_in_ms parameter configuration for  Cassandra cluster

juliayakovlev:
Comment commitlog_sync_batch_window_in_ms parameter configuration for  Cassandra cluster because of Cassandra fails on None value. It expects "NaN".

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -431,7 +431,9 @@ class Cluster(object):
             else:
                 self._config_options["commitlog_sync"] = "periodic"
                 self._config_options["commitlog_sync_period_in_ms"] = 10000
-                self._config_options["commitlog_sync_batch_window_in_ms"] = None
+                # Commented because of Cassandra fails on None value. It expects "NaN".
+                # self._config_options["commitlog_sync_batch_window_in_ms"] = None
+
 
         self._update_config()
         for node in list(self.nodes.values()):


ccmlib/cluster.py
```diff
@@ -431,7 +431,9 @@ def set_configuration_options(self, values=None, batch_commitlog=None):
             else:
                 self._config_options["commitlog_sync"] = "periodic"
                 self._config_options["commitlog_sync_period_in_ms"] = 10000
-                self._config_options["commitlog_sync_batch_window_in_ms"] = None
+                # Commented because of Cassandra fails on None value. It expects "NaN".
+                # self._config_options["commitlog_sync_batch_window_in_ms"] = None
+
```
roydahan: @slivne, the mc tests fail due to this missing PR.
I'm not sure about this code and if it affects scylla or not.
Can you please take a look?

slivne: @juliayakovlev / @roydahan  while your change is correct - its the first case in which set_configuration is passed with batch_commitlog False  (even for cassandra - they still have this bug upstream) - why are we changing the batch_commitlog in set_configuration_options ?

roydahan: we have no idea...

roydahan: In this case, I'll merge this change and backport it.

roydahan:
backported to 3.0

PR #117, opened 2018-11-21
Added class ScyllaREST for REST API operations, and ScyllaNode.api() method

rafie:
Usage example:
resp = node.api().get('column_family/toppartitions/ks:t1', {'duration': '1000'})
# use resp.text or resp.json()

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -19,6 +19,7 @@ from six import print_
 from ccmlib import common
 from ccmlib.node import Node
 from ccmlib.node import NodeError
+from ccmlib.scylla_rest import ScyllaREST
 
 
 def wait_for(func, timeout, first=0.0, step=1.0, text=None):
@@ -796,3 +797,7 @@ class ScyllaNode(Node):
     def flush(self):
         self.nodetool("flush")
         self._wait_no_pending_flushes()
+
+    def api(self):
+        ip = self.network_interfaces['binary'][0]
+        return ScyllaREST(ip)
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -73,6 +73,7 @@ class ScyllaNode(Node):
         self._process_jmx_waiter = None
         self._process_scylla = None
         self._process_scylla_waiter = None
+        self._rest_api = ScyllaREST(self.network_interfaces['binary'][0])
 
     def get_install_cassandra_root(self):
         return os.path.join(self.get_install_dir(), 'resources', 'cassandra')
@@ -798,6 +799,6 @@ class ScyllaNode(Node):
         self.nodetool("flush")
         self._wait_no_pending_flushes()
 
-    def api(self):
-        ip = self.network_interfaces['binary'][0]
-        return ScyllaREST(ip)
+    @property
+    def rest_api(self):
+        return self._rest_api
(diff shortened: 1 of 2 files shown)


ccmlib/scylla_rest.py
```diff
@@ -0,0 +1,27 @@
+
+import json
+import requests
+
+class ScyllaREST():
+    def __init__(self, ip = '127.0.0.1', port = 10000):
+        self.ip = ip
```
roydahan: @slivne, will it work with the way we run tests in parallel?

ccmlib/scylla_rest.py
```diff
@@ -0,0 +1,27 @@
+
+import json
+import requests
+
+class ScyllaREST():
+    def __init__(self, ip = '127.0.0.1', port = 10000):
```
bentsi: not PEP8, should be:
```
def __init__(self, ip='127.0.0.1', port=10000)
```

ccmlib/scylla_rest.py
```diff
@@ -0,0 +1,27 @@
...
+        self.ip = ip
+        self.port = port
+        self.url = 'http://{}:{}'.format(ip, port)
+
+    def post(self, item, params):
+        headers = {
```
bentsi: It looks like the headers in `get` and `post` methods are same, so should be defined in the constructor

ccmlib/scylla_node.py
```diff
@@ -796,3 +797,7 @@ def no_pending_flushes():
     def flush(self):
         self.nodetool("flush")
         self._wait_no_pending_flushes()
+
+    def api(self):
```
bentsi: better to make this property with `@property` decorator and rename it to rest_api.
Usage example: ```node.rest_api.get(...)```
In addition, the instance of the ScyllaREST object will be created on each call to the method, so need to prevent it

ccmlib/scylla_rest.py
slivne: if we don't change the ip it won't ...
- I think the best would be not to have the default at all - so this will fail if not set.
- the other thing that we should do is not expose this class and make it internal in my view - its better if the calls come through a scylla_node and not through a generic class that is not clear how its connected to a cluster - think about nodetool we do not have have a generic nodetool class - we have nodetool affiliated with a cluster which means run on all nodes in the cluster and nodetool affiliated with a node that means run on this node - the same goes for the rest api.

so in my view having an external file and not having it inside scylla_node is bad ....

rafie: Having a standalone class simply reflects the fact that one can perform a REST API operation without a scylla_node context (which it true). Same logic applies to the default IP address - it is reasonable to try 127.0.0.1 if nothing else is known.
Of course, when scylla_node embeds ScyllaREST, it will make up its own bed.

rafie:
Well, I guess that the problem arises from the fact that I'm trying to reason about an arbitrary scylla node, and CCM's scylla_node is a specific kind of node is a CCM-specific cluster (although the scylla REST API is the same for all nodes). Since there's no natural place for generic scylla node automation, I'm abusing CCM as a placeholder for such logic.
It looks obvious that in the presence of a CCM cluster (as in dtest) one should consume REST API via scylla_node.rest_api(). However, if one just launches scylla (e.g during development), a ScyllaREST class can come in handy.
It is possible, of course, to replace ScyllaREST class with scylla_node methods, but then the latter use case is lost.

bentsi:
@slivne @roydahan do we still need this?

PR #118, opened 2018-12-26
node or cluster stop: make timeout configurable

nyh:
Currently, node.stop(), and cluster.stop() which uses it, are hard-coded
to allow Scylla 127 seconds (2^7-1), just a bit over two minutes, to
shutdown before failing. In some tests, see for example
https://github.com/scylladb/scylla/issues/4019, we know that it will
take more than that for the shutdown to complete.

So this patch makes the waiting time configurable, via an optional parameter
"wait_iterations" defaulting to 7. The timeout will be 2^wait_iterations-1
seconds.

Signed-off-by: Nadav Har'El <nyh@scylladb.com>

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -347,10 +347,10 @@ class Cluster(object):
 
         return started
 
-    def stop(self, wait=True, gently=True, wait_other_notice=False):
+    def stop(self, wait=True, gently=True, wait_other_notice=False, wait_iterations=7):
         not_running = []
         for node in list(self.nodes.values()):
-            if not node.stop(wait, gently=gently, wait_other_notice=wait_other_notice):
+            if not node.stop(wait, gently=gently, wait_other_notice=wait_other_notice, wait_iterations=wait_iterations):
                 not_running.append(node)
         return not_running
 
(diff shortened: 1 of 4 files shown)


ccmlib/cluster.py
```diff
@@ -347,10 +347,10 @@ def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False, wait_
 
         return started
 
-    def stop(self, wait=True, gently=True, wait_other_notice=False):
+    def stop(self, wait=True, gently=True, wait_other_notice=False, wait_iterations=7):
```
slivne: I think its better to have a time limit then iterations - its hard to convert one to the other

nyh: The comment explains the conversion (2^wait_iterations-1 seconds). 

I wanted my change to change absolutely nothing in the existing behavior of the function (which did exactly 7 iterations in successively doubled sleeps up to 127 seconds) when this option is set to the default.

But there's something else I can do that will avoid any change to the default case: ask for the wait_seconds, not iterations, and always do 7 iterations, i.e., the sleep will start with 1.0*wait_seconds/127 instead of 0. Would you prefer this?

nyh:
Will change the API and resubmit.

PR #119, opened 2018-12-26
node or cluster stop: make timeout configurable

nyh:
Currently, node.stop(), and cluster.stop() which uses it, are hard-coded
to allow Scylla 127 seconds (=2^7-1), just a bit over two minutes, to
shutdown before failing. In some tests, see for example
https://github.com/scylladb/scylla/issues/4019, we know that we need more
than that time for the shutdown to complete.

So this patch makes the waiting time configurable, via an optional parameter
"wait_seconds" which defaults to the previously hard-coded 127 seconds and
can now be set to any number of seconds.

Signed-off-by: Nadav Har'El <nyh@scylladb.com>

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -603,11 +603,14 @@ class Node(object):
 
         return process
 
-    def stop(self, wait=True, wait_other_notice=False, gently=True):
+    def stop(self, wait=True, wait_other_notice=False, gently=True, wait_seconds=127):
         """
         Stop the node.
           - wait: if True (the default), wait for the Cassandra process to be
             really dead. Otherwise return after having sent the kill signal.
+            stop() will wait up to wait_seconds, by default 127 seconds, for
+            the Cassandra process to die. After this wait, it will throw an
+            exception stating it couldn't stop the node.
           - wait_other_notice: return only when the other live nodes of the
             cluster have marked this node has dead.
           - gently: Let Cassandra clean up and shut down properly. Otherwise do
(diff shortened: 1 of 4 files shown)


nyh:
BTW, this needs backport to branch 3.0 (if we have such a branch for ccm)

ccmlib/node.py
```diff
@@ -651,7 +654,10 @@ def stop(self, wait=True, wait_other_notice=False, gently=True):
 
             still_running = self.is_running()
             if still_running and wait:
-                wait_time_sec = 1
+                # The sum of 7 sleeps starting at 1 and doubling each time
+                # is 2**7-1 (=127). So to sleep an arbitrary wait_seconds
+                # we need the first sleep to be wait_seconds/(2**7-1).
+                wait_time_sec = wait_seconds/(2**7-1.0)
```
slivne: what happens if wait_seconds passed is 10 ...

will this be rounded down to 0 ... and then there will be no wait ?

slivne: what happens when its 150 seconds ... will it also work ? or will it be rounded down wait_time_sec to 1 and then we will wait only for 127 seconds ?

nyh: No, no rounding. Because of the "1.0" I used in the formula, the entire calculation is done in floating point. You can end up with wait_time_sec = 0.1 or 2.3, or whatever. This floating-pointed-ness will "infect" all the other calculations in the code, and sleep of floating point also works correctly.

So I think this code will work correctly for any number, not just a power of two.

nyh: An example:
```
>>> wait_seconds=10
>>> wait_time_sec = wait_seconds/(2**7-1.0)
>>> print(wait_time_sec)
0.0787401574803
```

slivne: my concern is with the actual python time.sleep function will it support the fractions ?

time.sleep(wait_time_sec)

if it does then its ok :)

nyh: It does. I checked. 
By the way, even it didn't, it doesn't really matter - it could only break the first, tiny, sleeps. The last sleeps are anyway very long and the fractional part in them is not important.

nyh:
ping

nyh:
Oops, sorry, I missed that it was already committed (didn't refresh automatically...)

PR #125, opened 2019-01-22
Small fixes: remove file during teardown or search for the errors just if the file/folder exists

juliayakovlev:
We need this fix for the case, when Cassadra cluster is created within the test after the Scylla cluster and KEEP_DIR=False

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -257,15 +257,16 @@ class Cluster(object):
     def remove_dir_with_retry(self, path):
         tries = 0
         removed = False
-        while removed is False:
-            try:
-                common.rmdirs(path)
-                removed = True
-            except Exception as e:
-                tries = tries + 1
-                time.sleep(.1)
-                if tries == 5:
-                    raise e
+        if os.path.exists(path):
+            while not removed:
+                try:
+                    common.rmdirs(path)
+                    removed = True
+                except:
+                    tries = tries + 1
+                    time.sleep(.1)
+                    if tries == 5:
+                        raise
 
     def clear(self):
         self.stop()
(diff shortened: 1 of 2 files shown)


ccmlib/cluster.py
```diff
@@ -257,15 +257,16 @@ def remove(self, node=None, wait_other_notice=False):
...
-            try:
-                common.rmdirs(path)
-                removed = True
-            except Exception as e:
-                tries = tries + 1
-                time.sleep(.1)
-                if tries == 5:
-                    raise e
+        if os.path.exists(path):
+            while removed is False:
```
bentsi: `while not removed:`

bentsi: and BTW why not to use `retry` decorator?

ccmlib/node.py
```diff
@@ -332,7 +332,11 @@ def grep_log_for_errors(self, filename='system.log', distinct_errors=False, sear
         if search_str:
             search_str = search_str if case_sensitive else search_str.lower()
 
-        with open(os.path.join(self.get_path(), 'logs', filename)) as f:
+        log_file = os.path.join(self.get_path(), 'logs', filename)
+        if not os.path.exists(log_file):
+            return []
```
bentsi: if there is no log file it is a problem, no?

ccmlib/cluster.py
```diff
@@ -257,15 +257,16 @@ def remove(self, node=None, wait_other_notice=False):
...
+                try:
```
bentsi: try:
....
except:
    do_something
    if ....:
         raise

ccmlib/node.py
juliayakovlev: As I explain above it may happens when we remove one cluster and create another one and parameter KEEP_DIR=False

juliayakovlev:
@bentsi  Please, review

bentsi:
@juliayakovlev can reopen this PR to merge to `next` branch instead of master?

PR #126, opened 2019-01-27
node: use overprovisioned flag if no cpuset

avikivity:
Allow different nodes to share cpus without a performance penalty. This allows increasing utilization when running mulitple test clusters in parallel.

Commit 6a3225e28701cff3a0b098212baddfb1761e04bd reverted it, but forgot to explain why. If there are problems, we should fix them.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -354,11 +354,7 @@ class ScyllaNode(Node):
         if '--collectd' not in args:
             args += ['--collectd', '0']
         if '--cpuset' not in args:
-            # Code above guarantees --smp is always in the args list
-            smp = int(args[args.index('--smp') + 1])
-            id = int(data['listen_address'].split('.')[3]) - 1
-            cpuset = self.cpuset(id, smp, self.cluster.id)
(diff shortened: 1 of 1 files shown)


avikivity:
ping @slivne @roydahan

roydahan:
@slivne any chance you remember why you did it?

slivne:
search for "So all the "Can't send migration request ..." are all gone with the revert of the ccm using overprovisioned change" on dev-int and you should find an email thread about it.

Having said that - I don't understand why it would be an issue we should merge this and figure out why it makes a difference and fix that.

roydahan:
Are you sure you want to do it to Benny? :)
While we are trying to stabilize master, introduce new issues?

slivne:
Well ... 
- now that we have ccm in the pipleline if one of your stable dtests fails then this will not go in
- if hell will break loose we will check this tomorrow and try to understand
- previous attempt to get this in was at the end of 2.3 cycle,  now we are not at the end of 3.1 cycle so I think its ok.

roydahan:
Ok.
@bhalevy FYI

PR #127, opened 2019-02-05
ccmlib: make ccm aware of dev build mode

psarna:
This allows creating ccm from build/dev directory.


psarna:
Turns out the support was already added by Benny and merged, so this one became a duplicate :) Closing.

PR #129, opened 2019-02-10
Support cassandra-stress new version (3.0) parameters

fruch:


ccmlib/node.py
```diff
@@ -1083,6 +1083,24 @@ def stress(self, stress_options=None, capture_output=False, **kwargs):
...
+
+        # handle cassandra-stress >= 3.11 parameters
+        has_limit = any(['limit=' in o for o in stress_options])
+        has_throttle = any(['throttle=' in o for o in stress_options])
+        if has_limit or has_throttle:
+            args = [stress, 'version']
+            p = subprocess.Popen(args, cwd=common.parse_path(stress),
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+                                 **kwargs)
+            stdout, stderr = p.communicate()
```
bentsi: if you don't use `stdout` better to replace by `_`

juliayakovlev:
Can you add some explanation about this, please?

bentsi:
@juliayakovlev in cassandra-stress >3.11 `limit` argument was renamed to `throttle`, so we need handle this situations since most of the tests are written using oldest version of c-s.

ccmlib/node.py
```diff
@@ -1083,6 +1083,24 @@ def stress(self, stress_options=None, capture_output=False, **kwargs):
...
+
+            if p.returncode == 1 and has_throttle:
```
juliayakovlev: Why that? Does the c-s > 3.11 have no "version" argument?

bentsi: exactly opposite, c-s> 3.11 has *version* argument and  c-s<3.11 doesn't have *version* argument

PR #130, opened 2019-02-13
Small fixes: remove file during teardown or search for the errors just if the file/folder exists

juliayakovlev:


bentsi:
This the same as #125 but merging in `next` branch

PR #136, opened 2019-02-21
Collecting stress stdout/stderr and printing when error

bentsi:
Addressing issue: https://github.com/scylladb/scylla/issues/4259
Collecting stdout/stderr of the stress process and printing them when we have an error.
Current stress method behavior isn't changed.


bhalevy:
Makes sense.  thanks

PR #140, opened 2019-03-04
setup.py: also install resources/bin/*.sh

denesb:
Otherwise scylla-dtest will not be able to use the installed ccmlib.


denesb:
Ping.

roydahan:
@denesb Can you please elaborate on the issue it fixes?
Did you or someone test it doesn't break how we run it in Jenkins?

denesb:
Last I tried I couldn't reproduce the issue anymore. The issue was that when installing `scylla-ccm` into a python virtualenv, the dtests wouldn't work unless I made this change. I will close this now and will reopen if I bump into it again.

PR #144, opened 2019-03-21
node: maintain old_pids list

bhalevy:
To be used for associating coredump files with old pids.
See scylladb/scylla-dtest#900

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -38,7 +38,6 @@ class NodeError(Exception):
         Exception.__init__(self, msg)
         self.process = process
 
-
 class TimeoutError(Exception):
 
     def __init__(self, data):
@@ -99,6 +98,7 @@ class Node(object):
         self.remote_debug_port = remote_debug_port
         self.initial_token = initial_token
         self.pid = None
+        self.all_pids = []
         self.data_center = None
         self.workload = None
         self.__config_options = {}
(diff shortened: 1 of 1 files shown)


bhalevy:
Test run with the changes here: [manual-and-scheduled-tests/job/byo_build_tests_dtest/111](http://jenkins.cloudius-systems.com:8080/view/master/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/111)

ccmlib/node.py
```diff
@@ -1660,6 +1660,10 @@ def _update_pid(self, process):
 
         try:
             with open(pidfile, 'r') as f:
+                if self.old_pids is None:
+                    self.old_pids = []
+                if self.pid is not None:
+                    self.old_pids += [self.pid]
```
slivne: this is ok - yet I am not a pythonist - it may be preffered to use extend vs += 

@roydahan / @bentsi - whats your view

slivne: why are you using old_pids - it incldued the current running pid - maybe all_pids is better (but I agree its not a real improvement)

bhalevy: I used `old_pids` for backward compatibility so the code in dtest that relies on it could work with ccm without this change.  That said, `all_pids` could work too.
Regarding `+=`, besides it being used all over the place, `append(self.pid)` would be a better alternative for adding a single item (and `extend` for splicing a list / multiple elements)

bhalevy:
@slivne please review the revised change: a8674820efd9f66018be1454fce56fa96ac3b0a7

bhalevy:
@slivne ping

slivne:
LGTM

Issue #145, opened 2019-04-02
dtest failed with Inet address 127.0.70.3:9042 is not available: [Errno 98] Address already in use


bhalevy:
I have no clue about what caused this as there is no indication that the reported address was used by another dtest that might have run in parallel to the one failed.  Opening this issue to track the problem in case we keep hitting it.

scylla-ccm version @476c0d1928d8773748f9c7dabd532b49d5ed9ea1

Seen in [dtest-release/75/testReport/materialized_views_test/TestMaterializedViews/mv_populating_from_existing_data_during_node_remove_test](http://jenkins.cloudius-systems.com:8080/view/master/job/scylla-master/job/dtest-release/75/testReport/materialized_views_test/TestMaterializedViews/mv_populating_from_existing_data_during_node_remove_test/):
```
Stacktrace
  File "/usr/lib64/python2.7/unittest/case.py", line 367, in run
    testMethod()
  File "/jenkins/workspace/scylla-master/dtest-release@2/scylla-dtest/materialized_views_test.py", line 581, in mv_populating_from_existing_data_during_node_remove_test
    self._mv_populating_from_existing_data_during_changes_test('remove node', nodes=4, rf=3, mvs=10, prefill=40000, fail=False)
  File "/jenkins/workspace/scylla-master/dtest-release@2/scylla-dtest/materialized_views_test.py", line 596, in _mv_populating_from_existing_data_during_changes_test
    session = self.prepare(rf=rf, nodes=nodes, options={'prometheus_port': 0})
  File "/jenkins/workspace/scylla-master/dtest-release@2/scylla-dtest/materialized_views_test.py", line 89, in prepare
    cluster.start(jvm_args=jvm_args,wait_other_notice=True,wait_for_binary_proto=True)
  File "/jenkins/workspace/scylla-master/dtest-release@2/scylla-ccm/ccmlib/scylla_cluster.py", line 87, in start
    profile_options=profile_options)
  File "/jenkins/workspace/scylla-master/dtest-release@2/scylla-ccm/ccmlib/scylla_node.py", line 305, in start
    common.check_socket_available(itf)
  File "/jenkins/workspace/scylla-master/dtest-release@2/scylla-ccm/ccmlib/common.py", line 449, in check_socket_available
    raise UnavailableSocketError("Inet address %s:%s is not available: %s" % (addr, port, msg))

Inet address 127.0.70.3:9042 is not available: [Errno 98] Address already in use
```

[...]

PR #148, opened 2019-04-22
scylla_node: print offending processes when socket is unavailable

bhalevy:
We're trying to understand why the socket is not available.
It might be in use by a stale scylla process so try to find
any running scylla process using the cluster ip prefix.
Its path includes the cluster test directory that could help
understand which test case started it.

Ref #145


bhalevy:
@roydahan ping
byo test run [#271](http://jenkins.cloudius-systems.com:8080/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/271)

bhalevy:
@roydahan ping

PR #152, opened 2019-04-29
node: grep_log should only look at INFO/ERROR messages

bhalevy:
DEBUG and TRACE messages may include intermediate printouts that
shouldn't be taken into account as these logging levels are optional.

Seen with commitlog_test:TestCommitLog.test_commitlog_replay_on_startup
when it looks for no occurances of " 0 replayed mutations", but
these are printed validly by DEBUG log messages, causing the test
to fail with DEBUG=true.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -310,17 +310,21 @@ class Node(object):
     def debuglogfilename(self):
         return os.path.join(self.get_path(), 'logs', 'debug.log')
 
-    def grep_log(self, expr, filename='system.log'):
+    def grep_log(self, expr, filter_expr=None, filename='system.log'):
         """
         Returns a list of lines matching the regular expression in parameter
         in the Cassandra log of this node
         """
         matchings = []
         pattern = re.compile(expr)
+        if filter_expr:
+            filter_pattern = re.compile(filter_expr)
+        else:
+            filter_pattern = None
         with open(os.path.join(self.get_path(), 'logs', filename)) as f:
             for line in f:
                 m = pattern.search(line)
-                if m:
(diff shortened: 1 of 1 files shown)


bentsi:
generally looks good, but why DEBUG level log messages are printed? is DEBUG default log level for Scylla?

bhalevy:
@bentsi DEBUG messages aren't printed by default but there's are options / environment variables to enable them, which I used for manual testing, and they caused this issue.

ccmlib/node.py
```diff
@@ -319,9 +319,10 @@ def grep_log(self, expr, filename='system.log'):
         pattern = re.compile(expr)
         with open(os.path.join(self.get_path(), 'logs', filename)) as f:
             for line in f:
-                m = pattern.search(line)
-                if m:
-                    matchings.append((line, m))
+                if 'INFO' in line or 'ERROR' in line:
```
bentsi: better: `if "DEBUG" not in line:`
(since you missed WARN and adding WARN case makes the line longer and also there
is some degree of unpredictability, so better to exclude what you know)

bhalevy:
Makes sense. Will look for not DEBUG nor TRACE

bhalevy:
@bentsi I updated the patch. Please re-review.

bhalevy:
@bentsi I found cases where grep_log actually needs to look at DEBUG messages.
I'll change the solution to use an optional parameter with the filter to use on the log, similar to `grep_log_for_errors`.

bhalevy:
@bentsi please review.
Tested with the following tests, modified to use `filter_expr='DEBUG'` for grep_log of `replayed mutations`, with and without `DEBUG=true`

PR #153, opened 2019-05-01
Add support for using scylla relocatable packages

### LINKED ISSUE #151: add support for scylla relocatable package
This is a follow up on #150 which was resolved in a hackish way by creating ad-hoc wrapper script called `scylla.sh` that runs scylla using the dynamic loader.

As discussed with @avikivity, in addition to running a plain scylla binary (via CASSANRA_DIR), we would like to add support to use a scylla relocatable package, install it, and run scylla using the installed dynamic libraries.

fruch:
Also added optinal JAVA_TOOLS_DIR, so we don't need to create symlink to inside resources/cassandra to the java tools.

building java-tools:
```bash
./reloc/build_reloc.sh
./scripts/create-relocatable-package.py --version 3.11.3-SNAPSHOT scylla-java-tools-3.0.5.tar.gz
```
building jmx:
```bash
mvn package
./scripts/create-relocatable-package.py scylla-jmx-3.0.5.tar.gz
```
Inside dbuild we can build like that:
```bash
   python3 scripts/create-relocatable-package.py --mode release scylla-3.0.5.tar.gz
   mkdir scylla-3.0.5
   tar xvvf temp.tar.gz -C scylla-3.0.5
```

And then use CCM like those options:
```bash
   # from specific directory, after extracting 
   export JAVA_TOOLS_DIR=/home/fruch/Projects/scylla-java-tools
   ccm create fruch --install-dir=/home/fruch/Projects/scylla-next/scylla-3.0.5
   ccm populate -n 1
   ccm node1 start

   # from specific version downloading from s3
   ccm create fruch --scylla --version unstable/master:239
   ccm populate -n 1
   ccm node1 start

   # from specific version overriding specific tarfiles locally or from remote server
   export SCYLLA_PACKAGE="/home/fruch/Projects/scylla-next/temp.tar.gz"
   export SCYLLA_JAVA_TOOLS_PACKAGE="https://s3.amazonaws.com/downloads.scylladb.com/relocatable/unstable/master/239/scylla-tools-package.tar.gz"
   ccm create fruch --scylla --version 3.0.5
   ccm populate -n 1
   ccm node1 start

```

Note:
For running dtest it would be better if we open the relocatable package into a scylla repo,
so it won't break dtest git depended functionality

Fixes #151

### DIFF
--- a/ccmlib/cmds/cluster_cmds.py
+++ b/ccmlib/cmds/cluster_cmds.py
@@ -123,6 +123,16 @@ class ClusterCreateCmd(Cmd):
                           help="Supports 'org.apache.cassandra.locator.PropertyFileSnitch','org.apache.cassandra.locator.GossipingPropertyFileSnitch' used only in multidc clusters")
         parser.add_option("--id", type="int", dest="id",
                           help="Allows running multiple clusters in paralell (up to 100) each must have a unique id value 0-99, this will set the ipprefix value if one was not set to 127.0.<id>.")
+
+        parser.add_option('--scylla-package-uri', type="string", dest="scylla_package_uri",
+                          help="The path scylla relocatable package", default=None)
+
+        parser.add_option('--scylla-java-tools-package-uri', type="string", dest="scylla_java_tools_package_uri",
+                          help="The path scylla java tools relocatable package", default=None)
+
+        parser.add_option('--scylla-jmx-package-uri', type="string", dest="scylla_jmx_package_uri",
+                          help="The path scylla jmx relocatable package", default=None)
+
         return parser
 
     def validate(self, parser, options, args):
@@ -156,6 +166,14 @@ class ClusterCreateCmd(Cmd):
             print_("""WARN: c:\windows\system32\java.exe exists.
                 This may cause registry issues, and jre7 to be used, despite jdk8 being installed.
                 """)
+
+        if options.scylla_package_uri:
+            os.environ['SCYLLA_PACKAGE'] = options.scylla_package_uri
+        if options.scylla_java_tools_package_uri:
+            os.environ['SCYLLA_JAVA_TOOLS_PACKAGE'] = options.scylla_java_tools_package_uri
+        if options.scylla_jmx_package_uri:
+            os.environ['SCYLLA_JMX_PACKAGE'] = options.scylla_jmx_package_uri
+
 
     def run(self):
         try:
@@ -297,7 +315,6 @@ class ClusterAddCmd(Cmd):
             print_("This JMX port is already in use. Choose another.", file=sys.stderr)
             parser.print_help()
             sys.exit(1)
-
 
         self.jmx_port = options.jmx_port
         self.remote_debug_port = options.remote_debug_port
--- a/ccmlib/cmds/cluster_cmds.py
+++ b/ccmlib/cmds/cluster_cmds.py
@@ -7,7 +7,7 @@ from six import print_
 from ccmlib import common, repository
 from ccmlib.cluster import Cluster
 from ccmlib.cluster_factory import ClusterFactory
-from ccmlib.cmds.command import Cmd
+from ccmlib.cmds.command import Cmd, PlainHelpFormatter
 from ccmlib.common import ArgumentError
 from ccmlib.dse_cluster import DseCluster
 from ccmlib.scylla_cluster import ScyllaCluster
@@ -60,7 +60,6 @@ def parse_populate_count(v):
     else:
         return [int(t) for t in tmp]
 
-
 class ClusterCreateCmd(Cmd):
 
     def description(self):
@@ -68,7 +67,7 @@ class ClusterCreateCmd(Cmd):
 
     def get_parser(self):
         usage = "usage: ccm create [options] cluster_name"
-        parser = self._get_default_parser(usage, self.description())
+        parser = self._get_default_parser(usage, self.description(), formatter=PlainHelpFormatter())
         parser.add_option('--no-switch', action="store_true", dest="no_switch",
                           help="Don't switch to the newly created cluster", default=False)
         parser.add_option('-p', '--partitioner', type="string", dest="partitioner",
@@ -132,6 +131,20 @@ class ClusterCreateCmd(Cmd):
 
         parser.add_option('--scylla-jmx-package-uri', type="string", dest="scylla_jmx_package_uri",
                           help="The path scylla jmx relocatable package", default=None)
+
+        parser.epilog = """
+        
+        Examples of using relocatable packages:
+        
+        # create cluster from version uploaded to s3 (of the daily/nightly as example)
+        ccm create fruch --scylla --version unstable/master:239
+        
+        # create cluster with own versions of each package
+        ccm create fruch --scylla --version temp \\
+            --scylla-package-uri=../scylla/temp.tar.gz \\ 
+            --scylla-java-tools-package-uri=../scylla-tools-java/temp.tar.gz \\
+            --scylla-jmx-package-uri=../scylla-jmx/temp.tar.gz
+        """
 
         return parser
 
--- a/ccmlib/cmds/cluster_cmds.py
+++ b/ccmlib/cmds/cluster_cmds.py
@@ -123,7 +123,7 @@ class ClusterCreateCmd(Cmd):
         parser.add_option("--id", type="int", dest="id",
                           help="Allows running multiple clusters in paralell (up to 100) each must have a unique id value 0-99, this will set the ipprefix value if one was not set to 127.0.<id>.")
 
-        parser.add_option('--scylla-package-uri', type="string", dest="scylla_package_uri",
+        parser.add_option('--scylla-core-package-uri', type="string", dest="scylla_core_package_uri",
                           help="The path scylla relocatable package", default=None)
 
         parser.add_option('--scylla-java-tools-package-uri', type="string", dest="scylla_java_tools_package_uri",
@@ -180,8 +180,8 @@ class ClusterCreateCmd(Cmd):
                 This may cause registry issues, and jre7 to be used, despite jdk8 being installed.
                 """)
 
-        if options.scylla_package_uri:
-            os.environ['SCYLLA_PACKAGE'] = options.scylla_package_uri
+        if options.scylla_core_package_uri:
+            os.environ['SCYLLA_CORE_PACKAGE'] = options.scylla_core_package_uri
         if options.scylla_java_tools_package_uri:
             os.environ['SCYLLA_JAVA_TOOLS_PACKAGE'] = options.scylla_java_tools_package_uri
         if options.scylla_jmx_package_uri:
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -320,6 +320,7 @@ def get_stress_bin(install_dir):
         os.path.join(install_dir, 'tools', 'stress', 'bin', 'stress'),
         os.path.join(install_dir, 'tools', 'bin', 'stress'),
         os.path.join(install_dir, 'tools', 'bin', 'cassandra-stress'),
+        os.path.join(install_dir, 'scylla-java-tools', 'tools', 'bin', 'cassandra-stress'),
         os.path.join(install_dir, 'resources', 'cassandra', 'tools', 'bin', 'cassandra-stress')
     ]
     candidates = [platform_binary(s) for s in candidates]
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -371,7 +371,8 @@ def isScylla(install_dir):
     return (os.path.exists(os.path.join(install_dir, 'scylla')) or
             os.path.exists(os.path.join(install_dir, 'build', 'debug', 'scylla')) or
             os.path.exists(os.path.join(install_dir, 'build', 'dev', 'scylla')) or
-            os.path.exists(os.path.join(install_dir, 'build', 'release', 'scylla')))
+            os.path.exists(os.path.join(install_dir, 'build', 'release', 'scylla')) or
+            os.path.exists(os.path.join(install_dir, 'bin', 'scylla')))
 
 
 def isOpscenter(install_dir):
@@ -398,6 +399,8 @@ def scylla_extract_install_dir_and_mode(install_dir):
     elif install_dir.endswith('build/release') or install_dir.endswith('build/release/'):
         scylla_mode = 'release'
         install_dir = str(os.path.join(install_dir, os.pardir, os.pardir))
+    elif os.path.exists(os.path.join(install_dir, 'libreloc')):
+        scylla_mode = 'reloc'
     return install_dir, scylla_mode
 
 
@@ -567,9 +570,11 @@ def get_version_from_build(install_dir=None, node_path=None):
 
 
 def get_scylla_version(install_dir):
-    # FIXME
     if isScylla(install_dir):
-        return '2.2'
+        scylla_version_file = os.path.join(install_dir, 'SCYLLA-VERSION-FILE')
+        if os.path.exists(scylla_version_file):
+            return open(scylla_version_file).read().strip()
+        return '3.0'
     else:
         return None
 
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -12,6 +12,7 @@ from ccmlib import common
 from ccmlib.cluster import Cluster
 from ccmlib.scylla_node import ScyllaNode
 from ccmlib.node import NodeError
+from ccmlib import scylla_repository
 
 SNITCH = 'org.apache.cassandra.locator.GossipingPropertyFileSnitch'
 
@@ -22,13 +23,19 @@ class ScyllaCluster(Cluster):
                  create_directory=True, version=None, verbose=False,
                  force_wait_for_cluster_start=False, manager=None, **kwargs):
         install_func = common.scylla_extract_install_dir_and_mode
-        install_dir, self.scylla_mode = install_func(install_dir)
+
+        cassandra_version = kwargs.get('cassandra_version', version)
+        if kwargs.get('cassandra_version', version):
+            self.scylla_mode = 'reloc'
+        else:
+            install_dir, self.scylla_mode = install_func(install_dir)
+            
         self.started = False
         self.force_wait_for_cluster_start = force_wait_for_cluster_start
         super(ScyllaCluster, self).__init__(path, name, partitioner,
                                             install_dir, create_directory,
                                             version, verbose,
-                                            snitch=SNITCH)
+                                            snitch=SNITCH, cassandra_version=cassandra_version)
 
         self._scylla_manager=None
         if not manager:
@@ -45,7 +52,7 @@ class ScyllaCluster(Cluster):
             self._scylla_manager = ScyllaManager(self,manager)
 
     def load_from_repository(self, version, verbose):
-        raise NotImplementedError('ScyllaCluster.load_from_repository')
+        return scylla_repository.setup(version, verbose)
 
     def create_node(self, name, auto_bootstrap, thrift_interface,
                     storage_interface, jmx_port, remote_debug_port,
@@ -143,16 +150,6 @@ class ScyllaCluster(Cluster):
         if self._scylla_manager:
             self._scylla_manager.stop(gently)
         Cluster.stop(self,wait,gently,wait_seconds=wait_seconds, wait_other_notice=wait_other_notice, other_nodes=other_nodes)
-
-    def version(self):
-        return self.cassandra_version()
-
-    def cassandra_version(self):
-        # TODO: Handle versioning
-        # Return 2.2 as it changes some option values for tools. 
-        # Our tools are actually at 3.x level (-ish), but otoh
-        # the server is more or less 2.1-2.2
-        return '3.0'
 
     def get_scylla_mode(self):
         return self.scylla_mode
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -25,7 +25,7 @@ class ScyllaCluster(Cluster):
         install_func = common.scylla_extract_install_dir_and_mode
 
         cassandra_version = kwargs.get('cassandra_version', version)
-        if kwargs.get('cassandra_version', version):
+        if cassandra_version:
             self.scylla_mode = 'reloc'
         else:
             install_dir, self.scylla_mode = install_func(install_dir)
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -88,7 +88,7 @@ class ScyllaNode(Node):
         self._mem_set_during_test = True
 
     def get_install_cassandra_root(self):
-        return os.path.join(self.get_install_dir(), 'resources', 'cassandra')
+        return self.get_tool_java_dir()
 
     def get_node_cassandra_root(self):
         return os.path.join(self.get_path())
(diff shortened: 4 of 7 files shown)


slivne:
lets try todo this in a more complete way

there are relocatable packaged for scylla-tyools and scylla-jmx

please have all of those added and supported as well and then we cna remove
the change

The forward looking direction is not to have relocatable compiled locally
at all but downloaded from a repo.

ccmlib/scylla_node.py
```diff
@@ -552,9 +552,10 @@ def import_bin_files(self):
         files = ['cassandra.in.sh', 'nodetool']
         os.makedirs(os.path.join(self.get_path(), 'resources', 'cassandra',
                                  'bin'))
+        tool_dir = os.environ.get('TOOLS_JAVA_DIR', os.path.join(self.get_install_dir(), 'resources', 'cassandra'))
```
bhalevy: nit: let's call this variable `tools_java_dir` rather than invent yet another name. It'll be clearer to the reader.

ccmlib/scylla_node.py
```diff
@@ -567,24 +568,39 @@ def import_bin_files(self):
                                  'tools', 'bin'))
         for name in files:
             self.hard_link_or_copy(os.path.join(self.get_install_dir(),
-                                                'resources', 'cassandra',
+                                                tool_dir,
```
bhalevy: `tool_dir`'s definition includes `self.get_install_dir()` in the default case.
Seems odd it's join()ed here this way...  Shouldn't it be same as above?
```
-            self.hard_link_or_copy(os.path.join(self.get_install_dir(),
-                                                'resources', 'cassandra',
+            self.hard_link_or_copy(os.path.join(tool_dir,
```

fruch:
@slivne are those being uploaded regulatory anywhere yet ?
so we can download them by versions if needed ?

so we might have it in 3 options, like that:
```
# local file
export SCYLLA_PACKAGE=/home/myuser/git/scylla/scylla-3.0.5.tar.gz
# remote file
export SCYLLA_PACKAGE=http://s3-something/scylla-3.0.5.tar.gz
# a formal version, that we'll know where to download from (like Cassandra have)
export SCYLLA_VERSION=3.0.5
```

and then 
JAVA_TOOLS_VERSION, SCYLLA_JMX_PACKAGE, SCYLLA_JMX_VESION
in a similar fashion ?

ccmlib/scylla_node.py
bhalevy: Also, can you please extract this change into a separate patch?
It should be valid regardless of the relocatable package

ccmlib/scylla_node.py
```diff
@@ -567,24 +568,39 @@ def import_bin_files(self):
...
         # TODO: - currently no scripts only executable - copying exec
         scylla_mode = self.cluster.get_scylla_mode()
+        if scylla_mode == 'reloc':
+            repos_root = '../..'
+            self.hard_link_or_copy(os.path.join(self.get_install_dir(),
+                                                'bin', 'scylla'),
+                                   os.path.join(self.get_bin_dir(), 'scylla'))
+
+            os.symlink(os.path.join(self.get_install_dir(), 'libexec'),
+                                   os.path.join(self.get_path(), 'libexec'))
```
bhalevy: nit: misaligned line

ccmlib/scylla_node.py
```diff
@@ -567,24 +568,39 @@ def import_bin_files(self):
...
+
+            os.symlink(os.path.join(self.get_install_dir(), 'libreloc'),
+                                   os.path.join(self.get_path(), 'libreloc'))
```
bhalevy: nit: misaligned line

slivne:
On Wed, May 1, 2019 at 10:45 AM Israel Fruchter <notifications@github.com>
wrote:

> @slivne <https://github.com/slivne> are those being uploaded regulatory
> anywhere yet ?
> so we can download them by versions if needed ?
>

They aren't but will be in the nearest future - if we do this change - I
want it to be done across all.

so we might have it in 3 options, like that:
>
> # local file
> export SCYLLA_PACKAGE=/home/myuser/git/scylla/scylla-3.0.5.tar.gz
> # remote file
> export SCYLLA_PACKAGE=http://s3-something/scylla-3.0.5.tar.gz
> # a formal version, that we'll know where to download from (like Cassandra have)
> export SCYLLA_VERSION=3.0.5
>
> and then
> JAVA_TOOLS_VERSION, SCYLLA_JMX_PACKAGE, SCYLLA_JMX_VESION
> in a similar fashion ?
>

I think SCYLLA_VERSION is good enough - its all the packages for everything
- they will be uploaded side by side

In addition we need a way to download a version from a jenkins uploaded
build - that will alos be a single directory side by side

 SCYLLA_JMX_PACKAGE, SCYLLA_TOOLS_PACKAGE are for the other extreme that
someone wants to specify the 3 packages directly (this is what we should do
for now)

Please also allow override - e.g. someone wants to test SCYLLA_VERSION and
an overloaded SCYLLA_PACKAGE or an overloaded SCYLLA_TOOLS_JAVA.

ccmlib/scylla_node.py
fruch: Will do

ccmlib/scylla_node.py
fruch: you are correct, I have used those specific tools, hence I've missed it

fruch:
Now we can download packages from:
https://s3.console.aws.amazon.com/s3/buckets/downloads.scylladb.com/relocatable/unstable/master

but since they are no stripped, it's a bit unpractical to download 1Gb, I think we should upload both options, and raised: 
https://github.com/scylladb/scylla-pkg/issues/353

fruch:
triggered BYO, to see this one doesn't break anything:
https://jenkins.scylladb.com/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/300/

fruch:
@slivne 
care to review this one ?

@avikivity, FYI

fruch:
this passed byo (running from source):
http://jenkins.scylladb.com/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/304/

avikivity:
I don't understand how to use this. I expected something like ccm --package=... but I see code to download stuff from S3.

ccm should not download stuff from S3, those URLs are subject to change and are not available to a poor user who just build Scylla and wants to test it with ccm. Downloading packages should be separate from running them, since it depends on which version you want to run.

fruch:
You need to supply ccm with 3 tar.gz files, for now it's via environment
variables, they can be local file and can be http address.

the idea was that you'll be able to supply also specific formal versions,
and that would be downloaded from s3. again for the poor user that isn't
compiling scylla from source.

Cassandra is downloading binaries out of their site, I was trying to get to
the same end, that will be able to load multiple versions and even test
upgrade in dtest as Cassandra are doing.

but maybe I not sure exactly what are we aiming here...

avikivity:
Maybe it's possible but the documentation is unclear.

If it can be moved away from environment variables into a more self-documenting form (`ccm --help`) it would be a great help.

bentsi:
@fruch can you please add to `ccm --help` the instruction of how to use this as Avi suggested

fruch:
this is what added to `ccm create --help`:
```
  --scylla-package-uri=SCYLLA_PACKAGE_URI
                        The path scylla relocatable package
  --scylla-java-tools-package-uri=SCYLLA_JAVA_TOOLS_PACKAGE_URI
                        The path scylla java tools relocatable package
  --scylla-jmx-package-uri=SCYLLA_JMX_PACKAGE_URI                                                    
                        The path scylla jmx relocatable package                 
                                                                                           
                                                                                    
        Examples of using relocatable packages:                      
                                                                                                     
        # create cluster from version uploaded to s3 (of the daily/nightly as example)               
        ccm create fruch --scylla --version unstable/master:239                            
                                                                                           
        # create cluster with own versions of each package                          
        ccm create fruch --scylla --version temp \                                                  
            --scylla-package-uri=../scylla/temp.tar.gz \                                            
            --scylla-java-tools-package-uri=../scylla-tools-java/temp.tar.gz \          
            --scylla-jmx-package-uri=../scylla-jmx/temp.tar.gz  
```

@avikivity  is there a bit more clear on how to use it ? 

@penberg, addressing those two, would help use it for our daily needs much better/faster:
https://github.com/scylladb/scylla-pkg/issues/397
https://github.com/scylladb/scylla-pkg/issues/353

fruch:
gave it one more run in byo,  to make sure we aren't breaking anything with this vesion:
http://jenkins.scylladb.com/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/346/

bentsi:
@bhalevy @slivne please review

bentsi:
@fruch please remove `python3 fixes` commit

ccmlib/scylla_repository.py
```diff
@@ -0,0 +1,170 @@
...
+            return None
+    else:
+        return None
+
+
+def clean_all():
+    rmdirs(__get_dir())
+
+
+def __download(url, target, username=None, password=None, show_progress=False):
```
bentsi: Check if we can import this from the cassandra file

roydahan:
@fruch what is the status of this one?

ccmlib/scylla_repository.py
fruch: fixed

fruch:
@bentsi I also remove the python3 fixes, since it's merged already

@roydahan I consider this to be ready to merge

ccmlib/scylla_repository.py
```diff
@@ -0,0 +1,129 @@
...
+from six import print_
+from six.moves import urllib
+
+from ccmlib.common import (ArgumentError, CCMError, get_default_path, rmdirs, validate_install_dir)
+from ccmlib.repository import __download
+
+GIT_REPO = "http://github.com/scylladb/scylla.git"
+
+RELOCATABLE_URLS = {
+    'unstable/master': 'https://s3.amazonaws.com/downloads.scylladb.com/relocatable/unstable/master/'
```
roydahan: Do we need to have a map between the versions and their location?
I'm not sure I understand this part and the setup part...

roydahan:
@slivne can you please also review it?

ccmlib/scylla_repository.py
fruch: Yeah this mapping can be remove, I wasn't clear to me at the time where would be the locations of the packages, and I had only examples from unstable/master.

I can use a base url and try to append to that

I'm still missing where formal version would be located

fruch: I've removed the mapping, and now it's a string template.

what's not clear about the setup part ? the version related code ? or something else ?

the idea that I wanted that we could have things like:
```bash
# version format
# [branch]:[job id | "latest"]
unstable/master:234  # specific job version, only thing we support now
3.1:latest # latest version available for branch
3.1:22 # specific job for branch
```

roydahan: Now it's more clear to me.
I didn't understand the version part.

Eventually we will need to have a map of official versions that we can pull automatically (like cassandra has) so we can test things like upgrades inside dtest.

ccmlib/scylla_repository.py
```diff
@@ -0,0 +1,127 @@
...
+
+from six import print_
+from six.moves import urllib
+
+from ccmlib.common import (ArgumentError, CCMError, get_default_path, rmdirs, validate_install_dir)
+from ccmlib.repository import __download
+
+GIT_REPO = "http://github.com/scylladb/scylla.git"
+
+RELOCATABLE_URLS_BASE = 'https://s3.amazonaws.com/downloads.scylladb.com/relocatable/{0}/{1}'
```
slivne: that will not be the path for the official versions it will be probably be parallel to the scylla releases
http://downloads.scylladb.com/downloads/scylla/
if we will have one relocatable package it will be parallel to rpm/deb - if we will have one for each distribution it will be under rpm/deb

ccmlib/scylla_node.py
```diff
@@ -584,28 +590,41 @@ def import_bin_files(self):
...
-                                            '..', 'scylla-jmx', 'scripts',
-                                            'scylla-jmx'),
-                               os.path.join(self.get_bin_dir(),
-                                            'scylla-jmx'))
+        if scylla_mode == 'reloc':
+            relative_repos_root = '../..'
+            self.hard_link_or_copy(os.path.join(self.get_install_dir(), 'bin', 'scylla'),
+                                   os.path.join(self.get_bin_dir(), 'scylla'))
+
+            os.symlink(os.path.join(self.get_install_dir(), 'libexec'),
```
slivne: why is this a symlink and not a hard_lonk_or_copy like the others ?

ccmlib/cmds/cluster_cmds.py
```diff
@@ -123,6 +122,30 @@ def get_parser(self):
...
+        
+        # create cluster from version uploaded to s3 (of the daily/nightly as example)
+        ccm create fruch --scylla --version unstable/master:239
+        
+        # create cluster with own versions of each package
+        ccm create fruch --scylla --version temp \\
+            --scylla-package-uri=../scylla/temp.tar.gz \\ 
+            --scylla-java-tools-package-uri=../scylla-tools-java/temp.tar.gz \\
+            --scylla-jmx-package-uri=../scylla-jmx/temp.tar.gz
+        """
```
slivne: what happens if the user will specify only scylla-package-uri ? is this supported ?

ccmlib/common.py
```diff
@@ -398,6 +400,8 @@ def scylla_extract_install_dir_and_mode(install_dir):
     elif install_dir.endswith('build/release') or install_dir.endswith('build/release/'):
         scylla_mode = 'release'
         install_dir = str(os.path.join(install_dir, os.pardir, os.pardir))
+    elif os.path.exists(os.path.join(install_dir, 'libreloc')):
+        scylla_mode = 'reloc'
```
slivne: does reloc force to be release as well - e.g. is thre a reloc that is debug and a reloc that is release

ccmlib/scylla_cluster.py
```diff
@@ -22,13 +23,19 @@ def __init__(self, path, name, partitioner=None, install_dir=None,
                  create_directory=True, version=None, verbose=False,
                  force_wait_for_cluster_start=False, manager=None, **kwargs):
         install_func = common.scylla_extract_install_dir_and_mode
-        install_dir, self.scylla_mode = install_func(install_dir)
+
+        cassandra_version = kwargs.get('cassandra_version', version)
+        if kwargs.get('cassandra_version', version):
```
slivne: nit - if we extracted it already why not use cassandra_version ?

ccmlib/scylla_repository.py
```diff
@@ -0,0 +1,127 @@
...
+
+def is_valid(url, qualifying=None):
+    qualifying = min_attributes if qualifying is None else qualifying
+    token = urllib.parse.urlparse(url)
+    return all([getattr(token, qualifying_attr)
+                for qualifying_attr in qualifying])
+
+
+def download_version(version, url=None, verbose=False):
+    """Download, extract, and build Cassandra tarball.
```
slivne: Cassandra tarball ?

ccmlib/scylla_repository.py
```diff
@@ -0,0 +1,127 @@
...
+
+def setup(version, verbose=True):
+    s3_url = 'https://s3.amazonaws.com/downloads.scylladb.com/relocatable/unstable/master/latest'
+    type_n_version = version.split(':')
+    if len(type_n_version) == 2:
+        s3_version = type_n_version[1]
+        s3_url = RELOCATABLE_URLS_BASE.format(type_n_version[0], s3_version)
+        version = os.path.join(*type_n_version)
+
+    cdir = version_directory(version)
```
slivne: if we failed to download some part - the repo will be corrupt and the user will be stuck - better make sure we protect against this - e.g. temp direcoty and rename or something in those lines

ccmlib/scylla_repository.py
```diff
@@ -0,0 +1,127 @@
...
+    if cdir is None:
+        url = os.environ.get('SCYLLA_PACKAGE', os.path.join(s3_url, 'scylla-package.tar.gz'))
+        download_version(version, verbose=verbose, url=url)
+
+        url = os.environ.get('SCYLLA_JAVA_TOOLS_PACKAGE', os.path.join(s3_url, 'scylla-tools-package.tar.gz'))
+        download_version(os.path.join(version, 'scylla-java-tools'), verbose=verbose, url=url)
+
+        # hack to make the relocatable tools work
+        cdir = version_directory(version)
```
slivne: why again ?

ccmlib/scylla_repository.py
```diff
@@ -0,0 +1,127 @@
...
+        scylla_java_tools_dir = os.path.join(cdir, 'scylla-java-tools')
+        for jar_file in glob.glob(scylla_java_tools_dir + "/*.jar"):
+            shutil.copy(jar_file, os.path.join(scylla_java_tools_dir, "lib"))
+
+        url = os.environ.get('SCYLLA_JMX_PACKAGE', os.path.join(s3_url, 'scylla-jmx-package.tar.gz'))
+        download_version(os.path.join(version, 'jmx'), verbose=verbose, url=url)
+
+        cdir = version_directory(version)
```
slivne: why again ?

ccmlib/scylla_repository.py
```diff
@@ -0,0 +1,127 @@
...
+        download_version(version, verbose=verbose, url=url)
+
+        url = os.environ.get('SCYLLA_JAVA_TOOLS_PACKAGE', os.path.join(s3_url, 'scylla-tools-package.tar.gz'))
+        download_version(os.path.join(version, 'scylla-java-tools'), verbose=verbose, url=url)
+
+        # hack to make the relocatable tools work
+        cdir = version_directory(version)
+        scylla_java_tools_dir = os.path.join(cdir, 'scylla-java-tools')
+        for jar_file in glob.glob(scylla_java_tools_dir + "/*.jar"):
+            shutil.copy(jar_file, os.path.join(scylla_java_tools_dir, "lib"))
```
slivne: I don;t understand this copy

ccmlib/scylla_repository.py
```diff
@@ -0,0 +1,127 @@
...
+    """Download, extract, and build Cassandra tarball.
+
+    if binary == True, download precompiled tarball, otherwise build from source tarball.
+    """
+    target_dir = None
+    try:
+        if os.path.exists(url) and url.endswith('.tar.gz'):
+            target = url
+        elif is_valid(url):
+            _, target = tempfile.mkstemp(suffix=".tar.gz", prefix="ccm-")
```
slivne: is this relevant for us ?

ccmlib/scylla_repository.py
fruch: o.k. we could amend this, once we have at least one formal release with relocatable.

ccmlib/scylla_node.py
fruch: since hard_link_or_copy doesn't handle directories correctly.
the fallback shutil.copy isn't recursive

ccmlib/scylla_repository.py
fruch: copy pasted from Cassandra CCM , I'll rewrite this comment

ccmlib/scylla_repository.py
fruch: yes it is, for the cases we download from http (i.e. from s3)

ccmlib/scylla_cluster.py
fruch: hmmm, a disturbance in the force. I'll rectify that

ccmlib/common.py
fruch: That's a good question, right now it's a mode of it's own, and they are all release, but in ccm this mode is basically used for some of the path finding. dtest has some code querying those mode, and selecting/deselecting tests based on it.

we might need to make the dtest a bit clear about this usage of the mode, or get it in some other mean.

as for a general question, do we want to publish multiple version of the relocatable ?, i.e. release/debug/dev ?

ccmlib/cmds/cluster_cmds.py
fruch: yes it's supported, as long as a valid version passed down in --version, the other tarball will be download from s3.

you think that might be all or nothing here ? to avoid mixing wrong versions ? or ease of use here is more impotent ?

ccmlib/scylla_repository.py
fruch: cause now we have a directory, after some packages were downloaded into it.
we didn't had that before. i.e. it won't download again if directory exists as you noticed in previous comment

ccmlib/scylla_repository.py
fruch: It's a hack to copy some of the .jar file into the /lib directory, I'm not sure we they aren't there to begin with.
I guess it might be a bug, I forgot to open it... will do.
still need the code here until it would be solved or clarify.

ccmlib/scylla_repository.py
fruch: third time icecream, this one it actually not needed, removing

ccmlib/common.py
hagitsegev: On https://github.com/scylladb/scylla-pkg/pull/450 I support dev and debug.
Example build: https://jenkins.scylladb.com/view/master/job/scylla-master/job/manual-and-scheduled-tests/job/next_debug/426/ (failed on flaky tests, artefacts exist).

ccmlib/scylla_repository.py
fruch: I've opened an issue about it
https://github.com/scylladb/scylla-tools-java/issues/104
@glommer  seems to have a fix for it, somewhere...

ccmlib/scylla_repository.py
fruch: I've address this, and now the download and untar is done into a temp directory and the moved into place once everything succeeded.

ccmlib/scylla_cluster.py
fruch: fixed this

ccmlib/common.py
fruch: I would like to have some information inside the tarball, to identify the build type, instead of different names or paths for those files. once we'll have that defined we'll update ccm with this ability

fruch:
@slivne, thanks for you review, 2 pending issues left I would love to get you agreement on

ccmlib/common.py
slivne: @fruch please create an issue somewhere for this so we can merge this pull request and not assume you/hagit will remember this

ccmlib/scylla_node.py
slivne: @fruch the issue with symlink is that after you create a cluster with CCM its supposed to remain using the version you created it with not change under your "feet" when a new relocatable package is used.

Thats why we used hard_link or copy so if its on the same volume a hard link sufficies and if its not we copy the executable don't soft link to them.

Lets fix this - it caused issues in the past

- if the issue is only a symlink to a repo then I can leave with it (given you create a followup issue to fix it)
- if the issue is a symlink to a local relocatable that can change - thats not fine and we need a fix.

ccmlib/scylla_repository.py
slivne: it seems glauber's fix is merged scylladb/scylla-tools-java@1c6019ad1de6178104bf28e2008045881dd61736 can we closed this issue and fix it correctly. We do not need to support older relocatable formats - at least I don;t understand why we need to, this is not part of the ccm branch of 3.1

ccmlib/common.py
fruch: https://github.com/scylladb/scylla-ccm/issues/165

ccmlib/scylla_repository.py
fruch: I'll remove this code, and test it again with newer builds

ccmlib/scylla_node.py
fruch: I can see why it would be a problem when working from the build directory of scylla,
in relocatable is a bit likely to change under your feet, but I'll change it to hardlink / cp all the files in those two directories

Issue #124, opened 2019-01-17
Can't stop two nodes concurrently with wait_other_notice=True.


nyh:
The dtest "materialized_views_test.TestMaterializedViews.double_node_failure_during_mv_insert_4_nodes_test" starts 4 nodes, and kills two of them with wait_other_notice=True.
The log contains something like this:
```
2019-01-17 14:42:10.215307 START: stop node node3
2019-01-17 14:42:10.215669 START: stop node node2
2019-01-17 14:42:17.325150 FINISH: stop node node2
```
The "stop node3" never finished, 
and eventually produces
```
TimeoutError: 17 Jan 2019 12:52:18 [node2] Missing: ['127.0.0.3 is now [dead|DOWN]']:
```
What happens is that we stop node3, and then "wait for node2 to notice", which means grepping node2's log for message about node3 becoming dead. But since node2 has also died, it will obviously never notice.

The "waiting for notice" should stop waiting for a certain node's log if that node itself dies.

nyh:
The new `other_nodes` parameter added in https://github.com/scylladb/scylla-ccm/commit/7238928ea5c3509a05d931d4091d4b44445f3934 allows to solve this issue by listing only the nodes to wait for, and this list should not contain either node which we are planning to take down in the test.

So we should perhaps close this issue - even though the committed workaround isn't as automatic and fool-proof as it could have been and this issue described?

bhalevy:
Can be closed with eb2f95a97ca679e85c313e3b3a803c65a2d365b1

PR #155, opened 2019-05-15
cluster, node: support optional other_nodes for wait_other_notice

bhalevy:
When stopping more than one node in parallel and waiting for other to notice
we want to skip nodes that may be stopping concurrently.

Provide an optional other_nodes parameter for the caller to pass
only those nodes that require noticing.

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -237,7 +237,7 @@ class Cluster(object):
         tokens.extend(new_tokens)
         return tokens
 
-    def remove(self, node=None, wait_other_notice=False):
+    def remove(self, node=None, wait_other_notice=False, other_nodes=None):
         if node is not None:
             if node.name not in self.nodes:
                 return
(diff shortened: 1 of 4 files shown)


bhalevy:
BYO test run: [byo_build_tests_dtest/321](https://jenkins.scylladb.com/view/master/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/321)

bhalevy:
ping @nyh

nyh:
Note: this is related (and possibly should close) https://github.com/scylladb/scylla-ccm/issues/124.
The problem is also mentioned (among other problems) in https://github.com/scylladb/scylla-dtest/issues/795.
I'll review now.

bhalevy:
> I'm marking approval because the code is correct and can be used to fix real dtest problems, and is general and future-proof.
> But I'd still like to comment that I would have gone a different route to solve just the problem we actually encounter: As explained in #124, the code already, correctly, doesn't wait for non-live nodes to notice the event. But the problem is nodes can die **while** we wait for it to notice. So the fix should be: while waiting on another node's log to mention "noticing" the event, we should also be on the lookout for messages about the node going down, and if that happens - we should stop waiting on this log (and accept this as a success - not a failure). If we did this, the user would not need to manually build a list of nodes to wait for as this patch allows him to.

I see your point.  It is more complicated and can be added in the future.

PR #157, opened 2019-06-02
treewide: use yaml.load(f, Loader=yaml.FullLoader)

bhalevy:
Address following warning:
    YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated

Please read https://msg.pyyaml.org/load for full details.

Fixes #142

Signed-off-by: Benny Halevy <bhalevy@scylladb.com>

PR #158, opened 2019-06-03
YAML load isn't safe, opt to use safe_load

oz123:
YAML docs mention straight in the beginning:

> Warning: It is not safe to call yaml.load with any data received from
> an untrusted source! yaml.load is as powerful as pickle.load and so may
> call any Python function. Check the yaml.safe_load function though.

Many developers choose to use `yaml.load`. Probably, the result of people
using IDEs which offer completions in alphabetical order. It's easy to
do:

  import yaml

  yaml.(here comes a drop down list of options)
        add_constructor
	...
	...
        load
	....
	....
	#safe_load is hidding here ...

It's easy to use without ever knowing how dangerous this could be.
This commit replaces all occurences of `yaml.load` with
`yaml.safe_load`.

### DIFF
--- a/ccmlib/cluster_factory.py
+++ b/ccmlib/cluster_factory.py
@@ -17,7 +17,7 @@ class ClusterFactory():
         cluster_path = os.path.join(path, name)
         filename = os.path.join(cluster_path, 'cluster.conf')
         with open(filename, 'r') as f:
-            data = yaml.load(f)
+            data = yaml.safe_load(f)
         try:
             install_dir = None
             if 'install_dir' in data:
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -94,7 +94,7 @@ def get_config():
         return {}
 
     with open(config_path, 'r') as f:
-        return yaml.load(f)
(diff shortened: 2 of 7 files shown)


oz123:
Do note that this PR is very much related to https://github.com/scylladb/scylla-ccm/pull/157.
But #157 does not fix the potential security issue with `yaml.load` since it uses the `FullLoader`.
This loader is still very dangerous, since it has the ability to "unpickle" all sort of stuff.
`yaml.safe_load` is implemented as:

```
def safe_load(stream):
    """
    Parse the first YAML document in a stream
    and produce the corresponding Python object.
    Resolve only basic YAML tags.
    """
    return load(stream, SafeLoader)
```

This also obsoletes [`common.load_config`](https://github.com/scylladb/scylla-ccm/pull/157/files#diff-e19f0f0ec871623be418f8c4e97f9f46R90).

Appologies, if I am barging with this PR. I only noticed this so much related to @bhalevy's PR after submitting this one.

bhalevy:
@slivne this looks good to me.

PR #160, opened 2019-06-05
scylla_node: add _mem_mb_per_cpu

bhalevy:
ScyllaNode._mem_mb_per_cpu holds the memory size per cpu
for the node in MB (default 512).

Use self._mem_mb_per_cpu * self.smp to calculate --memory parameter,
if not given otherwise.

Add ScyllaNode.set_mem_mb_per_cpu() to set this parameter.

Ref scylladb/scylla-dtest#1006


bhalevy:
Cc @tgrabiec @bentsi @juliayakovlev

PR #161, opened 2019-06-20
scylla_cluster: enlarge start timeout

bhalevy:
As seen on the lab machines, when running a number of tests
in parallel (5) and using a spinning disk, 5 minutes are not
enough for clusters to start.  Seen normal times in the order of
8 minutes, so enlarge the timeout to 10 minutes to provide
additional slack.

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -101,7 +101,7 @@ class ScyllaCluster(Cluster):
                     # updated code, scylla starts CQL only by default
                     # process should not be checked for scylla as the
                     # process is a boot script (that ends after boot)
-                    node.watch_log_for(start_message, timeout=300,
+                    node.watch_log_for(start_message, timeout=600,
                                        verbose=verbose, from_mark=mark)
                 except RuntimeError:
                     raise Exception("Not able to find start "


fruch:
what is the backport-needed tag mean here in ccm ? to all branches ?

nyh:
This is very sad. It shouldn't take 5 minutes, and definitely not 10 minutes, to start a cluster. I don't know if we're running too much stuff in parallel without gain, or doing too much when initializing a dtest cluster, or probably both, but eventually we need to fix this, not to keep increasing this timeout....

bhalevy:
> This is very sad. It shouldn't take 5 minutes, and definitely not 10 minutes, to start a cluster. I don't know if we're running too much stuff in parallel without gain, or doing too much when initializing a dtest cluster, or probably both, but eventually we need to fix this, not to keep increasing this timeout....

@nyh I agree it's unfortunate.  We saw this only on a server using magnetic drive for the cluster storage and enough tests running in parallel. FWIW, with SSDs it takes much less.

fruch:
@bhalevy I did see this happening also on our build servers in GCE (again they are not with SSDs)

bhalevy:
> what is the backport-needed tag mean here in ccm ? to all branches ?

@fruch in this case all branches can be considered.  branch-3.1 in particular

PR #167, opened 2019-08-08
Make ccm great(installable as package) again

fruch:
* added the psutil that was missing
* run shell script were missing when used as installed package


fruch:
taking it for a spin in BYO:
https://jenkins.scylladb.com/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/589/console

ccmlib/scylla_node.py
```diff
@@ -637,7 +637,7 @@ def import_bin_files(self):
                                                  'scylla-jmx'))
 
         parent_dir = os.path.dirname(os.path.realpath(__file__))
-        resources_bin_dir = os.path.join(parent_dir, '..', 'resources', 'bin')
+        resources_bin_dir = os.path.join(parent_dir, 'resources', 'bin')
```
slivne: how come this works ?

fruch: The resources folder is inside ccm, moving into ccmlib and making is sure it's part of the installation doesn't change the behavior.

slivne:
Ok - you can merge if BYO says it works

fruch:
BYO said yes

PR #170, opened 2019-08-21
Use IPs instead of localhost for nodetool commands

tarzanek:
note this might break IPv6 IP tests

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -704,8 +704,9 @@ class Node(object):
         if capture_output and not wait:
             raise common.ArgumentError("Cannot set capture_output while wait is False.")
         env = self.get_env()
+        host = self.address()
         nodetool = self.get_tool('nodetool')
-        args = [nodetool, '-h', 'localhost', '-p', str(self.jmx_port)]
+        args = [nodetool, '-h', host, '-p', str(self.jmx_port)]
         args += cmd.split()
         if capture_output:
             p = subprocess.Popen(args, env=env, stdout=subprocess.PIPE, stderr=subprocess.PIPE)


tarzanek:
Note:
https://github.com/scylladb/scylla-ccm/commit/5b65941a2759e507a4ee88a65de0a985882ec69d
did in 2014 exact the opposite to get IPv6 test working

this change is needed for https://github.com/scylladb/scylla-tools-java/pull/100 to pass dtest

fruch:
@tarzanek can you take this ccm branch for a walk in the BYO with `-a next-gating` ?, also I don't know if this issue from 2014 is still relevant, i'll try looking for the relevant tests in dtest

tarzanek:
http://jenkins.scylladb.com/view/master/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/547/
runs now

tarzanek:
http://jenkins.scylladb.com/view/master/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/547/

tarzanek:
@fruch the jenkins job failed on something completely different - can you help me figuring out if it is relevant or how to move further?

tarzanek:
hmm rerun or Tomeks fix fixed it:
http://jenkins.scylladb.com/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/549/
@fruch is that OK with you?

fruch:
@tarzanek looks good.

I would only want one quick eye from @bhalevy or @slivne if they know if we have any ipv6 tests in dtest that might break cause of this.

tarzanek:
re: IPv6 - I think just enclosing the IP in tests into " "(or sanitizing them) would make them work, so I am puzzled the cassandra folks didn't rather fix enclosure than having a hack that depends on localhost name resolution

PR #173, opened 2019-09-01
Fix relocatable to use install.sh (fixes #169)

### LINKED ISSUE #169: Use scylla/install.sh for installing scylla binary and relcatable package
This is a followup to 468f559e248fb0144496b6807baacf4404b4a10d.

Depends on scylladb/scylla#4863

https://github.com/scylladb/scylla/pull/4863#issuecomment-522633384:
> ccm should use install.sh instead of duplicating it

fruch:
since https://github.com/scylladb/scylla/commit/698b72b5018868df6a839d08fd24c642db97ffcd#diff-3fbb47e318cd8802bd325e7da9aaabe8 
relocatable usage in ccm is broken, this PR should address it

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -5,6 +5,7 @@ import tarfile
 import tempfile
 import shutil
 import glob
+import subprocess
 
 from six import print_
 from six.moves import urllib
@@ -31,17 +32,20 @@ def setup(version, verbose=True):
         tmp_download = tempfile.mkdtemp()
 
         url = os.environ.get('SCYLLA_CORE_PACKAGE', os.path.join(s3_url, 'scylla-package.tar.gz'))
-        download_version(version, verbose=verbose, url=url, target_dir=tmp_download)
+        download_version(version, verbose=verbose, url=url, target_dir=os.path.join(tmp_download, 'scylla-core-package'))
 
         url = os.environ.get('SCYLLA_JAVA_TOOLS_PACKAGE', os.path.join(s3_url, 'scylla-tools-package.tar.gz'))
         download_version(version, verbose=verbose, url=url, target_dir=os.path.join(tmp_download, 'scylla-java-tools'))
 
         url = os.environ.get('SCYLLA_JMX_PACKAGE', os.path.join(s3_url, 'scylla-jmx-package.tar.gz'))
-        download_version(version, verbose=verbose, url=url, target_dir=os.path.join(tmp_download, 'jmx'))
+        download_version(version, verbose=verbose, url=url, target_dir=os.path.join(tmp_download, 'scylla-jmx'))
 
         cdir = directory_name(version)
 
         shutil.move(tmp_download, cdir)
+
+        # install using scylla install.sh
+        run_scylla_install_script(os.path.join(cdir, 'scylla-core-package'), cdir)
 
     return cdir, version
 
@@ -89,15 +93,6 @@ def download_version(version, url=None, verbose=False, target_dir=None):
         raise ArgumentError(msg)
     except tarfile.ReadError as e:
         raise ArgumentError("Unable to uncompress downloaded file: %s" % str(e))
-    except CCMError as e:
-        if target_dir:
-            # wipe out the directory if anything goes wrong.
-            try:
-                rmdirs(target_dir)
-                print_("Deleted %s due to error" % target_dir)
-            except:
-                raise CCMError("Downloading/extracting scylla version %s failed. Attempted to delete %s but failed. This will need to be manually deleted" % (version, target_dir))
-        raise e
 
 
 def directory_name(version):
@@ -122,3 +117,18 @@ def __get_dir():
     if not os.path.exists(repo):
         os.mkdir(repo)
     return repo
+
+
+def run_scylla_install_script(install_dir, target_dir):
+    scylla_target_dir = os.path.join(target_dir, 'scylla')
+
+    def run(cmd, cwd=None):
+        subprocess.check_call(['bash', '-c', cmd], cwd=cwd, stderr=None, stdout=None)
+
+    # FIXME: remove this hack once scylladb/scylla#4949 is fixed and merged
+    run('''sed 's|"$prefix|"$root/$prefix|' -i install.sh''', cwd=install_dir)
+
+    run('''{0}/install.sh --root {1} --target  centos --disttype redhat --pkg server'''.format(install_dir, scylla_target_dir), cwd=install_dir)
+    run('''mkdir -p {0}/conf; cp ./conf/scylla.yaml {0}/conf'''.format(scylla_target_dir), cwd=install_dir)
+    run('''ln -s {}/opt/scylladb/bin .'''.format(scylla_target_dir), cwd=target_dir)
+    run('''ln -s {}/conf .'''.format(scylla_target_dir), cwd=target_dir)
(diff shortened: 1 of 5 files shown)


ccmlib/scylla_repository.py
```diff
@@ -122,3 +126,10 @@ def __get_dir():
     if not os.path.exists(repo):
         os.mkdir(repo)
     return repo
+
+
+def run_scylla_install_script(install_dir, target_dir):
+    subprocess.check_call(['bash', '-c', ''' sed 's|"$prefix|"$root/$prefix|' -i install.sh '''], cwd=install_dir)
```
bhalevy: This is non-maintainable. Please open an issue against scylla for fixing this in `install.sh`. Does it need another option or is this valid at all times?

ccmlib/scylla_repository.py
```diff
@@ -122,3 +126,10 @@ def __get_dir():
...
+    subprocess.check_call(['bash', '-c', ''' {0}/install.sh --root {1}  --target  centos --disttype redhat --pkg server '''.format(install_dir, target_dir)], cwd=install_dir)
```
bhalevy: nit: two spaces before and after `--target``
(and also you don't really need the extra space after the 3 single quotes and before the closing 3 quotes, do you?)

ccmlib/scylla_repository.py
```diff
@@ -122,3 +126,10 @@ def __get_dir():
...
+    subprocess.check_call(['bash', '-c', 'mkdir -p {0}/conf; cp ./conf/scylla.yaml {0}/conf'.format(target_dir)], cwd=install_dir)
```
bhalevy: Nit: to be consistent with above command this might better:
```
subprocess.check_call(['bash', '-c', 'mkdir -p {1}/conf && cp {0}/conf/scylla.yaml {1}/conf/'.format(install_dir, target_dir)], cwd=install_dir)
```

ccmlib/scylla_repository.py
```diff
@@ -122,3 +126,10 @@ def __get_dir():
...
+    subprocess.check_call(['bash', '-c', 'ln -s {}/opt/scylladb/bin bin'.format(target_dir)], cwd=target_dir)
```
bhalevy: This might silently break in the future in case you already have a `./bin` directory.
How about `ln -s {}/opt/scylladb/bin .`?

ccmlib/scylla_repository.py
bhalevy: I opened scylladb/scylla#4949

bhalevy:
Also, please add example to the help message printed by `scylla_ccm_env.sh`

bhalevy:
When the `--scylla-core-package-uri` file is not found the exception is very misleading:
```
ccmlib.common.CCMError: Downloading/extracting scylla version unstable/master/380 failed. Attempted to delete /tmp/tmpHGgA3q/installer but failed. This will need to be manually deleted
```

Need to print which file was not found since the problem is not with the relocatable on s3.

bhalevy:
We also need to reorganize the the `scylla-repository` directory hierarchy:
- create directory per package: `scylla`, `scylla-tools-java`, `scylla-jmx`
  - make sure `bin` symlink is corrected too
- rename `installer` directory to `scylla-core-package`
- keep `bin` and `conf` symlinks in hierarchy root, pointing to `scylla/{bin,conf}`, respectively to be backward compatible with how these paths are used in libccm.

ccmlib/scylla_repository.py
fruch: I'll remove it, once it's merged into master

fruch:
@bhalevy, now a bit more organized:

```
total 32
drwx------  6 fruch fruch 4096 Sep  3 16:34 .
drwxr-xr-x  7 fruch fruch 4096 Sep  3 16:34 ..
lrwxrwxrwx  1 fruch fruch   78 Sep  3 16:34 bin -> /home/fruch/.ccm/scylla-repository/unstable/master/380/scylla/opt/scylladb/bin
lrwxrwxrwx  1 fruch fruch   66 Sep  3 16:34 conf -> /home/fruch/.ccm/scylla-repository/unstable/master/380/scylla/conf
drwxrwxr-x  3 fruch fruch 4096 Sep  3 16:34 jmx
drwxr-xr-x  7 fruch fruch 4096 Sep  3 16:34 scylla
drwxrwxr-x 12 fruch fruch 4096 Sep  3 16:34 scylla-core-package
drwxrwxr-x  9 fruch fruch 4096 Sep  3 16:34 scylla-java-tools
```

fruch:
@bhalevy, forgotted to change the jmx dir name, now it's better:
```
drwx------  6 fruch fruch 4096 Sep  4 09:45 .
drwxr-xr-x  8 fruch fruch 4096 Sep  4 09:51 ..
lrwxrwxrwx  1 fruch fruch   78 Sep  4 09:45 bin -> /home/fruch/.ccm/scylla-repository/unstable/master/390/scylla/opt/scylladb/bin
lrwxrwxrwx  1 fruch fruch   66 Sep  4 09:45 conf -> /home/fruch/.ccm/scylla-repository/unstable/master/390/scylla/conf
drwxr-xr-x  7 fruch fruch 4096 Sep  4 09:45 scylla
drwxr-xr-x 12 fruch fruch 4096 Sep  4 09:55 scylla-core-package
drwxr-xr-x  9 fruch fruch 4096 Sep  4 09:44 scylla-java-tools
drwxr-xr-x  3 fruch fruch 4096 Sep  4 09:45 scylla-jmx
```

ccmlib/scylla_repository.py
```diff
@@ -90,13 +94,6 @@ def download_version(version, url=None, verbose=False, target_dir=None):
     except tarfile.ReadError as e:
         raise ArgumentError("Unable to uncompress downloaded file: %s" % str(e))
     except CCMError as e:
-        if target_dir:
-            # wipe out the directory if anything goes wrong.
-            try:
-                rmdirs(target_dir)
-                print_("Deleted %s due to error" % target_dir)
-            except:
-                raise CCMError("Downloading/extracting scylla version %s failed. Attempted to delete %s but failed. This will need to be manually deleted" % (version, target_dir))
```
bhalevy: why?

ccmlib/scylla_repository.py
```diff
@@ -122,3 +119,16 @@ def __get_dir():
...
     return repo
+
+
+def run_scylla_install_script(install_dir, target_dir):
+    scylla_target_dir = os.path.join(target_dir, 'scylla')
+
+    def run(cmd, cwd=None):
+        subprocess.check_call(['bash', '-c', cmd], cwd=cwd, stderr=None, stdout=None)
+
+    run('''sed 's|"$prefix|"$root/$prefix|' -i install.sh''', cwd=install_dir)
```
bhalevy: This depends on scylladb/scylla#4949 so we should wait for the fix to get merged rather than doing this patching.

ccmlib/scylla_repository.py
fruch: this code isn't needed anymore, since we work with temp directory to begin with.
and it was only masking the error being raised

bhalevy: @fruch So you can remove the whole `except CCMError` part, no?

fruch: yep, you got a point there, removing

ccmlib/scylla_repository.py
bhalevy: If you want this to be merged before scylladb/scylla#4949 is fixed, at least add a comment mentioning the scylla issue and adding a FIXME note to remove this line when the latter is fixed.

fruch: done

PR #178, opened 2019-09-15
Force core if stop gently failed

bhalevy:
If gently stopping the node failed retry with signal.SIGQUIT

Aborting is intended to generate a core dump
so the reason the node didn't stop normally can be studied.

Refs scylladb/scylla#4384 scylladb/scylla#5020

Test: `bootstrap_test:TestBootstrap.start_stop_test`
and the same with instrumenting scylla to hang on shutdown

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -540,6 +540,22 @@ class ScyllaNode(Node):
             raise NodeError('Problem starting node %s scylla-agent due to %s' %
                             (self.name, e))
 
+    def wait_until_stopped(self, wait_seconds=127):
+        still_running = self.is_running()
+        if still_running:
+            # The sum of 7 sleeps starting at 1 and doubling each time
+            # is 2**7-1 (=127). So to sleep an arbitrary wait_seconds
+            # we need the first sleep to be wait_seconds/(2**7-1).
+            wait_time_sec = wait_seconds/(2**7-1.0)
+            for i in xrange(0, 7):
+                time.sleep(wait_time_sec)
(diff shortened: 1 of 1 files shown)


ccmlib/scylla_node.py
```diff
@@ -514,26 +529,31 @@ def stop(self, wait=True, wait_other_notice=False, other_nodes=None, gently=True
                     except OSError:
                         pass
 
+            if not wait and not wait_other_notice:
+                return True
+
+            time.sleep(1)
```
fruch: Why sleeping here ?

ccmlib/scylla_node.py
```diff
@@ -514,26 +529,31 @@ def stop(self, wait=True, wait_other_notice=False, other_nodes=None, gently=True
...
+                        except OSError:
+                            pass
+                    self.wait_until_stopped(wait_seconds)
+                if self.is_running():
+                    raise NodeError("Problem stopping node %s" % self.name)
+
             if wait_other_notice:
                 for node, mark in marks:
                     node.watch_log_for_death(self, from_mark=mark)
-            else:
```
fruch: Why switch the order ? first killing the node then waiting for others ?

bhalevy: Others won't notice if the node hangs during shutdown,
so we want to force kill it if needed and then look if the others noticed.

bhalevy:
Successful test [here](https://jenkins.scylladb.com/view/master/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/632/console)

PR #181, opened 2019-10-24
Fix relocatable packages to work with master version

fruch:
install.sh command line arguments has been change recently
and broke ccm relocatable support


fruch:
@aleksbykov confirm this is working on latest version from today

PR #183, opened 2019-10-29
Make dbuild scylla core dumps debuggable again

### LINKED ISSUE #182: core dumps of relocatable scylla cannot be debugged
Since b84d6dcb7a86bd770880d55d3dc23429329dc87d  and 468f559e248fb0144496b6807baacf4404b4a10d, when `SCYLLA_DBUILD_SO_DIR` is provided,
scylla_node.py runs scylla using the interpreter (`ld-linux-x86-64.so.2`).
This creates core dumps of the *interpreter*, not of *scylla*, and it becomes useless for thread debugging.  Instead we need to follow scylldb/scylla@698b72b5018868df6a839d08fd24c642db97ffcd and set the binary's interpreter using patchelf and run it setting the correct path to the loadable libs in LD_LIBRARY_PATH.

bhalevy:
copy and patchelf scylla if SCYLLA_DBUILD_SO_DIR is given
then run it natively, not via a shell script (`scylla.sh`), setting `LD_LIBRARY_PATH="$SCYLLA_DBUILD_SO_DIR"` in the environment

Fixes #182

Test: [byo/635](https://jenkins.scylladb.com/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/635/)

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -391,8 +391,6 @@ class ScyllaNode(Node):
         launch_bin = common.join_bin(self.get_path(), 'bin', 'scylla')
         options_file = os.path.join(self.get_path(), 'conf', 'scylla.yaml')
 
-        os.chmod(launch_bin, os.stat(launch_bin).st_mode | stat.S_IEXEC)
-
         # TODO: we do not support forcing specific settings
         # TODO: workaround for api-address as we do not load it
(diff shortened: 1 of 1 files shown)


bhalevy:
I also updated https://github.com/scylladb/scylla/wiki/Using-CCM

PR #184, opened 2019-10-30
Workaround patchelf small file limitation

bhalevy:
This series works around the 2 GB file size limitation (scylladb/scylla/issues/5245)
By:
1. running patchelf from SCYLLA_DBUILD_SO_DIR (based on scylladb/scylla-dtest@119fdb1e2bdfc01fd055968e984ca08b2abf3ed7), so when patchelf is upgraded to version 0.10 or higher in scylla dbuild docker image, we could use it and enjoy the support for large files.
2. stripping the to-be-patchelf'ed scylla binary if patchelf version < 0.10 so to reduce its size to under 2 GB.


ccmlib/scylla_node.py
```diff
@@ -710,12 +710,16 @@ def import_bin_files(self):
...
                 loader = res[0]
-                cmd = ['patchelf', '--set-interpreter', loader, dst]
-                subprocess.check_call(cmd)
 
                 self._launch_env = dict(os.environ)
                 self._launch_env['LD_LIBRARY_PATH'] = dbuild_so_dir
 
+                cmd = [loader, os.path.join(dbuild_so_dir, 'patchelf'), '--set-interpreter', loader, dst]
+                p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=self._launch_env)
+                (stdout, stderr) = p.communicate()
```
avikivity: Why aren't you using subprocess.check_call?

ccmlib/scylla_node.py
```diff
@@ -714,6 +714,10 @@ def import_bin_files(self):
                 self._launch_env = dict(os.environ)
                 self._launch_env['LD_LIBRARY_PATH'] = dbuild_so_dir
 
+                if scylla_mode == 'debug':
+                    cmd = ['strip', dst]
```
avikivity: Better to check the file size instead of the mode.

ccmlib/scylla_node.py
```diff
@@ -715,8 +716,20 @@ def import_bin_files(self):
...
+                        raise RuntimeError('{}: exited with status {}.\nstdout:{}\nstderr:\n{}'.format(cmd, p.returncode, stdout, stderr))
+                    match = re.match('^patchelf (\d)+\.(\d+)$', stdout)
+                    if match:
+                        major = int(match.group(1))
+                        minor = int(match.group(2))
+                        if major <= 0 and minor < 10:
+                            cmd = ['strip', dst]
+                            subprocess.check_call(cmd)
+                    else:
+                        print('Could not determine patchelf version: {}'.format(stdout))
```
avikivity: exit?

ccmlib/scylla_node.py
bhalevy: Because it doesn't support `env=` (needed for setting `LD_LIBRARY_PATH`, needed for running `patchelf` from the the reloc dir)

ccmlib/scylla_node.py
bhalevy: right, will fix.

ccmlib/scylla_node.py
bhalevy: The idea is to attempt to run patchelf any case and let it fail, need be.

ccmlib/scylla_node.py
avikivity: ok

fruch: btw, it's not documented very good, but you can pass env to `subprocess.check_call()`:
```
Python 2.7.15rc1 (default, Nov 12 2018, 14:31:15) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import subprocess
>>> subprocess.check_call('env', env={'fruch': 'test'})
fruch=test
0
>>> 
```

bhalevy:
In v2:
- rather than check for debug mode / file size and patchelf version, just
  try stripping the binary if patchelf failed and its stderr indicates the large file bug
  and then retry patchelf.

ccmlib/scylla_node.py
```diff
@@ -714,11 +714,19 @@ def import_bin_files(self):
                 self._launch_env = dict(os.environ)
                 self._launch_env['LD_LIBRARY_PATH'] = dbuild_so_dir
 
-                cmd = [loader, os.path.join(dbuild_so_dir, 'patchelf'), '--set-interpreter', loader, dst]
-                p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=self._launch_env)
-                (stdout, stderr) = p.communicate()
-                if p.returncode != 0:
-                    raise RuntimeError('{}: exited with status {}.\nstdout:{}\nstderr:\n{}'.format(cmd, p.returncode, stdout, stderr))
+                stripped = False
+                while True:
```
avikivity: It's nicer to extract the snippets into small local functions than to code this as a loop and force the reader to understand that it's "run at most twice".

bhalevy: ok, fixed in v3

PR #188, opened 2019-11-18
Few python3 related fixes

fruch:


fruch:
running now on BYO, for making sure no regression on python2 code:
https://jenkins.scylladb.com/job/scylla-master/job/manual-and-scheduled-tests/job/byo_build_tests_dtest/700/

fruch:
@hagitsegev @yarongilor, those fixes would also address the issue failing the python matrix

ccmlib/common.py
```diff
@@ -366,8 +366,14 @@ def isDse(install_dir):
 
 
 def isScylla(install_dir):
+    scylla_version = os.environ.get('SCYLLA_VERSION', None)
```
bhalevy: 1. Is this python3 related fix?  Seems orthogonal and so should be done in a separate patch.

2. why do it unconditionally?
can't it be done only `if install_dir is None:`

ccmlib/scylla_node.py
```diff
@@ -374,7 +374,6 @@ def start(self, join_ring=True, no_wait=False, verbose=False,
                 except Exception as msg:
                     print("{}. Looking for offending processes...".format(msg))
                     for proc in psutil.process_iter():
-                        print  proc.cmdline()
```
bhalevy: python3 related?

fruch: yes, it's breaks python, and it's not actully need (I think it's my mistake in older PR)

ccmlib/common.py
fruch: as for 1, yes it can be separated into it's own commit

fruch: as for 2, yes we could put it into the right context

fruch: opened
https://github.com/scylladb/scylla-ccm/pull/189
for this one

PR #192, opened 2019-11-29
enable_internode_ssl: support to assign internode_encryption

amoskong:
Currently we always use default 'all' for internode_encryption parameter.
This patch made this configurable, we will use it to cover more configuration in dtest.


amoskong:
We need backport this to branch-3.2

fruch:
@amoskong, back ported

amoskong:
Thanks

PR #195, opened 2019-12-02
`SCYLLA-VERSION-FILE` is in a diffrent location in reloc packages

fruch:
this isn't a big issue, but it would give us the ability to get the actual branch version.
not really helpful in master, where it's `666.development`

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -579,9 +579,11 @@ def get_version_from_build(install_dir=None, node_path=None):
 
 def get_scylla_version(install_dir):
     if isScylla(install_dir):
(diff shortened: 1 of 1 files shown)


fruch:
@bhalevy the this I think we are missing now, a clear way in dtest to differentiate enterprise.
so we could `@since('2019.1')` is working as we expect. now I'm not sure we could do those.

bhalevy:
we need a better way than comparing version strings to determine if the version we're testing supports some feature yet, or not.

fruch:
Something that isn't runtime... i.e. not stuff that I can check while scylla is up. I need to figure those before running

maybe have a feature list file ? that we'll be able to iterate on, and figure what's available ?

PR #196, opened 2019-12-06
node.py::watch_for_log(): make sure a single log line matches not mor…

### LINKED ISSUE #197: node::watch_for_log() is broken in the multi-patterns case
HEAD: af383300d28b1a5d650ec4d62d2fa713d8be5916

*Description*
There is a bug [node.py: line 426](https://github.com/scylladb/scylla-ccm/blob/next/ccmlib/node.py#L426) where an element is deleted from the list from inside the loop which iterates over this list.

This may lead to very unexpected results are described [here](https://stackoverflow.com/questions/1207406/how-to-remove-items-from-a-list-while-iterating).

Also, the seems to be a missing `break` in the code after [this](https://github.com/scylladb/scylla-ccm/blob/next/ccmlib/node.py#L428) line.

As a result the function may match the same log line for more than a single pattern and result in a very unexpected result - when you give X patterns you expect the function to return after X corresponding log lines have been seen.

vladzcloudius:
…e than a single pattern

This single liner fixes two issue at the same time.

First of all the code in question deletes from the list it iterates on which make lead to
very unexpected results.

Second, the code has a logic error as well: even if there wasn't for the bug above
the function would attempt to match the same log line to more than a single pattern
and if it matched - create multiple match results for the same log line.

It's unlikely that this was the original intention and it's likely that the idea was to
match a sinle log line at most once and the author simply forgot to put that "break".

This patch restores the sanity and the integrity of this code part.

Fixes #197

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -423,9 +423,15 @@ class Node(object):
                         m = e.search(line)
                         if m:
                             matchings.append((line, m))
+
+                            # This is bogus! - One should not modify the list from inside the loop which iterates on
+                            # that list.
+                            # However since we are going to break from the loop a few lines below that should be ok.
                             tofind.remove(e)
+
                             if len(tofind) == 0:
                                 return matchings[0] if isinstance(exprs, string_types) else matchings
+                            break
                 else:
                     # yep, it's ugly
                     time.sleep(0.01)


vladzcloudius:
@bhalevy @roydahan FYI

fruch:
@vladzcloudius I can only guess too, what the intent of whom initial wrote this function.
I think I should effect too much of the tests changing it.

@bhalevy, what do you think ?

bhalevy:
> @vladzcloudius I can only guess too, what the intent of whom initial wrote this function.
> I think I should effect too much of the tests changing it.
> 
> @bhalevy, what do you think ?

After discussion with @vladzcloudius we're in agreement that the intention of this function was probably to match all expression if given a list of expression.
It is possible to implement it in such away that it continues to match other expressions against an already matches log line, but this is not really required. A caller that wants "any of" semantics can pass a regular expression like `<expr i>|<expr j>|...`

PR #199, opened 2019-12-15
node cleanups: update_config, new_node, update_pid

bhalevy:
Fixes https://github.com/scylladb/scylla-dtest/issues/1189

We hit the following path:
```
Stacktrace
  File "/usr/lib64/python2.7/unittest/case.py", line 422, in doCleanups
    function(*args, **kwargs)
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-dtest/update_cluster_layout_tests.py", line 234, in stop_node3
    node3.stop()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/scylla_node.py", line 602, in stop
    if not self.wait_until_stopped(wait_seconds):
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/scylla_node.py", line 543, in wait_until_stopped
    if not self.is_running():
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/node.py", line 294, in is_running
    self.__update_status()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/node.py", line 1563, in __update_status
    self._update_config()
  File "/jenkins/workspace/scylla-master/dtest-release/scylla-ccm/ccmlib/node.py", line 1349, in _update_config
    os.mkdir(dir_name)
OSError: [Errno 2] No such file or directory: '/jenkins/workspace/scylla-master/dtest-release/scylla/.dtest/dtest-3MjM7K/test/node3'

Standard Output
Timed out waiting for pidfile to be filled (current time is 2019-12-06 03:53:35.094448)
```

This patchset fixes several issues related to https://github.com/scylladb/scylla-dtest/issues/1189:
1. do not call mkdir in update_config, but rather explicitly, when needed.
2. introduce cluster.new_node interface so that dtests won't second guess the args for create_node (ipaddr/port interfaces in particular)
3. print more information when timed out waiting for pidfile
    and bail out early if not waiting in `ScyllaNode._update_jmx_pid`

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -20,6 +20,7 @@ class Cluster(object):
         self.name = name
         self.id = 0
         self.ipprefix = None
+        self.ipformat = None
         self.nodes = {}
         self.seeds = []
         self.partitioner = partitioner
@@ -148,9 +149,15 @@ class Cluster(object):
         node._save()
         return self
 
-    def populate(self, nodes, debug=False, tokens=None, use_vnodes=False, ipprefix='127.0.0.', ipformat=None):
-        if self.ipprefix:
-            ipprefix = self.ipprefix
+    def populate(self, nodes, debug=False, tokens=None, use_vnodes=False, ipprefix=None, ipformat=None):
+        if ipprefix:
+            self.ipprefix = ipprefix
+        elif not self.ipprefix:
+            self.ipprefix = '127.0.0.'
+        if ipformat:
+            self.ipformat = ipformat
+        elif not self.ipformat:
+            self.ipformat = self.ipprefix + "%d"
         node_count = nodes
         dcs = []
         self.use_vnodes = use_vnodes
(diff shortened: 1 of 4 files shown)


ccmlib/node.py
```diff
@@ -1349,9 +1349,11 @@ def _save(self):
         self.__update_envfile()
         self._update_config()
 
-    def _update_config(self):
+    def _update_config(self, on_stop=False):
         dir_name = self.get_path()
         if not os.path.exists(dir_name):
+            if on_stop:
+                return
             os.mkdir(dir_name)
```
fruch: won't be easier to move the os.mkdir to a better place ? i.e. not part of update_config ?

fruch:
I think we should just call
https://docs.python.org/3/library/unittest.html#unittest.TestCase.doCleanups
at the begining of `tearDown()`, since the issue is that cleanup are called after teardowns

bhalevy:
> I think we should just call
> https://docs.python.org/3/library/unittest.html#unittest.TestCase.doCleanups
> at the begining of `tearDown()`, since the issue is that cleanup are called after teardowns

We call `cleanUpCluster` at the end of `Tester.tearDown` and also register it with `addCleanup`.
doCleanups itself is called after tearDown by case.py (at least in the python2.7 version) so I don't think that changing that design is a good idea.

What's the problem with having node.stop be resilient to an already stopped node (and removed from the file system)

ccmlib/node.py
bhalevy: maybe, it'd cleaner, but it might be a bigger change.
I'll look into it.

bhalevy:
BYO test: https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/21

fruch:
@bhalevy you might want to rename this PR, since it's doing much more than the title says.
But :+1: for new add_node cleanup, the dtest PR look good too (once we merge this one)

bhalevy:
BYO test: https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/24

ccmlib/scylla_node.py
```diff
@@ -505,12 +506,19 @@ def _update_jmx_pid(self, wait=True):
...
             if time.time() - start > 30.0 or not wait:
-                print_("Timed out waiting for pidfile to be filled "
-                       "(current time is %s)" % (datetime.datetime.now()))
-                return
+                print_("Timed out waiting for pidfile {} to be filled (after {} seconds): File {} size={}".format(
+                        pidfile,
+                        0 if not wait else time.time() - start,
+                        'exists' if os.path.isfile(pidfile) else 'does not exist' if not os.path.exists(pidfile) else 'is not a file',
+                        os.stat(pidfile).st_size if os.path.exists(pidfile) else -1))
+                break
```
fruch: the switch from `return` to `break`, is that on purpose here ?

ccmlib/cluster.py
```diff
@@ -148,9 +149,15 @@ def add(self, node, is_seed, data_center=None):
         node._save()
         return self
 
-    def populate(self, nodes, debug=False, tokens=None, use_vnodes=False, ipprefix='127.0.0.', ipformat=None):
-        if self.ipprefix:
-            ipprefix = self.ipprefix
+    def populate(self, nodes, debug=False, tokens=None, use_vnodes=False, ipprefix=None, ipformat=None):
```
fruch: one of the test you've run, run cassandra ? to see we didn't broke that ?

bhalevy: I'm not sure the BYO ran cassandra cluster but there is no change in observable behavior (at least there shouln't be)

ccmlib/scylla_node.py
bhalevy: Yes, since the `wait` case is handled outside the loop now, and the timeout case is handled consistently like the other similar loops.

ccmlib/cluster.py
bhalevy: @fruch validated with `migration_test:TTLWithMigrate.big_table_with_ttls_test`

PR #201, opened 2019-12-18
scylla_node: support SCYLLA_EXT_ENV

bhalevy:
Allow passing extra environment variables for running scylla.
Those are represented in a single string comprised of one or more
pairs of "var=value" separated by either space or semicolon (';')

The planned use case is for setting the environment for ASAN
to dump core on error when running dtest in debug mode.

Refs https://github.com/scylladb/scylla-dtest/issues/1186

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -15,6 +15,7 @@ import threading
 import psutil
 import yaml
 import glob
+import re
 
 from six import print_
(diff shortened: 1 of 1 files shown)


bhalevy:
Successful BYO run: https://github.com/scylladb/scylla-dtest/issues/1186

bhalevy:
Cc @espindola

PR #202, opened 2019-12-25
Standardize on sleep time of 0.1 second.

kostja:
Remove too long and too short sleeps. Remove unchecked sleeps.
This change speeds up dtest by ~20%.

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -288,7 +288,7 @@ class Cluster(object):
                     removed = True
                 except:
                     tries = tries + 1
-                    time.sleep(.1)
+                    time.sleep(0.1)
                     if tries == 5:
                         raise
 
@@ -337,7 +337,7 @@ class Cluster(object):
                 started.append((node, p, mark))
 
         if no_wait and not verbose:
-            time.sleep(2)  # waiting 2 seconds to check for early errors and for the pid to be set
+            time.sleep(0.1)  # waiting 0.1 seconds to check for early errors and for the pid to be set
         else:
             for node, p, mark in started:
                 try:
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -395,7 +395,7 @@ class Node(object):
 
         log_file = os.path.join(self.get_path(), 'logs', filename)
         while not os.path.exists(log_file):
-            time.sleep(.1)
+            time.sleep(0.1)
             if process:
                 process.poll()
                 if process.returncode is not None:
@@ -436,9 +436,9 @@ class Node(object):
                             break
                 else:
                     # yep, it's ugly
-                    time.sleep(0.01)
-                    elapsed = elapsed + 1
-                    if elapsed > 100 * timeout:
+                    time.sleep(0.1)
+                    elapsed = elapsed + 0.1
+                    if elapsed > timeout:
                         raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) + " [" + self.name + "] Missing: " + str(
                             [e.pattern for e in tofind]) + ":\n" + reads[:50] + ".....\nSee {} for remainder".format(filename))
 
@@ -671,15 +671,15 @@ class Node(object):
                 for node, mark in marks:
                     node.watch_log_for_death(self, from_mark=mark)
             else:
-                time.sleep(.1)
+                time.sleep(0.1)
 
             still_running = self.is_running()
             if still_running and wait:
                 # The sum of 7 sleeps starting at 1 and doubling each time
                 # is 2**7-1 (=127). So to sleep an arbitrary wait_seconds
                 # we need the first sleep to be wait_seconds/(2**7-1).
-                wait_time_sec = wait_seconds/(2**7-1.0)
-                for i in xrange(0, 7):
+                wait_time_sec = 0.1
+                while wait_time_sec <= wait_seconds + 0.1:
                     # we'll double the wait time each try and cassandra should
                     # not take more than 1 minute to shutdown
                     time.sleep(wait_time_sec)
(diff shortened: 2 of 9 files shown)


ccmlib/node.py
```diff
@@ -671,15 +671,15 @@ def stop(self, wait=True, wait_other_notice=False, other_nodes=None, gently=True
...
 
             still_running = self.is_running()
             if still_running and wait:
                 # The sum of 7 sleeps starting at 1 and doubling each time
                 # is 2**7-1 (=127). So to sleep an arbitrary wait_seconds
                 # we need the first sleep to be wait_seconds/(2**7-1).
-                wait_time_sec = wait_seconds/(2**7-1.0)
-                for i in xrange(0, 7):
+                wait_time_sec = 0.1
+                while wait_time_sec <= wait_seconds + 0.1:
```
slivne: we are not really forcing every iteration to 0.1 seconds in this case since wait_time_sec is growing ..... but fine

ccmlib/node.py
```diff
@@ -671,15 +671,15 @@ def stop(self, wait=True, wait_other_notice=False, other_nodes=None, gently=True
                 for node, mark in marks:
                     node.watch_log_for_death(self, from_mark=mark)
             else:
-                time.sleep(.1)
+                time.sleep(0.1)
 
             still_running = self.is_running()
             if still_running and wait:
                 # The sum of 7 sleeps starting at 1 and doubling each time
```
slivne: remove comments if the algo changed

ccmlib/scylla_cluster.py
```diff
@@ -96,13 +96,13 @@ def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False,
                                profile_options=profile_options)
                 # Let's ensure the nodes start at different times to avoid
                 # race conditions while creating system tables
-                time.sleep(1)
+                time.sleep(0.1)
```
slivne: this is without a loop its changing the behavior - not sure it will not cause tests to fail on slow machines

ccmlib/scylla_cluster.py
```diff
@@ -96,13 +96,13 @@ def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False,
...
                 started.append((node, p, mark))
 
         if no_wait and not verbose:
             # waiting 2 seconds to check for early errors and for the
             # pid to be set
-            time.sleep(2)
+            time.sleep(0.1)
```
slivne: this is without a loop its changing the behavior - not sure it will not cause tests to fail on slow machines

ccmlib/scylla_node.py
```diff
@@ -242,7 +242,7 @@ def _start_scylla(self, args, marks, update_pid, wait_other_notice,
         if wait_for_binary_proto:
             self.wait_for_binary_interface(from_mark=self.mark, process=self._process_scylla)
         else:
-            time.sleep(2)
+            time.sleep(0.1)
```
slivne: this is without a loop its changing the behavior - not sure it will not cause tests to fail on slow machines

ccmlib/scylla_node.py
```diff
@@ -568,7 +568,7 @@ def _update_scylla_agent_pid(self):
 
     def wait_until_stopped(self, wait_seconds=127):
         start_time = time.time()
-        wait_time_sec = 1
+        wait_time_sec = 0.1
```
slivne: this will not wait 0.1 in each iteration it grows exponentially - its fine, maybe the correct solution is not to have the growth so high .... mac 1 sec

ccmlib/cluster.py
```diff
@@ -337,7 +337,7 @@ def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False, wait_
                 started.append((node, p, mark))
 
         if no_wait and not verbose:
-            time.sleep(2)  # waiting 2 seconds to check for early errors and for the pid to be set
+            time.sleep(0.1)  # waiting 0.1 seconds to check for early errors and for the pid to be set
```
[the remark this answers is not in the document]
kostja: Shlomi, there should be no unconditional sleeps at all. Any unconditional sleep is a guarantee of a flaky test.

ccmlib/scylla_node.py
kostja: This is a nowait branch. There should be no sleep at all. Whoever added it could have had any motives, but we wouldn't know if it is needed at all without removing it.

ccmlib/scylla_cluster.py
kostja: This  is nowait branch - I can just as well remove the sleep altogether

ccmlib/scylla_cluster.py
kostja: The comment says it all - it is "to avoid race conditions when creating system tables". I can check with Tomek if there are any race conditions, but in any case, if there are, a sleep won't help. The test will be flaky anyway.

ccmlib/scylla_cluster.py
slivne: AFAIK no_wait means you do not wait for cql port to be open ..., so you can't remove the sleep

ccmlib/scylla_node.py
slivne: AFAIK no_wait means you do not wait for cql port to be open ..., so you can't remove the sleep

ccmlib/scylla_node.py
```diff
@@ -661,17 +667,17 @@ def import_config_files(self):
         self.__update_yaml()
         self.__copy_logback_files()
 
-    def get_tool_java_dir(self):
+    def get_tools_java_dir(self):
```
fruch: how those name change are related to this PR ?

ccmlib/scylla_repository.py
```diff
@@ -49,10 +49,10 @@ def setup(version, verbose=True):
         download_version(version, verbose=verbose, url=url, target_dir=os.path.join(
             tmp_download, 'scylla-core-package'))
 
-        url = os.environ.get('SCYLLA_JAVA_TOOLS_PACKAGE', os.path.join(
+        url = os.environ.get('SCYLLA_TOOLS_JAVA_PACKAGE', os.path.join(
```
fruch: please don't change those names, it's use in multiple other repos already.
Now this PR can run in BYO...

PR #206, opened 2020-01-10
Naming

kostja:
Consistent naming for scylla-tools-java, and use scylla-tools-java environment variable in get_stress_bin()

### DIFF
--- a/ccmlib/cmds/cluster_cmds.py
+++ b/ccmlib/cmds/cluster_cmds.py
@@ -126,7 +126,7 @@ class ClusterCreateCmd(Cmd):
         parser.add_option('--scylla-core-package-uri', type="string", dest="scylla_core_package_uri",
                           help="The path scylla relocatable package", default=None)
 
-        parser.add_option('--scylla-java-tools-package-uri', type="string", dest="scylla_java_tools_package_uri",
+        parser.add_option('--scylla-tools-java-package-uri', type="string", dest="scylla_tools_java_package_uri",
                           help="The path scylla java tools relocatable package", default=None)
 
         parser.add_option('--scylla-jmx-package-uri', type="string", dest="scylla_jmx_package_uri",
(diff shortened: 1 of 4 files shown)


ccmlib/cmds/cluster_cmds.py
```diff
@@ -188,8 +188,8 @@ def validate(self, parser, options, args):
 
         if options.scylla_core_package_uri:
             os.environ['SCYLLA_CORE_PACKAGE'] = options.scylla_core_package_uri
-        if options.scylla_java_tools_package_uri:
-            os.environ['SCYLLA_JAVA_TOOLS_PACKAGE'] = options.scylla_java_tools_package_uri
+        if options.scylla_tools_java_package_uri:
+            os.environ['SCYLLA_TOOLS_JAVA_PACKAGE'] = options.scylla_tools_java_package_uri
```
fruch: This environment variable is using in at least 6 other repos, can't merge this without being backward compatible.

kostja: Why not fix all of them? It's a trivial change. Backward compatibility is technical debt. There is already 4 ways to provide tools-java package, and backward compatibility will add the fifth one. I can keep the old variable, but this upsets the purpose of the patch. The goal of the patch is to make the code easier to follow, not harder.

kostja: For dtest, I made a matching pull request to fix.
And as I have been suggesting earlier, if scylla-ccm was a submodule, there would have been no issue.

fruch: * https://github.com/scylladb/scylla-pkg/blob/a196f0b3187b95602fb606200f837308699549d5/scripts/jenkins-pipelines/groovy_modules/test.groovy#L136
* https://github.com/scylladb/scylla-java-driver-matrix/blob/0eb756c877fe9047de3fb27090f948854bfe948c/scripts/run_test.sh#L100
* https://github.com/scylladb/python-driver-matrix/blob/master/scripts/run_test.sh#L86
* https://github.com/scylladb/scylla-cassandra-unit-tests/blob/master/scripts/run_test.sh#L106

fruch: The main issue is that we need to do all of them in sync, hence why I prefer it to be backward compatible first, so we can merge the ccm change, and then the others.
and when all are done, remove the unneeded code.

PR #207, opened 2020-01-13
node: watch_log_for_death: fix regex

bhalevy:
`[dead|DOWN]` matches any letter in the brackets.
The correct regular expression is `(dead|DOWN)` that
matches either of the strings `dead` or `DOWN`.


bhalevy:
@fruch please review

fruch:
@bhalevy I'm not sure I want to ask how you figure this one out... as the regex joke goes you we have :two: problems :+1:

PR #209, opened 2020-01-14
Standardize on sleep time of 0.1 second.

kostja:
Remove too long and too short sleeps in any kind
of a loop.

For now, preserve most of unchecked sleeps at their
current values for backward compatibility.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -395,7 +395,7 @@ class Node(object):
 
         log_file = os.path.join(self.get_path(), 'logs', filename)
         while not os.path.exists(log_file):
-            time.sleep(.1)
+            time.sleep(0.1)
             if process:
                 process.poll()
                 if process.returncode is not None:
@@ -436,9 +436,9 @@ class Node(object):
                             break
                 else:
                     # yep, it's ugly
-                    time.sleep(0.01)
-                    elapsed = elapsed + 1
-                    if elapsed > 100 * timeout:
+                    time.sleep(0.1)
+                    elapsed = elapsed + 0.1
+                    if elapsed > timeout:
                         raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) + " [" + self.name + "] Missing: " + str(
                             [e.pattern for e in tofind]) + ":\n" + reads[:50] + ".....\nSee {} for remainder".format(filename))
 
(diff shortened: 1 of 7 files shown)


fruch:
taking it for a ride in CI:
https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/dtest-byo-parallel/89/

fruch:
@kostja 

https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/dtest-byo-parallel/89/testReport/junit/cql_tests/TruncateTester/dtest___split0___truncate_after_restart_test/

this test seem to alway pass quite steady on next:
https://jenkins.scylladb.com/view/nexts/job/scylla-master/job/next/test_results_analyzer/

can you please take a look, why ?

kostja:
> @kostja
> 
> https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/dtest-byo-parallel/89/testReport/junit/cql_tests/TruncateTester/dtest___split0___truncate_after_restart_test/
> 
> this test seem to alway pass quite steady on next:
> https://jenkins.scylladb.com/view/nexts/job/scylla-master/job/next/test_results_analyzer/
> 
> can you please take a look, why ?

Doesn't fail here. Let's restart the build and see if it fails again. And then I wlll take a look at the log, maybe it gives some insight.

ccmlib/node.py
```diff
@@ -671,20 +671,19 @@ def stop(self, wait=True, wait_other_notice=False, other_nodes=None, gently=True
...
-                for i in xrange(0, 7):
+                wait_time_sec = 0.1
+                elapsed = 0.0
+                while elapsed <= wait_seconds:
                     # we'll double the wait time each try and cassandra should
                     # not take more than 1 minute to shutdown
                     time.sleep(wait_time_sec)
                     if not self.is_running():
                         return True
+                    elapsed += wait_time_sec
```
bhalevy: instead of calculating it, we can compare to a `start_time = time.time()`
With this in mind, the last sleep period can be trimmed to `start_time + wait_seconds`, otherwise, since it's double in every iteration, it can overshoot quite a bit.

bhalevy: See `wait_until_stopped`

ccmlib/scylla_node.py
```diff
@@ -318,7 +318,7 @@ def _wait_java_up(self, data):
                 s.close()
             except:
                 pass
-            time.sleep(1)
+            time.sleep(0.1)
```
nyh, line 321: Not only was 1 second sleep too much here (this wait is usually around 250ms on my laptop), this silly loop did an extra sleep after succeeding! The sleep should be where the pass is :-) @Orenef11 just re-discovered this problem and fixed it in PR https://github.com/scylladb/scylla-ccm/pull/356

bhalevy:
Can this PR be closed?

bhalevy:
> Can this PR be closed?

^?

fruch:
> Can this PR be closed?

Yes, I don't think any of those issues tried to be handled here are relevant, if they are, it doesn't bother anyone enough to have them fix and tested

kostja:
I think scylla-ccm needs to avoid sleeps. I am not ready to spend my time on this personally after seeing how much needs to be done to make it happen, but I think whoever is using scylla-ccm should be bothered by it.

PR #210, opened 2020-01-19
Increase debug blocked reactor notify ms

bhalevy:
Timing in debug mode is much slower than release/dev modes and
the default reactor stall threshold of 200 ms produces
a lot of false positives. Therefore, increase it in this case using
the `--blocked-reactor-notify-ms` option to 5 seconds.

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -411,8 +411,8 @@ def scylla_extract_install_dir_and_mode(install_dir):
     elif install_dir.endswith('build/release') or install_dir.endswith('build/release/'):
         scylla_mode = 'release'
         install_dir = str(os.path.join(install_dir, os.pardir, os.pardir))
-    elif os.path.exists(os.path.join(install_dir, 'scylla-core-package', 'SCYLLA-RELOCATABLE-FILE')):
-        scylla_mode = 'reloc'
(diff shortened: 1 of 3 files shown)


ccmlib/common.py
```diff
@@ -399,20 +399,25 @@ def isOpscenter(install_dir):
     opscenter_script = os.path.join(bin_dir, 'opscenter')
     return os.path.exists(opscenter_script)
 
+def scylla_extract_mode(path):
+    m = re.match('(.*/)?build/(\w+)(/|$)', path)
```
fruch: can you add a few examples here ? i.e. one url one file path ?

bhalevy: Added in v2

bhalevy:
In v2:
- extended scylla_extract_mode to support also `scylla(-{mode})?-package.tar.gz` urls.
- added path/url examples to scylla_extract_mode

bhalevy:
BYO: https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/59
(with debug printouts: https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/60/consoleFull)

bhalevy:

PR #216, opened 2020-02-02
Force wait for cluster start

bhalevy:
Infrastructure for changing the dtest default to `wait_for_binary_proto=True` and wait_other_notice=True` when starting cluster and single nodes.


bhalevy:
Successful byo-dtest (gating) run: https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/84

w/ https://github.com/scylladb/scylla-dtest/pull/1284

PR #218, opened 2020-02-04
scylla_cluster: handle no_wait flag

kbr-:
The `no_wait` parameter in `start_nodes` was not used anywhere.
This commit passes the flag to `node.start(...)` and changes the default
value to `True`, because that's what `node.start(...)` previously used.


bhalevy:
Please rebase on top of e469c28a96a228b8bf19eaf7dc0da249f965b57b (currently in next)

PR #219, opened 2020-02-10
ccmlib/node: grep_log, grep_log_for_errors: add optional from_mark parameter

bhalevy:
Currently the usage of log marker is quite messy:
- `grep_log` does not support optional from_mark parameter, like other similar functions.
- `grep_log_for_errors` considers only self.error_mark, if set via `mark_log_for_errors`.
- Other similar methods: watch_log_for and those who call it support optional from_mark parameter.

This change unifies the handling of log marker, so that all realted methods accept
and option `from_mark` parameter and consider it if set.
In addition grep_log_for_errors still considers also self.error_mark, if set.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -312,7 +312,7 @@ class Node(object):
     def debuglogfilename(self):
         return os.path.join(self.get_path(), 'logs', 'debug.log')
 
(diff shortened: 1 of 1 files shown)


bhalevy:
This is needed to fix flakiness in dtests like `replace_first_boot_test`

PR #225, opened 2020-03-03
ccmlib/scylla_node: _start_scylla: wait for repair

bhalevy:
_start_scylla with `wait_for_binary_proto=True` may timeout
in debug mode when a node bootstraps with repair-based bootstrap.

Tolerate that by tracking `repair - Repair \d+ out of \d+ ranges`
log messages, to wait while reapir makes progress.

Allow 120 seconds of inactivity.

Tested with materialized_views_test.py:TestMaterializedViews.mv_populating_from_existing_data_during_node_decommission_test in debug mode

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -23,6 +23,7 @@ from six.moves import xrange
 from ccmlib import common
 from ccmlib.node import Node
 from ccmlib.node import NodeError
+from ccmlib.node import TimeoutError
 
 
 def wait_for(func, timeout, first=0.0, step=1.0, text=None):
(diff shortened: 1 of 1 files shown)


bhalevy:
@fruch please review

ccmlib/scylla_node.py
```diff
@@ -250,7 +251,23 @@ def _start_scylla(self, args, marks, update_pid, wait_other_notice,
                 node.watch_log_for_alive(self, from_mark=mark)
 
         if wait_for_binary_proto:
-            self.wait_for_binary_interface(from_mark=self.mark, process=self._process_scylla)
+            try:
+                self.wait_for_binary_interface(from_mark=self.mark, process=self._process_scylla)
+            except TimeoutError as e:
+                if not self.grep_log('storage_service - JOINING: getting bootstrap token', from_mark=self.mark):
```
fruch: why not move this logic into it's own function ? `wait_for_bootstrap_repair`  ? I'm guessing we'll have more place it would be needed...

bhalevy: we can do it when needed, but no problem making it a method in this change

fruch: I think it would be a bit more clearer, the start/stop functions already has lots of logic in them already.

PR #227, opened 2020-03-15
ccmlib/scylla_node: _start_scylla: fix expected bootstrap message

bhalevy:
Current message introduced in #225 is not always printed.
For example, in:
  https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-debug-random/90/artifact/logs-debug.2/1584197488046_replace_address_test.TestReplaceAddress.replace_shutdown_node_test/node4.log
`JOINING: Starting to bootstrap` is always printed by `storage_service::bootstrap`
so use it instead.


bhalevy:
@fruch please review

PR #228, opened 2020-03-16
scylla_node: increase wait_for_binary_interface timeout

bhalevy:
Accomodate for the default ring_delay_ms timeout of 300 seconds.


bhalevy:
@fruch please review

PR #229, opened 2020-03-23
fix(scylla_manager): set segments_per_repair=16 in manager yaml

ShlomiBalalis:
In an effort to make manager repairs faster, I set segments_per_repair to 16 in scylla_manager.yaml
### **Backport to manager 2.0 branch**


ShlomiBalalis:
Solves https://github.com/scylladb/mermaid/issues/1817

PR #231, opened 2020-03-30
Support enabling alternator

fruch:
since in ccm we listen of specific address, we need to make sure
alternator is listening on the same address

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -950,6 +950,8 @@ class ScyllaNode(Node):
                 except KeyError:
                     data[name] = full_options[name]
 
+        if 'alternator_port' in data or 'alternator_https_port' in data:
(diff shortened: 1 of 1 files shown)


fruch:
would need to backport to branch-4.0

ccmlib/scylla_node.py
```diff
@@ -950,6 +950,8 @@ def __update_yaml(self):
                 except KeyError:
                     data[name] = full_options[name]
 
+        if 'alternator_port' in data or 'alternator_https_port' in data:
```
yarongilor: it could also be considered to have a property of is_alternator_enabled. or a list of all enabled features (but that may be out of ccm scope).

fruch: you are right, this is not the scope of this PR, and not CCM, AFAIK.

PR #232, opened 2020-04-07
node.py: Fix node:watch_log_for to handle multiple expressions matchi…

elcallio:
…ng same line

Original impl did not handle doing something like:

watch_log_for([".*Prepare.*kossa.*", ".*Prepare.*mu.*"])

and the log producing either

[...] Prepare kossa, elk, ninja
[...] Prepare mu, tuta

or

[...] Prepare kossa, elk, mu

i.e. expressions that match same or different lines, depending on
timing.

This will run each match on every line instead, which seems more
correct.

See /scylladb/scylla#6067

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -426,19 +426,16 @@ class Node(object):
                 line = f.readline()
                 if line:
                     reads = reads + line
-                    for e in tofind:
+                    search = tofind
+                    for e in search:
                         m = e.search(line)
                         if m:
                             matchings.append((line, m))
-
-                            # This is bogus! - One should not modify the list from inside the loop which iterates on
-                            # that list.
-                            # However since we are going to break from the loop a few lines below that should be ok.
                             tofind.remove(e)
 
-                            if len(tofind) == 0:
-                                return matchings[0] if isinstance(exprs, string_types) else matchings
-                            break
+                    if len(tofind) == 0:
+                        return matchings[0] if isinstance(exprs, string_types) else matchings
+                    break
                 else:
                     # yep, it's ugly
                     time.sleep(0.01)


ccmlib/node.py
```diff
@@ -426,19 +426,16 @@ def watch_log_for(self, exprs, from_mark=None, timeout=600, process=None, verbos
...
-                            # that list.
-                            # However since we are going to break from the loop a few lines below that should be ok.
                             tofind.remove(e)
 
-                            if len(tofind) == 0:
-                                return matchings[0] if isinstance(exprs, string_types) else matchings
-                            break
+                    if len(tofind) == 0:
+                        return matchings[0] if isinstance(exprs, string_types) else matchings
+                    break
```
bhalevy, line 438: This break previously applied to the inner search loop, now it will break from outer loop on the first line, no?

vladzcloudius:
@bhalevy Yes, you are right. This fundamentally changes the semantics of `wait_log_for()`. In particular it will be impossible to use a single call for this function when we wait for multiple appearances of the same line like we do in `hintedhandoff_removenode_test()`.

In particular it I don't see how it would be possible to check such an event after this patch.

What we see in the patch description is exactly what you expect when you want to continue after all two patterns have been seen. 

If we want to match all patterns to all log lines then we need to return the `mark` where each match has been detected. - This way we would be able to solve the problem I described above.

ccmlib/node.py
elcallio: Yes. The break marks that we've found all expected matches. Instead of iterating and modifying the same set, we now copy the search and use "tofind" (original) only as carrier of matches remaining.

elcallio:
> 
> 
> @elcallio @bhalevy Yes, you are right. This fundamentally changes the semantics of `wait_log_for()`. In particular it will be impossible to use a single call for this function when we wait for multiple appearances of the same line like we do in `hintedhandoff_removenode_test()`.
> 
> In particular it I don't see how it would be possible to check such an event after this patch.
> 
So, basically the inverse of what my perceived use case was?
The obvious answer would be "loop repeat `wait_log_for`, but then the obvious retort is "you can implement the same tactics for the match-same-line use case".

Perhaps simply add a default-no parameter that determines match mode?

ccmlib/node.py
elcallio: No, wait, now I got your question. This looks wrong. Hang on...

vladzcloudius:
> > @elcallio @bhalevy Yes, you are right. This fundamentally changes the semantics of `wait_log_for()`. In particular it will be impossible to use a single call for this function when we wait for multiple appearances of the same line like we do in `hintedhandoff_removenode_test()`.
> > In particular it I don't see how it would be possible to check such an event after this patch.
> 
> So, basically the inverse of what my perceived use case was?
> The obvious answer would be "loop repeat `wait_log_for`, but then the obvious retort is "you can implement the same tactics for the match-same-line use case".
> 
> Perhaps simply add a default-no parameter that determines match mode?

@elcallio The semantics of `wait_log_for()` today is that if is going to actually wait till specified lines appear in the log.

What you suggested before can be achieved by `grep_log()` from `node.py` and the whole reason for existence of this `wait_log_for()` is to avoid these loops.

elcallio:
Yeah, I conclude modifying this is just gonna be error prone and cause conflict. I solved it in test instead. A little more expensive, but good enough.

PR #233, opened 2020-04-27
node: get_sstables: cleanup unsealed sstables

bhalevy:
To cleanup temporary, unsealed SSTables.

Fixes https://github.com/scylladb/scylla/issues/6211


bhalevy:
@fruch can you please review?

PR #236, opened 2020-05-25
ccmlib/scylla_repository: support relocatable package format v2

syuu1228:
Since we modified relocatable package structure, scylla-ccm need to
detect and handle different relocatable package format versions.
To do this easily, we added ".relocatable_package_version" file on
relocatable package.
If .relocatable_package_version exists and it contains "2",
all files under the package is located "scylla*/" subdirectory.
If not, it's legacy format package.

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -7,6 +7,8 @@ import shutil
 import subprocess
 import re
 import gzip
+import sys
+import glob
 
 from subprocess import Popen, PIPE, STDOUT
 
@@ -68,7 +70,6 @@ def setup(version, verbose=True):
         # install using scylla install.sh
         run_scylla_install_script(os.path.join(
             cdir, 'scylla-core-package'), cdir)
-
     setup_scylla_manager()
 
     return cdir, get_scylla_version(cdir)
@@ -172,6 +173,22 @@ def download_version(version, url=None, verbose=False, target_dir=None):
         tar = tarfile.open(target)
         tar.extractall(path=target_dir)
         tar.close()
+
+        # if relocatable package format >= 2, need to extract files under subdir
+        package_version_file = "{}/.relocatable_package_version".format(target_dir)
+        if os.path.exists(package_version_file):
+            with open(package_version_file) as f:
+                package_version = f.read().strip()
+            if package_version != '2':
+                print('Unknown relocatable package format version: ' + package_version)
+                sys.exit(1)
+            print('Relocatable package format version 2 detected.')
+            pkg_dir = glob.glob('{}/*/'.format(target_dir))[0]
+            shutil.move(str(pkg_dir), target_dir + '.new')
+            shutil.rmtree(target_dir)
+            shutil.move(target_dir + '.new', target_dir)
+        else:
+            print('Legacy relocatable package format detected.')
 
         # add breadcrumb so we could list the origin of each part easily for debugging
         # for example listing all the version we have in ccm scylla-repository


ccmlib/scylla_repository.py
```diff
@@ -173,6 +174,25 @@ def download_version(version, url=None, verbose=False, target_dir=None):
...
+        package_version_file = "{}/.relocatable_package_version".format(target_dir)
+        if os.path.exists(package_version_file):
+            with open(package_version_file) as f:
+                package_version = f.read().strip()
+            if package_version != '2':
+                print('Unknown relocatable package format version: ' + package_version)
+                sys.exit(1)
+            print('Relocatable package format version 2 detected.')
+            for p in Path(target_dir).iterdir():
+                if p.is_dir():
```
fruch, line 187: there only on directory inside of it ? doesn't it has a specific name ?

ccmlib/scylla_repository.py
```diff
@@ -173,6 +174,25 @@ def download_version(version, url=None, verbose=False, target_dir=None):
...
+                    pkg_dir = p
+                    break
+            shutil.move(str(pkg_dir), target_dir + '.new')
+            shutil.rmtree(target_dir)
+            shutil.move(target_dir + '.new', target_dir)
```
fruch, line 192:
```suggestion
            shutil.move(str(pkg_dir), target_dir + '.tmp')
            shutil.rmtree(target_dir)
            shutil.move(target_dir + '.tmp', target_dir)
```

ccmlib/scylla_repository.py
```diff
@@ -7,11 +7,13 @@
 import subprocess
 import re
 import gzip
+import sys
 
 from subprocess import Popen, PIPE, STDOUT
 
 import hashlib
 import requests
+from pathlib import Path
```
fruch, line 16: this breaks python2 compatibility, there are still a few places which are running with python2.
can you switch to `os.listdir` or add the backport `pathlib2` to the the `setup.py`

fruch: see:
https://www.python.org/dev/peps/pep-0508/
for how to add it to setup.py for version < 3.2

ccmlib/scylla_repository.py
syuu1228: @fruch The new format contain only single directory which has same name as package (scylla-package.tar.gz -> scylla, scylla-jmx-package -> scylla-jmx ...and so on).
Why do we need this is, we want to combine all tarballs to single, unified tarball without file conflict:
https://github.com/scylladb/scylla-pkg/issues/1218
Since each package has different directory name, this script trying to find a directory in indirect way.

There is another approach, we may able to add another file to store the directory name (= package name) just like ".relocatable_package_version" file, then we don't need to do iterdir().

ccmlib/scylla_repository.py
syuu1228: Ah sorry, didn't realize we need python2 compatibility here.
I can drop pathlib then.

avikivity: Where do we use python2? Can we drop it? it's dead. An ex-python.

syuu1228: Pushed python2 compatible version.

fruch: @avikivity it's dead, but I thought we have some zombies left....
I've went over all of them, and seem like I've moved all the places to use pyhon3.7 already....

so it was just a habit I guess...

I would want to remove the python 2.7 classifiers our of `setup.py` to announce this doesn't support python2 anymore

ccmlib/scylla_repository.py
fruch: if that's the reason that's fine

PR #242, opened 2020-07-22
fix(manager): Added version-dependent configuration

ShlomiBalalis:
Due to changes in the configuration after manager 2.1, the segments_per_repair config option can no longer be used.
As such, I added a condition that checks whether the manager version that is being used supports segments_per_repair.

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -9,6 +9,7 @@ import uuid
 import datetime
 
 from six import print_
+from distutils.version import LooseVersion
 
 from ccmlib import common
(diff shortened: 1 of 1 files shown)


ccmlib/scylla_cluster.py
```diff
@@ -9,6 +9,7 @@
 import datetime
 
 from six import print_
+from packaging.version import Version
```
fruch, line 12: this package isn't part of ccm dependency (you are adding it in dtest)

please check if you can use this:
```
from distutils.version import LooseVersion 
```

PR #245, opened 2020-09-06
Change manager-server port to 5080

fruch:
since port 9090 is used by scylla-monitoring stack
we are change this port to default to 5080


amnonh:
LGTM, not sure if it's should be configurable or not, but changing it to 5080 is a good start

PR #246, opened 2020-09-14
scylla_cluster: start_nodes: make watch_log_for "Schema version changed" looser

bhalevy:
A change in scylla moves this log message onto a different logger
so the strict pattern won't match.


fruch (approved):
LGTM

PR #248, opened 2020-10-07
Use parse_version to compare versions conforming to PEP-440

bhalevy:
from `pkg_resources` import `parse_version` and use to to compare structured versions conforming to https://www.python.org/dev/peps/pep-0440/.

This will provided the following ordering:
`'i.j(.|-)dev[N]' < 'i.j.rc[N]' < 'i.j.k' < i.j(.|-)post[N]`

So we can use 'i.j.dev' for the master branch, before it's forked to `branch-i.j` that will initially versioned as `i.j.rc0` and so on until it becomes `i.j.0`.

Cc @avikivity @penberg @fruch 

See also usage in https://github.com/bhalevy/scylla-dtest/commits/parse_version

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -618,7 +618,12 @@ def _get_scylla_version(install_dir):
                                 os.path.join(install_dir, 'scylla-core-package', 'SCYLLA-VERSION-FILE') ]
     for version_file in scylla_version_files:
         if os.path.exists(version_file):
-            return open(version_file).read().strip()
+            v = open(version_file).read().strip()
+            # return only version strings (loosly) conforming to PEP-440
+            # See https://www.python.org/dev/peps/pep-0440/
+            # 'i.j(.|-)dev[N]' < 'i.j.rc[N]' < 'i.j.k' < i.j(.|-)post[N]
+            if re.fullmatch('(\d+!)?\d+([.-]\d+)*([a-z]+\d?)?([.-]post\d?)?([.-]dev\d?)?', v):
+                return v
     return '3.0'
 
 def get_scylla_version(install_dir):
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -593,7 +593,7 @@ def get_version_from_build(install_dir=None, node_path=None):
         install_dir = get_install_dir_from_cluster_conf(node_path)
     if install_dir is not None:
         if isScylla(install_dir):
-            return '3.0'    # return cassandra-compatible version
+            return _get_scylla_version(install_dir)
         # Binary cassandra installs will have a 0.version.txt file
         version_file = os.path.join(install_dir, '0.version.txt')
         if os.path.exists(version_file):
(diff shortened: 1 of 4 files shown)


bhalevy:
Both branches passed `next-gating` on my desktop until it ran out of space.
https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/byo/job/byo-dtest/66 tests full dtest-release.

fruch:
@bhalevy did you test it with 2019.1 version strings ?

fruch (approved):
Looks good, hopefully we'll be able to build ontop better logic in dtest.

bhalevy:
> @bhalevy did you test it with 2019.1 version strings ?

yes, manually...

```python
$ python
Python 3.8.5 (default, Aug 12 2020, 00:00:00) 
[GCC 10.2.1 20200723 (Red Hat 10.2.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pkg_resources import parse_version
>>> parse_version('2019.1.0') > parse_version('3.0')
True
>>> '2019.1.0' > '3.0'
False
```

ccmlib/common.py
```diff
@@ -613,14 +613,17 @@ def get_version_from_build(install_dir=None, node_path=None):
...
 
 
+def _get_scylla_version(install_dir):
+    scylla_version_files = [ os.path.join(install_dir, 'build', 'SCYLLA-VERSION-FILE'),
+                                os.path.join(install_dir, 'scylla-core-package', 'SCYLLA-VERSION-FILE') ]
+    for version_file in scylla_version_files:
+        if os.path.exists(version_file):
+            return open(version_file).read().strip()
+    return '3.0'
+
```
avikivity, line 628: Not this patch's fault, but this looks fragile. Can't we call scylla --version? Or is it before installation?

bhalevy: @avikivity I tried that and it's awfully complicated.  As you said, it needs to be installed first and there are paths getting here way before it is installed.

avikivity: maybe extract it from the tarball name?

bhalevy: we still support installing from the source/build tree, so there won't be a tarball there, unless we force installing only from packages.
Maybe we can get `scylla_version_str` from the symbol table and dump it out of the binary or use gdb to print it.

avikivity: That's also fragile.

It's okay to be fragile when using the source/build tree, that's for developers.

Package install should be more robust.

fruch: that's correct we now have this unified package that has the version in it's name
that wasn't the case before:
http://downloads.scylladb.com/relocatable/unstable/master/2020-10-08T10%3A44%3A59Z/

@syuu1228 AFAIK, there was no changes to CCM to support the unified tarball, right ?

bhalevy: FWIW, here's a bash script that prints the `scylla_version_str` from the binary `.rodata` section:
```bash
#!/usr/bin/env bash

function usage () {
    echo "Usage: $(basename $0) <path_to_scylla>" 1>&2
    exit 1
}

file="$1"

if [ -z ${file} ]; then
    usage
fi

# objdump -ht build/dev/scylla | awk '$2 ~ /\.(t|ro)?data/ { print; next } /scylla_version_str$/ { print }'
# Sections:
# Idx Name          Size      VMA               LMA               File off  Algn
#  11 .rodata       003a04d8  00000000002224e0  00000000002224e0  000224e0  2**5
#  20 .tdata        00000040  000000000218c000  000000000218c000  01f8c000  2**3
#  27 .data         0001c7b8  000000000218f000  000000000218f000  01f8e000  2**5
#
# SYMBOL TABLE:
# 0000000000353888 l     O .rodata      0000000000000009              _ZL18scylla_version_str

dd_opts=$(objdump -ht ${file} | awk '
$2 ~ /\.(t|ro)?data/ {
    sect = $2
    sections[sect] = sect
    sect_base_addr[sect] = $4
    sect_file_offs[sect] = $6
    next
}
/scylla_version_str$/ {
    addr = $1
    sect = $4
    size = $5
    base_addr = sect_base_addr[sect]
    file_offs = sect_file_offs[sect]

    if (base_addr == "" || file_offs == "") {
        printf("Section \"%s\" not found\n", sect) > "/dev/stderr"
        exit(1)
    }

    printf("bs=1 skip=$((0x%s + (0x%s - 0x%s))) count=$((0x%s - 1))\n",
        file_offs, addr, base_addr, size)
}
')

err_file="/tmp/dd.$$.err"
version=$(bash -c "dd ${dd_opts} if=${file} 2>${err_file}") || {
    status=$?
    cat ${err_file} 1>&2
    exit $status
}
rm -f ${err_file}
echo $version
```

PR #250, opened 2020-10-12
ccmlib/scylla_repository: use --packaging option, drop sed to prevent running systemctl

### LINKED ISSUE #249: rewriting install.sh by sed causes syntax error
We currently causing bash syntax error on install.sh while merging scylladb/scylla#7187, likely because we tries to add one more bash function to call systemctl:
```
/jenkins/workspace/scylla-master/next/scylla/.ccm/scylla-repository/5d75ecb069f7db44b0457a634edf9ed596d3de95/scylla-core-package/install.sh: line 386: syntax error near unexpected token `fi'
21:33:39  start_stop_test_node (bootstrap_test.TestBootstrap) ... ERROR
```
Also see: https://groups.google.com/g/scylladb-dev/c/9mK1qMnPEV4/m/w3DeyMB7AQAJ

It's better to use install.sh option to skip systemctl instead of rewriting by sed, it can easily break the script.
I going to merge a patch to support --packaging with nonroot mode, so we can use the option:
https://github.com/scylladb/scylla/pull/7405

syuu1228:
To skip running systemctl in scylla-ccm, rewriting install.sh by sed can break
the script, we can use --packaging option instead.

Fixes #249

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -19,6 +19,8 @@ from six import print_
 from six.moves import urllib
 from six import BytesIO
 
+import packaging.version
+
 from ccmlib.common import (
     ArgumentError, get_default_path, rmdirs, validate_install_dir, get_scylla_version)
 from ccmlib.repository import __download
@@ -48,7 +50,7 @@ def setup(version, verbose=True):
 
         url = os.environ.get('SCYLLA_CORE_PACKAGE', os.path.join(
             s3_url, 'scylla-package.tar.gz'))
-        download_version(version, verbose=verbose, url=url, target_dir=os.path.join(
+        package_version = download_version(version, verbose=verbose, url=url, target_dir=os.path.join(
             tmp_download, 'scylla-core-package'))
         # Try the old name for backward compatibility
         url = os.environ.get("SCYLLA_TOOLS_JAVA_PACKAGE") or \
@@ -69,7 +71,7 @@ def setup(version, verbose=True):
 
         # install using scylla install.sh
         run_scylla_install_script(os.path.join(
-            cdir, 'scylla-core-package'), cdir)
+            cdir, 'scylla-core-package'), cdir, package_version)
     setup_scylla_manager()
 
     return cdir, get_scylla_version(cdir)
@@ -178,16 +180,17 @@ def download_version(version, url=None, verbose=False, target_dir=None):
         package_version_file = "{}/.relocatable_package_version".format(target_dir)
         if os.path.exists(package_version_file):
             with open(package_version_file) as f:
-                package_version = f.read().strip()
-            if package_version != '2':
-                print('Unknown relocatable package format version: ' + package_version)
+                package_version = packaging.version.parse(f.read().strip())
+            if package_version > packaging.version.parse('2.1'):
+                print(f'Unknown relocatable package format version: {package_version}')
                 sys.exit(1)
-            print('Relocatable package format version 2 detected.')
+            print(f'Relocatable package format version {package_version} detected.')
             pkg_dir = glob.glob('{}/*/'.format(target_dir))[0]
             shutil.move(str(pkg_dir), target_dir + '.new')
             shutil.rmtree(target_dir)
             shutil.move(target_dir + '.new', target_dir)
         else:
+            package_version = packaging.version.parse('1')
             print('Legacy relocatable package format detected.')
 
         # add breadcrumb so we could list the origin of each part easily for debugging
@@ -198,6 +201,7 @@ def download_version(version, url=None, verbose=False, target_dir=None):
             f.write("version=%s\n" % version)
             f.write("url=%s\n" % url)
 
+        return package_version
     except urllib.error.URLError as e:
         msg = "Invalid version %s" % version if url is None else "Invalid url %s" % url
         msg = msg + " (underlying error is: %s)" % str(e)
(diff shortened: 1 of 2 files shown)


syuu1228:
@fruch It looks like works with my local branch of https://github.com/scylladb/scylla/pull/7405.
But it does not compatible with latest next, we need to merge https://github.com/scylladb/scylla/pull/7405 first.
 
```
[syuu@monster scylla-ccm.3]$ ccm create scylla-reloc1 -n 1 --scylla --version temp --scylla-core-package-uri=/home/syuu/scylla.2/build/release/scylla-package.tar.gz --scylla-tools-java-package-uri=/home/syuu/scylla.2/tools/java/build/scylla-tools-package.tar.gz --scylla-jmx-package-uri=/home/syuu/scylla.2/tools/jmx/build/scylla-jmx-package.tar.gz
Extracting /home/syuu/scylla.2/build/release/scylla-package.tar.gz as version temp ...
Relocatable package format version 2 detected.
Extracting /home/syuu/scylla.2/tools/java/build/scylla-tools-package.tar.gz as version temp ...
Relocatable package format version 2 detected.
Extracting /home/syuu/scylla.2/tools/jmx/build/scylla-jmx-package.tar.gz as version temp ...
Relocatable package format version 2 detected.
Scylla non-root install completed.
Current cluster is now: scylla-reloc1
```

But this change requires

fruch:
> @fruch It looks like works with my local branch of [scylladb/scylla#7405](https://github.com/scylladb/scylla/pull/7405).
> But it does not compatible with latest next, we need to merge [scylladb/scylla#7405](https://github.com/scylladb/scylla/pull/7405) first.
> 
> ```
> [syuu@monster scylla-ccm.3]$ ccm create scylla-reloc1 -n 1 --scylla --version temp --scylla-core-package-uri=/home/syuu/scylla.2/build/release/scylla-package.tar.gz --scylla-tools-java-package-uri=/home/syuu/scylla.2/tools/java/build/scylla-tools-package.tar.gz --scylla-jmx-package-uri=/home/syuu/scylla.2/tools/jmx/build/scylla-jmx-package.tar.gz
> Extracting /home/syuu/scylla.2/build/release/scylla-package.tar.gz as version temp ...
> Relocatable package format version 2 detected.
> Extracting /home/syuu/scylla.2/tools/java/build/scylla-tools-package.tar.gz as version temp ...
> Relocatable package format version 2 detected.
> Extracting /home/syuu/scylla.2/tools/jmx/build/scylla-jmx-package.tar.gz as version temp ...
> Relocatable package format version 2 detected.
> Scylla non-root install completed.
> Current cluster is now: scylla-reloc1
> ```

can we version `install.sh` somehow ? so we can make the code in ccm backward compatible ? 
I would like that ccm would be able older version without a problem...

penberg:
> I would like that ccm would be able older version without a problem...

The problem is not in `install.sh`, but in ccm doing a string replacement on it. How would you go about versioning this...?

syuu1228:
> can we version `install.sh` somehow ? so we can make the code in ccm backward compatible ?
> I would like that ccm would be able older version without a problem...

@fruch  We can use relocatable package format just like we did on tarball layout change.
So, package relocatable format v3 or later support to use `--packaging` option to skip systemctl.
I just pushed new version of the PR implementing this.

But before merging this, we need to ask @avikivity okay to continue merging,
because there were discussion to renaming the option name:
https://github.com/scylladb/scylla/pull/7405#issuecomment-707531233

@avikivity Do we want to rename `--packaging` first, or can we continue merging this with current option name?

avikivity:
Let's do the renaming later (and keep --packaging for compatibility).

syuu1228:
@fruch Please review this again, if it looks okay I will send fixes on main repo side, and increment relocatable package version.

ccmlib/scylla_repository.py
```diff
@@ -48,7 +48,7 @@ def setup(version, verbose=True):
 
         url = os.environ.get('SCYLLA_CORE_PACKAGE', os.path.join(
             s3_url, 'scylla-package.tar.gz'))
-        download_version(version, verbose=verbose, url=url, target_dir=os.path.join(
+        core_version = download_version(version, verbose=verbose, url=url, target_dir=os.path.join(
```
fruch, line 51: why are you adding this ? I think the version is already part of the cdir, why adding it again ?

ccmlib/scylla_repository.py
```diff
@@ -178,16 +178,17 @@ def download_version(version, url=None, verbose=False, target_dir=None):
         package_version_file = "{}/.relocatable_package_version".format(target_dir)
         if os.path.exists(package_version_file):
             with open(package_version_file) as f:
-                package_version = f.read().strip()
-            if package_version != '2':
-                print('Unknown relocatable package format version: ' + package_version)
+                package_version = int(f.read().strip())
```
fruch, line 181: should we consider using something like LooseVersion instead of int ? so later we could have things like `1.2` ?

ccmlib/scylla_repository.py
fruch: o.k, now I get it, should be named `package_version` like where it's used

fruch (changes requested):
* s/core_version/package_version/
* we should consider not treating versions as int

syuu1228:
@fruch rename core_version with package_version, switch to packaging.version to handle version number.
Since we stopped using int, change new version from 3 to 2.1.
Please review it again.

fruch (approved):
LGTM

PR #251, opened 2020-10-12
ccmlib/scylla_repository: specify correct path to install.sh, drop sed

syuu1228:
We don't want to modify install.sh by sed since it may break the script,
specify correct path to install.sh and drop sed.

Related: #249
Related: scylladb/scylla#7405

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -238,8 +238,8 @@ def __get_dir():
 def run_scylla_install_script(install_dir, target_dir, package_version):
     scylla_target_dir = os.path.join(target_dir, 'scylla')
 
-    # FIXME: remove this hack once scylladb/scylla#4949 is fixed and merged
-    run('''sed 's|"$prefix|"$root/$prefix|' -i install.sh''', cwd=install_dir)
+    run('''{0}/install.sh --prefix {1}/opt/scylladb --nonroot'''.format(
+        install_dir, scylla_target_dir), cwd=install_dir)
 
     if package_version <= packaging.version.parse('2'):
         # leave this for compatibility


fruch:
@syuu1228 as I've commented on #249, can we make this backward compatible ? so it won't break when using older versions of scylla ?

fruch (changes requested):
Let's try to make it backward compatible, 
so older version can be used (for example formal 4.1/4.2 versions)

fruch:
@syuu1228 is this mean #250 can be closed ?

fruch:
@syuu1228 we can close this one, right ?

syuu1228:
@fruch It different with #249, trying to drop `sed 's|"$prefix|"$root/$prefix|' -i install.sh` line, not systemctl one.

About versioning this fix, unlike #249, we don't really need versioning for this PR.
Because the result of the script should be identical with current implementation, also it doesn't require changes on install.sh of scylla.
But if you still prefer versioning this to make sure not to break anything, I'll increment pkg version to 2.2 for this change.

Note: rebased with latest next.

fruch:
> @fruch It different with #249, trying to drop `sed 's|"$prefix|"$root/$prefix|' -i install.sh` line, not systemctl one.
> 
> About versioning this fix, unlike #249, we don't really need versioning for this PR.
> Because the result of the script should be identical with current implementation, also it doesn't require changes on install.sh of scylla.
> But if you still prefer versioning this to make sure not to break anything, I'll increment pkg version to 2.2 for this change.
> 
> Note: rebased with latest next.

sorry, it's need a rebase again.
and yes I would rather be on the safe side with those kind of changes, not to break installation of previous version (since we are introducing new upgrade test the really on this code path)

syuu1228:
Seems like b7251e3eeba80e183efcaf61133e218420e51921 already fixed same issue, closing.

PR #253, opened 2020-10-18
nodetool: add timeout

bhalevy:
Add a timeout parameter to nodetool since we saw to hanging in
https://github.com/scylladb/scylla/issues/7244

Currently, the default timeout is `None`, so the behavior remains backward compatible
and nodetool will be waited for forever.
Only the explicit paths seen in https://github.com/scylladb/scylla/issues/7244 pass a timeout in this series.
We may want to consider as a follow up setting the default timeout to a long, but not infinite, timeout, like 10 minutes or more, so that if it hangs, the test will fail and if scylla hangs we can kill it and get a core dump.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -709,11 +709,13 @@ class Node(object):
                 break
             time.sleep(10)
 
-    def nodetool(self, cmd, capture_output=True, wait=True):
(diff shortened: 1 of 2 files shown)


bhalevy:
@fruch please review

fruch (approved):
LGTM

PR #257, opened 2020-10-22
feature(.travis.yaml): Creating CI via travis

Orenef11:
* Move old CCM test to `old_tests` directory
* Create new `session` fixtures:
   * results_dir - The directory of the cluster
   * test_id - unique cluster ID
   * docker_cluster - Cluster with 3 nodes
 * test_config.py - Create one place to create test variables
 * .gitignore - Add `tests/test_results` to file


setup.py
```diff
@@ -24,7 +24,7 @@
     url='https://github.com/pcmanus/ccm',
     packages=['ccmlib', 'ccmlib.cmds'],
     scripts=[ccmscript],
-    install_requires=['pyYaml', 'psutil', 'six >=1.4.1', 'requests'],
+    install_requires=['pyYaml', 'psutil', 'six >=1.4.1', 'requests', 'pytest'],
```
fruch, line 27: Nope don't add it to the package this a test requirement

fruch: Use tests_require

fruch:
@Orenef11 you are removing files here which are modified in other people PRs. please sync with them if those files should be removed or not

fruch:
@Orenef11 please undo the deletion of test files, and lets get this merged regardless of it the tests passes or not
so we could start seeing this running on the PRs we are opening, and then we could address the tests issues.

Orenef11 (commented):
> @Orenef11 please undo the deletion of test files, and lets get this merged regardless of it the tests passes or not
> so we could start seeing this running on the PRs we are opening, and then we could address the tests issues.

@fruch 
The old tests do not work and cannot be run.
I can move them to another folder so we can use them.
Is it okay?

Orenef11:
@fruch 
All issues are fixed

old_tests/__init__.py
```diff
@@ -0,0 +1,23 @@
+import os
```
Orenef11, line 1: @fruch 
I move the file to new path (this is the original code)

tests/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+@pytest.fixture(scope="session")
+def test_id():
+    ccm_test_id = TEST_ID or datetime.now().strftime("%Y%m%d-%H%M%S")
+    LOGGER.info("Using test id: '{}'".format(ccm_test_id))
+    return ccm_test_id
+
+
+@pytest.fixture(scope="session")
+def test_dir(test_id, results_dir):
+    max_test_dirs, dir_count = 1, 100
```
fruch, line 31: why all this `max_test_dirs` ? why can't we just use one directory ?

fruch (approved):
LGTM

tests/conftest.py
Orenef11: @fruch 
I copied the code from Siren, it is meant to limit the number of directories - we do not want to have a lot of directories on the computer.
If will be free time, the idea is that it will be possible to stop the test and use the same cluster in the next run as well (assuming the cluster is up).

PR #259, opened 2020-10-22
fix(unittest): modified the test to run with scylla

ShlomiBalalis:


fruch:
@ShlomiBalalis what the all the test in TestCCMLib are working ? if they aren't pytest markers should be handy here, like `xfail` and friends 

@Orenef11 if you have some suggestion regarding pytest, this is the time :)

tests/test_lib.py
```diff
@@ -1,53 +1,59 @@
 import sys
 sys.path = [".."] + sys.path
 
+import os
 import decimal
 
-from . import TEST_DIR
-from . import ccmtest
-from ccmlib.cluster import Cluster
+from tests import TEST_DIR
```
fruch, line 7: @Orenef11 I think this answered you question, seem you weren't the only encounter it

tests/test_lib.py
```diff
@@ -1,53 +1,59 @@
...
+from tests import ccmtest
+# from ccmlib.cluster import Cluster
```
fruch, line 9: if we ditch cassandra test out of here, we could remove those comments

tests/test_lib.py
```diff
@@ -1,53 +1,59 @@
...
+from ccmlib.scylla_cluster import ScyllaCluster
 import ccmlib
+# from ccmlib
 from six import StringIO
 
 CLUSTER_PATH = TEST_DIR
+if not os.path.exists(CLUSTER_PATH):
```
fruch, line 16: let move this kind of code to `setUpClass` or to pytest session scoped fixture.
doing thing like that on module loading, it can't be problematic

tests/test_lib.py
ShlomiBalalis: Of course, this is why it's a draft

fruch: lets clean it up just a bit and merge...

tests/test_lib.py
ShlomiBalalis: It was indeed a bit weird. Is this a remainder from python 2?

Orenef11:
> @ShlomiBalalis what the all the test in TestCCMLib are working ? if they aren't pytest markers should be handy here, like `xfail` and friends
> 
> @Orenef11 if you have some suggestion regarding pytest, this is the time :)

I thinking we sou

> @ShlomiBalalis what the all the test in TestCCMLib are working ? if they aren't pytest markers should be handy here, like `xfail` and friends
> 
> @Orenef11 if you have some suggestion regarding pytest, this is the time :)

Once the PR will ready, I can convert it to the pytest.
The change should not impact on the content of the tests (I thinking) :-)

fruch (changes requested):
remove the `os.mkdir(CLUSTER_PATH)` from the file import (i.e. we don't want side effect like that in import a module)

fruch:
we already have running tests

PR #261, opened 2020-10-22
feature(pytest): Added pytest config flies

aleksbykov:
Added config files for pytest framework

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,7 @@
+[pytest]
+python_classes = *Test Test*
+python_files = test_* *_test
+python_functions = test_* *_test
+testpaths = tests
+addopts = -rA -x -l --tb=short
+norecursedirs = .* ssl *.egg-info ccmlib docs


pytest.ini
```diff
@@ -0,0 +1,7 @@
+[pytest]
+python_classes = *Test Test*
+python_files = test_* *_test
+python_functions = test_* *_test
+testpaths = tests
+addopts = -rsxX -l --tb=short
```
fruch, line 6: Why using all those defaults ? half of them I don't remember exactly.

But I'm more sure we want by default to stop of first failure

tests/test_sanity.py
```diff
@@ -0,0 +1,74 @@
...
+        print(os.path.abspath(os.curdir))
+
+        workdir = os.path.abspath(os.curdir)
+
+        self.create_keyspace = """CREATE KEYSPACE ks WITH replication = {'class': 'SimpleStrategy', 'replication_factor' : 2};"""
+
+        # self.ccm_cmd = [python_int, f'{workdir}/ccm']
+        self.ccm_cmd = [f'{workdir}/ccm']
+        self.name = "test_one"
+        print(self.ccm_cmd)
```
fruch, line 23: pytest has very good capture of logs, let use logging module  and not prints

PR #262, opened 2020-10-22
Initial crude implmention of `ScyllaDockerCluster`

fruch:


ccmlib/scylla_cluster.py
```diff
@@ -133,9 +139,9 @@ def start_nodes(self, nodes=None, no_wait=True, verbose=False, wait_for_binary_p
     def start(self, no_wait=True, verbose=False, wait_for_binary_proto=None,
               wait_other_notice=None, jvm_args=None, profile_options=None,
               quiet_start=False):
-        args = locals()
```
fruch, line 136: This is a change needed in CCM, regardless of our work,
without it, this part of code is failing under debugger

ccmlib/scylla_cluster.py
```diff
@@ -168,9 +174,9 @@ def stop_nodes(self, nodes=None, wait=True, gently=True, wait_other_notice=False
     def stop(self, wait=True, gently=True, wait_other_notice=False, other_nodes=None, wait_seconds=127):
         if self._scylla_manager:
             self._scylla_manager.stop(gently)
-        args = locals()
-        del args['self']
-        return self.stop_nodes(**args)
+        kwargs = dict(**locals())
```
fruch, line 177: This is a change needed in CCM, regardless of our work,
without it, this part of code is failing under debugger

ccmlib/cluster.py
```diff
@@ -41,14 +41,19 @@ def __init__(self, path, name, partitioner=None, install_dir=None, create_direct
...
+        version = kwargs.get('cassandra_version', version)
+        install_dir = kwargs.get('cassandra_dir', install_dir)
+        docker_image = kwargs.get('docker_image')
         if create_directory:
             # we create the dir before potentially downloading to throw an error sooner if need be
             os.mkdir(self.get_path())
 
+        if docker_image:
+            self.__install_dir = None
+            self.__version = '3.0'
```
fruch, line 53: meanwhile we'll hardcode this version, until we'll get a reasonable way to pull the version out of the image

ccmlib/scylla_node.py
```diff
@@ -891,10 +891,10 @@ def run_patchelf(patchelf_cmd):
 
     def _save(self):
         # TODO: - overwrite node
-        self.__update_yaml()
```
fruch, line 894: Had to rename this function, since function with __ are not overwritable

ccmlib/cluster.py
fgelcer: what is this version about?

ccmlib/scylla_docker_cluster.py
```diff
@@ -0,0 +1,53 @@
...
+                    storage_interface, jmx_port, remote_debug_port,
+                    initial_token, save=True, binary_interface=None):
+
+        return ScyllaDockerNode(name, self, auto_bootstrap, thrift_interface,
+                                storage_interface, jmx_port, remote_debug_port,
+                                initial_token, save=save, binary_interface=binary_interface,
+                                scylla_manager=self._scylla_manager)
+
+
+class ScyllaDockerNode(ScyllaNode):
```
fgelcer, line 23: shouldn't this class go under `ccmlib/scylla_docker_node.py` instead of here?

fgelcer (commented):
comments below

ccmlib/scylla_docker_cluster.py
fruch: Probably, I'm just trying to sketch things up first, splitting to a new file would take 2 min...

ccmlib/cluster.py
fruch: Across ccm there lots of place logic is based on the version. In other options we have the version available on a file, it's less critical, but would be nice to get the version at an early stage somehow (maybe like I'm getting the scylla.yaml running the image temporary just to run `scylla --version`

ccmlib/node.py
```diff
@@ -803,6 +803,8 @@ def run_cli(self, cmds=None, show_output=False, cli_options=[]):
 
     def run_cqlsh(self, cmds=None, show_output=False, cqlsh_options=[], return_output=False):
         cqlsh = self.get_tool('cqlsh')
+        if not isinstance(cqlsh, list):
```
fruch, line 806: had to change `get_tool` to return a list of arguments, so `docker exec` can be added

ccmlib/scylla_docker_cluster.py
```diff
@@ -0,0 +1,208 @@
...
+        res = run(['bash', '-c', f'docker rm -f {self.pid}'], stdout=PIPE, stderr=PIPE)
+        self.log_thread.stop()
+        super(ScyllaDockerNode, self).clear(*args, **kwargs)
+
+    def _start_jmx(self, data):
+        jmx_status = self.service_status('scylla-jmx')
+        if not jmx_status and jmx_status.upper() == 'RUNNING':
+            self.service_start('scylla-jmx')
+
+    def is_running(self):
```
fruch, line 133: had to copy out those functions, since `__update_status` couldn't overwrite

ccmlib/scylla_docker_cluster.py
```diff
@@ -0,0 +1,208 @@
...
+    def _wait_java_up(self, ip_addr, jmx_port):
+        return True
+
+    def _update_pid(self, process):
+        pass
+
+    def get_tool(self, toolname):
+        return ['docker',  'exec', '-i',  f'{self.pid}', f'{toolname}']
+
+    def get_env(self):
```
fruch, line 170: no need for all kind of java environment ccm is doing, since everything should be configured correctly inside the docker

tests/test_scylla_docker_cluster.py
```diff
@@ -0,0 +1,29 @@
+import os.path
+from subprocess import run
+from unittest import TestCase
+
+from ccmlib.scylla_docker_cluster import ScyllaDockerCluster
+
+
+class TestScyllaDockerCluster(TestCase):
+    def test_01(self):
+        test_path = os.path.expanduser("~/.test")
```
fruch, line 10: now we won't need to edit he cluster name again and again

tests/test_scylla_docker_cluster.py
```diff
@@ -0,0 +1,29 @@
...
+        run(["bash", "-c", f"rm -rf {test_path}/test_cluster01"])
+        cluster = ScyllaDockerCluster(test_path, name='test_cluster01',
+                                      docker_image='scylladb/scylla-nightly:666.development-0.20201015.8068272b466')
+        self.addCleanup(lambda: cluster.clear())
```
fruch, line 14: tested clear, that should remove the docker instances

ccmlib/scylla_docker_cluster.py
```diff
@@ -0,0 +1,208 @@
...
+        return os.environ.copy()
+
+
+import subprocess
+from threading import Thread, Event as ThreadEvent
+
+
+class DockerLogger:
```
fruch, line 178: Add this to get the `docker logs` in real time into a file, so all the `mark_log` functionality would work

ccmlib/scylla_docker_cluster.py
```diff
@@ -0,0 +1,208 @@
...
+        # TODO: handle smp correctly via the correct param/api (or only via commandline params)
+        # TODO: mount of the data dir
+        # TODO: pass down the full command line params, since the docker ones doesn't support all of them ?
+        # TODO: pass down a unique tag, with the cluster name, or id, if we have such in ccm, like test_id in SCT ?
+
+        rc, out, err = self.create_docker()
+        if not rc:
+            raise BaseException(f'failed to create docker {self.docker_name}')
+
+        time.sleep(5)
```
fruch, line 106: we need to replace this with a loop until running

PR #263, opened 2020-10-22
improvement(setup.py): Updating Setup file

Orenef11:
* Updating the `author` field value
* Updating `install_requires` field and read the
   require from `requirements.txt` file
* Creating `requirements.txt` file with following packages:
  * pyYaml==5.3.1
  * psutil==5.7.2
  * six==1.15.0
  * requests==2.24.0
  * pytest==6.1.1


setup.py
```diff
@@ -1,39 +1,42 @@
...
-    version='2.0.5',
-    description='Cassandra Cluster Manager',
-    long_description=open(abspath(join(dirname(__file__), 'README.md'))).read(),
-    author='Sylvain Lebresne',
-    author_email='sylvain@datastax.com',
-    url='https://github.com/pcmanus/ccm',
-    packages=['ccmlib', 'ccmlib.cmds'],
-    scripts=[ccmscript],
-    install_requires=['pyYaml', 'psutil', 'six >=1.4.1', 'requests'],
+    name="ccm-docker",
```
fruch, line 20: don't rename the package

requirements.txt
```diff
@@ -0,0 +1,6 @@
+pyYaml==5.3.1
```
fruch, line 1: again, I don't want those version pinned in CCM
also pytest isn't a dependency for ccm, it's `test_require` part not the `install_requires` part

setup.py
Orenef11: fixed

ccmlib/containers/docker_client.py
```diff
@@ -0,0 +1,143 @@
...
+        Check if class container state is running
+        :return: boolean
+        """
+        try:
+            return self.container.status == "running"
+        except docker.errors.APIError:
+            return False
+
+    @staticmethod
+    def server_ip(destination_address="8.8.8.8"):
```
fruch, line 102: what this function is doing ? and for which propose ?

requirements.txt
Orenef11: I removed the version from each package.
I fixed the dependency

setup.py
```diff
@@ -1,39 +1,43 @@
...
-    packages=['ccmlib', 'ccmlib.cmds'],
-    scripts=[ccmscript],
-    install_requires=['pyYaml', 'psutil', 'six >=1.4.1', 'requests'],
+    name="ccm",
+    version="2.0.5",
+    description="Cassandra Cluster Manager",
+    long_description=long_description,
+    author="Balalis Shlomo, Bykov Aleksandr, Efraimov Oren, Fruchter Israel, Gelcer Fabio, Janković Aleksandar and"
+           "Kaikov Yaron",
+    package_data={"": ["*"]},
```
fruch, line 26: what's this ?

setup.py
```diff
@@ -1,39 +1,43 @@
...
-    author='Sylvain Lebresne',
-    author_email='sylvain@datastax.com',
-    url='https://github.com/pcmanus/ccm',
-    packages=['ccmlib', 'ccmlib.cmds'],
-    scripts=[ccmscript],
-    install_requires=['pyYaml', 'psutil', 'six >=1.4.1', 'requests'],
+    name="ccm",
+    version="2.0.5",
+    description="Cassandra Cluster Manager",
+    long_description=long_description,
```
fruch, line 23: if you are touch it, please add `long_description_content_type='text/markdown'`
https://packaging.python.org/guides/making-a-pypi-friendly-readme/

ccmlib/containers/docker_client.py
```diff
@@ -0,0 +1,143 @@
...
+    def restart_service(self, service_name):
+        """
+        This method restarts the given service name
+        :param service_name: str name of the service
+        """
+        LOGGER.info(f"Restarting the service {service_name} in container {self.name}")
+        self.execute_command(f"service {service_name} restart")
+
+    @property
+    def logs(self):
```
fruch, line 139: I think this would be good enough, since we have the `mark_log` functionally which is very important for dtests, and getting the logs like that won't play nice with it.

we need a file that keep getting written to

ccmlib/containers/docker_client.py
```diff
@@ -0,0 +1,143 @@
...
+        self.container_pid = None
+        self.container_name = None
+        self.client = docker.from_env()
+        self.volumes = volumes
+
+        self.pull()
+        self.container_pid, self.container_name = self.create_container(
+            ports={"9180/tcp": port}, volumes=volumes, is_clear_container=is_clear_container, **create_kwargs)
+
+    def pull(self):
```
fgelcer, line 25: why do we need this function?
to create this object, i have to already send the image to the `__init__`, so this seems useless

ccmlib/containers/docker_client.py
```diff
@@ -0,0 +1,143 @@
...
+            self.remove_old_containers()
+        LOGGER.debug("Creating container using image '{}'".format(self.image))
+        self.container = self.client.containers.create(
+            image=self.image, hostname=self.name, detach=True, stdin_open=True, tty=True, ports=ports, volumes=volumes,
+            name=self.name, network_disabled=False, privileged=True, pid_mode="host", **kwargs)
+        LOGGER.debug("Start container '{}'".format(self.name))
+        self.container.start()
+        return self.container.attrs["State"]["Pid"], self.container.name
+
+    def close(self):
```
fgelcer, line 56: i think it is better to call it `stop_container`

ccmlib/containers/docker_client.py
```diff
@@ -0,0 +1,143 @@
...
+        LOGGER.info(f"Starting the service {service_name} in container {self.name}")
+        self.execute_command(f"service {service_name} start")
+
+    def restart_service(self, service_name):
+        """
+        This method restarts the given service name
+        :param service_name: str name of the service
+        """
+        LOGGER.info(f"Restarting the service {service_name} in container {self.name}")
+        self.execute_command(f"service {service_name} restart")
```
fgelcer, line 136: i think all of this must move to the `scylla_docker_node.py`

setup.py
```diff
@@ -1,39 +1,43 @@
...
-    description='Cassandra Cluster Manager',
-    long_description=open(abspath(join(dirname(__file__), 'README.md'))).read(),
-    author='Sylvain Lebresne',
-    author_email='sylvain@datastax.com',
-    url='https://github.com/pcmanus/ccm',
-    packages=['ccmlib', 'ccmlib.cmds'],
-    scripts=[ccmscript],
-    install_requires=['pyYaml', 'psutil', 'six >=1.4.1', 'requests'],
+    name="ccm",
+    version="2.0.5",
```
fgelcer, line 21: if you are touching it, we should maybe manage the version by some git command, or a version file?

ccmlib/containers/scylla_node.py
```diff
@@ -0,0 +1,10 @@
+from ccmlib.containers.docker_client import BaseDocker
```
fgelcer, line 1: i would call the whole file with a different name to reflect this is basically a docker container and not a `scylla_node` from the CCM perspective

fgelcer (commented):
comments and questions below

setup.py
Orenef11: which field?

setup.py
Orenef11: Done

setup.py
fruch: `package_data`, I don't understand why are you adding everything into the package ?

ccmlib/containers/docker_client.py
fruch: I've handled logging in #262 already (borrowed it from SCT)

ccmlib/containers/docker_client.py
fruch: Pull would be a good thing to add as a very first step, like part where we automatically download the relocatable packages.
to have a early info, if the image exist or not

ccmlib/containers/docker_client.py
Orenef11: This method returns the node public IP

setup.py
Orenef11: My mistake.
fixed

fgelcer:
@Orenef11 , what happens with this PR?
also, change the base to `docker_ccm`

Orenef11 (commented):
> @Orenef11 , what happens with this PR?
> also, change the base to `docker_ccm`

@fgelcer 
I did not want to make drastic changes and interfere with the rest of the PRs.
Because most of the PRs are already merged, it is possible today to merge this PR.

> @Orenef11 , what happens with this PR?
> also, change the base to `docker_ccm`

@fgelcer 
I did not want to make drastic changes and interfere with the rest of the PRs.
Because most of the PRs are already merged, it is possible today to merge this PR.

fruch:
@Orenef11 you still think this one is need ? for now I'll close it.

PR #264, opened 2020-10-24
Docker ccm

fgelcer:
started from @fruch PR on #262 and added the start function as it should be to the `scylla_docker_node` for both Scylla and jmx.

now, it is failing as the node didn't return a PID, so maybe we need to override that function that comes from `node`

### DIFF
--- a/ccmlib/scylla_docker_cluster.py
+++ b/ccmlib/scylla_docker_cluster.py
@@ -1,0 +1,53 @@
+import os
+from subprocess import run
+
+from ccmlib.scylla_cluster import ScyllaCluster
+from ccmlib.scylla_node import ScyllaNode
+
+
+class ScyllaDockerCluster(ScyllaCluster):
+    def __init__(self, *args, **kwargs):
+        super(ScyllaDockerCluster, self).__init__(*args, **kwargs)
+        self.docker_image = kwargs['docker_image']
+
+    def create_node(self, name, auto_bootstrap, thrift_interface,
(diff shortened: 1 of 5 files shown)


ccmlib/scylla_docker_cluster.py
```diff
@@ -24,30 +25,78 @@ class ScyllaDockerNode(ScyllaNode):
...
+        # res = run([f'docker run --rm --entrypoint cat {self.cluster.docker_image}  '
+        #            f'/etc/scylla/scylla.yaml > {self.local_yaml_path}'],
+        #           stdout=PIPE, stderr=PIPE)
+        # if not res.returncode:
+        #     print(f'Failed to copy scylla.yaml to {self.local_yaml_path}\n{res.stderr}')
+        rc, out, err = self.create_docker()
+        if not rc:
+            raise BaseException(f'failed to create docker {self.docker_name}')
         super(ScyllaDockerNode, self).start(*args, **kwargs)
+        return random.randint(100, 10000)
```
fruch, line 86: Why are we returning a random number here ?
To simulate a pid number ?

I don't think I would be helpful, we will need to overwrite all this pid handling code somehow...

fgelcer: yeah, the idea was to simulate a PID number, but it didn't work either (and i forgot to remove it)

and yes, we will need to make changes in all this PID handling (it might be easier than we think, if we override the methods in the base node code to use the docker classes)

fruch:
I've took the commit from here into #262

PR #266, opened 2020-10-25
Docker ccm fix docker name

fgelcer:
from `get_path` i got the last part and added to the `docker_name` to make dockers unique..
the path when running from dtest will be generated by the framework and the path is the only difference between the "clusters"

### DIFF
--- a/ccmlib/scylla_docker_cluster.py
+++ b/ccmlib/scylla_docker_cluster.py
@@ -36,7 +36,7 @@ class ScyllaDockerNode(ScyllaNode):
         self.docker_id = None
         self.local_data_path = os.path.join(self.get_path(), 'data')
         self.local_yaml_path = os.path.join(self.get_path(), 'conf')
-        self.docker_name = f'{self.cluster.name}-{self.name}'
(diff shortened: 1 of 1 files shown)


ccmlib/scylla_docker_cluster.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, *args, **kwargs):
         self.docker_id = None
         self.local_data_path = os.path.join(self.get_path(), 'data')
         self.local_yaml_path = os.path.join(self.get_path(), 'conf')
-        self.docker_name = f'{self.cluster.name}-{self.name}'
+        self.docker_name = f'{self.cluster.get_path().split("/")[-1]}-{self.cluster.name}-{self.name}'
```
Orenef11, line 34: @fgelcer 
What the result of `self.cluster.get_path().split("/")[-1]}`?

fgelcer: if running from ccm (like command line) it will be the `cluster_name`
if running from dtest, then it will be the test directory with a unique name

Orenef11:
@fgelcer 
The travis CI is failed

Orenef11:
LGTM

PR #267, opened 2020-10-25
Docker ccm nodetool status test

fgelcer:
added 1st test to check if `nodetool status` returns the expected values

### DIFF
--- a/tests/test_scylla_docker_cluster.py
+++ b/tests/test_scylla_docker_cluster.py
@@ -1,5 +1,48 @@
+import re
+
+
 class TestScyllaDockerCluster:
-    def test_cqlsh(self, docker_cluster):
+    @staticmethod
+    def parse_nodetool_status(lines):
+        """parse output of nodetool status
+
+        Nodetool status output:
+        Datacenter: eu-west
+        ===================
+        Status=Up/Down
+        |/ State=Normal/Leaving/Joining/Moving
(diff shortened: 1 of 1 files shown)


tests/test_scylla_docker_cluster.py
```diff
@@ -1,4 +1,46 @@
...
+            {"status": "UN", "address": "10.0.15.114", size: "36.53", dimension: KB|MB|GB},
+            {"status": "UN", "address": "10.0.126.57", size: "88.17", dimension: KB|MB|GB},
+            {"status": "UN", "address": "10.0.74.155", size: "90.62", dimension: KB|MB|GB},
+            {"status": "UN", "address": "10.0.65.254", size: "101.39", dimension: KB|MB|GB}
+
+        ]
+        """
+        keys = ["status", "address", "size", "dimension"]
+        nodes_statuses = []
+        line_re = re.compile(
```
Orenef11, line 30: 1. The IP address will be only `IPv4` address?
2. some fields not parsed, why?

fruch (approved):
LGTM

tests/test_scylla_docker_cluster.py
fgelcer: for this simple test case, we could assume that the fields we need are here, and that most likely the cluster will use IPv4... if we get to work with IPv6 we could update this test to reflect it

PR #268, opened 2020-10-25
    feature(.node.py): Fixing stress command for docker cluster

Orenef11:
* Create `docker_image` variable under `Node` class
* Change the `jmx_port` field under ScyllaDockerNode from
    int to string (because the old CCN code expected to get
    string)
* stress method - Convert the `stress` variable to list and 
    set None value for `cwd` variable in case the cluster is 
    docker

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1133,14 +1133,20 @@ class Node(object):
         else:
             stress_options = stress_options[:]
 
-        stress = common.get_stress_bin(self.get_install_dir())
+        if self.is_docker():
+            stress = self.get_tool("cassandra-stress")
+            cmd = None
+        else:
+            stress = common.get_stress_bin(self.get_install_dir())
+            cmd = common.parse_path(stress)
+            stress = [stress]
 
         # handle cassandra-stress >= 3.11 parameters
         has_limit = any(['limit=' in o for o in stress_options])
         has_throttle = any(['throttle=' in o for o in stress_options])
         if has_limit or has_throttle:
-            args = [stress, 'version']
-            p = subprocess.Popen(args, cwd=common.parse_path(stress),
+            args = stress + ['version']
+            p = subprocess.Popen(args, cwd=cmd,
                                  stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True,
                                  **kwargs)
             stdout, stderr = p.communicate()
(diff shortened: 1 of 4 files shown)


ccmlib/scylla_cluster.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, path, name, partitioner=None, install_dir=None,
             install_dir, self.scylla_mode = install_func(install_dir)
 
         self.started = False
-        self.force_wait_for_cluster_start = (force_wait_for_cluster_start != False)
```
fruch, line 44: why are you changing this ?

ccmlib/node.py
```diff
@@ -1133,14 +1134,20 @@ def stress(self, stress_options=None, capture_output=False, **kwargs):
         else:
             stress_options = stress_options[:]
 
-        stress = common.get_stress_bin(self.get_install_dir())
+        if self.docker_name:
```
fruch, line 1137: replace with `self.is_docker()`

ccmlib/scylla_cluster.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, path, name, partitioner=None, install_dir=None,
         install_func = common.scylla_extract_install_dir_and_mode
 
         cassandra_version = kwargs.get('cassandra_version', version)
-        docker_image = kwargs.get('docker_image')
+        docker_image = kwargs.pop('docker_image', None)
```
fruch, line 31: why are you using pop here ?

ccmlib/scylla_cluster.py
Orenef11: I forgot to revert this change.
Fixed

ccmlib/node.py
Orenef11: Fixed

ccmlib/scylla_cluster.py
Orenef11: I dont know there is `is_docker` method and I passed it tho all Node object (I revert this change)

fgelcer: i still see it is `poping` it here

fgelcer (commented):
waiting for a you to replace the `kwargs.pop` to be reverted...

ccmlib/cluster.py
```diff
@@ -43,12 +43,12 @@ def __init__(self, path, name, partitioner=None, install_dir=None, create_direct
         # backwards compatibility.
         version = kwargs.get('cassandra_version', version)
         install_dir = kwargs.get('cassandra_dir', install_dir)
-        docker_image = kwargs.get('docker_image')
+        self.docker_image = kwargs.get('docker_image')
```
fruch, line 46:
```suggestion
        docker_image = kwargs.get('docker_image')
```

ccmlib/cluster.py
```diff
@@ -43,12 +43,12 @@ def __init__(self, path, name, partitioner=None, install_dir=None, create_direct
...
         if create_directory:
             # we create the dir before potentially downloading to throw an error sooner if need be
             os.mkdir(self.get_path())
 
-        if docker_image:
+        if self.docker_image:
```
fruch, line 51:
```suggestion
        if docker_image:
```

fruch (changes requested):
don't assign `docker_image` in `cluster.py`

ccmlib/scylla_cluster.py
Orenef11: Thanks, done :-)

ccmlib/cluster.py
Orenef11: Thanks, done :-)

ccmlib/cluster.py
Orenef11: Thanks, done :-)

PR #271, opened 2020-10-27
    feature(scylla_docker_cluster.py): Enabling SSL keys

Orenef11:
* Create share directory for RSA keys

### DIFF
--- a/ccmlib/scylla_docker_cluster.py
+++ b/ccmlib/scylla_docker_cluster.py
@@ -1,4 +1,5 @@
 import os
+from shutil import copyfile
 from subprocess import run, PIPE
 import logging
 
@@ -65,6 +66,8 @@ class ScyllaDockerCluster(ScyllaCluster):
 class ScyllaDockerNode(ScyllaNode):
     def __init__(self, *args, **kwargs):
         kwargs['save'] = False
+        # This line should appear before the "super" method
+        self.share_directories = ['data', 'commitlogs', 'hints', 'view_hints', 'saved_caches', 'keys', 'logs']
         super(ScyllaDockerNode, self).__init__(*args, **kwargs)
         self.base_data_path = '/usr/lib/scylla'
         self.local_base_data_path = os.path.join(self.get_path(), 'data')
@@ -76,8 +79,8 @@ class ScyllaDockerNode(ScyllaNode):
 
     def _get_directories(self):
         dirs = {}
-        for i in ['data', 'commitlogs', 'conf', 'logs', 'hints', 'view_hints', 'saved_caches']:
-            dirs[i] = os.path.join(self.get_path(), i)
+        for dir_name in self.share_directories + ['conf']:  # conf dir is handle in other way in `update_yaml()`
+            dirs[dir_name] = os.path.join(self.get_path(), dir_name)
         return dirs
 
     @staticmethod
@@ -118,6 +121,14 @@ class ScyllaDockerNode(ScyllaNode):
         for directory in ['hints', 'view_hints', 'saved_caches']:
             data[f'{directory}_directory'] = os.path.join(self.base_data_path, directory)
 
+        server_encryption_options = data.get("server_encryption_options", {})
+        if server_encryption_options:
+            keys_dir_path = os.path.join(self.get_path(), "keys")
+            for key, file_path in server_encryption_options.items():
+                if os.path.isfile(file_path):
+                    file_name = os.path.split(file_path)[1]
+                    copyfile(src=file_path, dst=os.path.join(keys_dir_path, file_name))
+                    server_encryption_options[key] = os.path.join(self.base_data_path, "keys", file_name)
         with open(conf_file, 'w') as f:
             yaml.safe_dump(data, f, default_flow_style=False)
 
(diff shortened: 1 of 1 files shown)


ccmlib/scylla_docker_cluster.py
```diff
@@ -39,6 +41,8 @@ def __init__(self, *args, **kwargs):
         self.docker_name = f'{self.cluster.get_path().split("/")[-2]}-{self.cluster.name}-{self.name}'
         self.jmx_port = "7199"  # The old CCM code expected to get a string and not int
         self.log_thread = None
+        self.share_directory = os.path.join(tempfile.mkdtemp(), os.path.split(self.get_path())[1])
```
fgelcer, line 44: shouldn't it be inside the cluster or node configuration tree (under .ccm)

fgelcer (commented):
what are you doing that for? to have the encryption files mapped into the container?

ccmlib/scylla_docker_cluster.py
Orenef11: fixed

ccmlib/scylla_docker_cluster.py
```diff
@@ -48,18 +50,21 @@ def _update_config(self):
 class ScyllaDockerNode(ScyllaNode):
     def __init__(self, *args, **kwargs):
         kwargs['save'] = False
+        # This line should appear before the "supper" method
```
fgelcer, line 53: typo
```suggestion
        # This line should appear before the "super" method
```

ccmlib/scylla_docker_cluster.py
```diff
@@ -48,18 +50,21 @@ def _update_config(self):
...
+        self.share_directories = ['data', 'commitlogs', 'conf', 'hints', 'view_hints', 'saved_caches', 'keys']
         super(ScyllaDockerNode, self).__init__(*args, **kwargs)
         self.base_data_path = '/usr/lib/scylla'
         self.local_base_data_path = os.path.join(self.get_path(), 'data')
         self.local_yaml_path = os.path.join(self.get_path(), 'conf')
-        self.docker_name = f'{self.cluster.get_path().split("/")[-1]}-{self.cluster.name}-{self.name}'
+        folder_name = self.cluster.get_path().split("/")[-2]
```
fgelcer, line 59: i became skeptic about this... should it really be `[-2]` or `[-1]`?
i saw when running from ccm cmd line it is not good with -2... locally i'm running it with -1 and it works fine (from ccm cmd line POV)

ccmlib/scylla_docker_cluster.py
```diff
@@ -48,18 +50,21 @@ def _update_config(self):
...
+        self.docker_name = f'{folder_name}-{self.cluster.name}-{self.name}'
         self.jmx_port = "7199"  # The old CCM code expected to get a string and not int
         self.log_thread = None
 
     def _get_directories(self):
         dirs = {}
-        for i in ['data', 'commitlogs', 'conf', 'logs', 'hints', 'view_hints', 'saved_caches']:
-            dirs[i] = os.path.join(self.get_path(), i)
+        for dir_name in self.share_directories + ['logs', 'hints']:
```
fgelcer, line 66: i see `hints` is already part of `self.share_directories`
in this case, shouldn't `logs` be part of that too?

ccmlib/scylla_docker_cluster.py
```diff
@@ -118,13 +131,12 @@ def create_docker(self):
             if 'alternator_https_port' in scylla_yaml:
                 ports += f" -v {scylla_yaml['alternator_https_port']}"
 
-            mount_points = []
-            for directory in ['data', 'commitlogs', 'hints', 'view_hints', 'saved_caches']:
-                mount_points.append(
-                    f'-v {os.path.join(self.get_path(),directory)}:{os.path.join(self.base_data_path, directory)}')
+            mount_points = ' '.join([
+                f'-v {os.path.join(self.get_path(),directory)}:{os.path.join(self.base_data_path, directory)}'
+                for directory in self.share_directories])
```
fgelcer, line 136: liked it

fgelcer (commented):
few comments

ccmlib/scylla_docker_cluster.py
Orenef11: Fixed, thanks for notice

ccmlib/scylla_docker_cluster.py
Orenef11: Ill check it

Orenef11:
@fgelcer @fruch 
All comments are fixed

image
The last test also failed for "regular" ccm

ccmlib/scylla_docker_cluster.py
```diff
@@ -48,6 +49,8 @@ def _update_config(self):
 class ScyllaDockerNode(ScyllaNode):
     def __init__(self, *args, **kwargs):
         kwargs['save'] = False
+        # This line should appear before the "super" method
+        self.share_directories = ['data', 'commitlogs', 'conf', 'hints', 'view_hints', 'saved_caches', 'keys', 'logs']
```
fruch, line 53: please remove `conf` from here

ccmlib/scylla_docker_cluster.py
```diff
@@ -119,16 +130,23 @@ def create_docker(self):
...
+                       f" -d {self.cluster.docker_image} --smp 1 {seeds}"], stdout=PIPE, stderr=PIPE)
             self.pid = res.stdout.decode('utf-8').strip()
-
+            if self.cluster.snitch:
+                cassandra_rackdc_name = "cassandra-rackdc.properties"
+                run(["sudo", "bash", "-c",
+                     f"docker exec {self.pid}"
+                     f" cp {os.path.join(self.base_data_path, 'conf', cassandra_rackdc_name)}"
+                     f" {os.path.join('etc', 'scylla', cassandra_rackdc_name)}"])
+                run(['bash', '-c', f"docker exec {self.pid} supervisorctl restart scylla"],
```
fruch, line 148: don't restart scylla here, handle cassandra-rackdc.properties exactly as we are doing scylla.yaml
copy it out of the docker, and mount the exact file into place

restarting scylla with different configuration might take time, and some wrong data might be create and cached already.

fruch (changes requested):
handle `cassandra-rackdc.properties` exactly as `scylla.yaml`

ccmlib/scylla_docker_cluster.py
fgelcer: why? we want it mapped as a volume into the node in the end, no?

fruch: we don't want the directory mapped, only specific files

Orenef11: I cannt.
I got error when I did it

fgelcer: you still need to add here something like `conf/<conf_file>`
otherwise, it won't work

ccmlib/scylla_docker_cluster.py
fruch: this is handled in
https://github.com/scylladb/scylla-ccm/commit/3f3e3f80722e84751f99a985fee2265744cda9d4

fruch:
from this one we only need to fix for the keys

ccmlib/scylla_docker_cluster.py
Orenef11: Thanks!

PR #275, opened 2020-10-29
nodetool timeout: timeout missing on `communicate()` call

fruch:
a change in 373a62bd54d2377708f7f527937714730ed3a842 was missing
some small but importent part.


bhalevy (approved):
LGTM. Thanks!

PR #276, opened 2020-10-29
feature(docker_args): Support passing args to scylla docker

aleksbykov:
Added support for passing replace-address argument

### DIFF
--- a/ccmlib/scylla_docker_cluster.py
+++ b/ccmlib/scylla_docker_cluster.py
@@ -1,6 +1,6 @@
 import os
 from shutil import copyfile
-from subprocess import run, PIPE
+from subprocess import check_call, run, PIPE
 import logging
 
 import yaml
@@ -9,7 +9,6 @@ from ccmlib.scylla_cluster import ScyllaCluster
 from ccmlib.scylla_node import ScyllaNode
 from ccmlib.node import Status
 from ccmlib import common
-
 
 LOGGER = logging.getLogger("ccm")
 
@@ -132,12 +131,13 @@ class ScyllaDockerNode(ScyllaNode):
         with open(conf_file, 'w') as f:
             yaml.safe_dump(data, f, default_flow_style=False)
 
-    def create_docker(self):
+    def create_docker(self, args):
         # TODO: handle smp correctly via the correct param/api (or only via commandline params)
         # TODO: mount of the data dir
         # TODO: pass down the full command line params, since the docker ones doesn't support all of them ?
         # TODO: pass down a unique tag, with the cluster name, or id, if we have such in ccm, like test_id in SCT ?
         # TODO: add volume map to: hints, ...
+        # TODO: improve support for passing args to scylla docker.
 
         if not self.pid:
             node1 = self.cluster.nodelist()[0]
@@ -161,7 +161,7 @@ class ScyllaDockerNode(ScyllaNode):
 
             res = run(['bash', '-c', f"docker run {ports} "
                                      f"{mount_points} --name {self.docker_name}  "
-                                     f"-d {self.cluster.docker_image} --smp 1 {seeds}"], stdout=PIPE, stderr=PIPE, universal_newlines=True)
+                                     f"-d {self.cluster.docker_image} {seeds} {' '.join(args)}"], stdout=PIPE, stderr=PIPE, universal_newlines=True)
             self.pid = res.stdout.strip()
 
             if not res.returncode == 0:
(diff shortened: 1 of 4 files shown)


ccmlib/scylla_docker_cluster.py
```diff
@@ -17,6 +17,8 @@ class ScyllaDockerCluster(ScyllaCluster):
     def __init__(self, *args, **kwargs):
         super(ScyllaDockerCluster, self).__init__(*args, **kwargs)
         self.docker_image = kwargs['docker_image']
+        self.ipprefix = '172.17.0.'
+        self.ipformat = '{}%d'.format(self.ipprefix)
```
Orenef11, line 21: Why not use with f-string?

ccmlib/scylla_docker_cluster.py
```diff
@@ -44,6 +46,15 @@ def _update_config(self):
                 'docker_image': docker_image
             }, f)
 
+    def get_node_ip(self, nodeid):
+        try:
+            node = self.nodelist()[nodeid - 1]
+            return node.address()
+        except IndexError:
```
Orenef11, line 53: Error is a very expensive thing, it is better to change it to the following code:
```
if 0<= nodeid - 1 < len(self.nodelist()):
   return node.address()
 nodeid += len(self.nodelist())
return self.get_ipformat() % nodeid
```

ccmlib/scylla_docker_cluster.py
```diff
@@ -241,7 +257,17 @@ def _update_config(self):
 
     def _start_scylla(self, args, marks, update_pid, wait_other_notice,
                       wait_for_binary_proto, ext_env):
-        self.create_docker()
+
+        if "--replace-address-first-boot" in args:
```
Orenef11, line 261: when args is `None`, we will got error.

ccmlib/scylla_docker_cluster.py
```diff
@@ -322,7 +348,7 @@ def _update_pid(self, process):
         pass
 
     def get_tool(self, toolname):
-        return ['docker',  'exec', '-i',  f'{self.pid}', f'{toolname}']
+        return ['docker', 'exec', '-i', f'{self.pid}', f'{toolname}']
```
Orenef11, line 351: good job!

ccmlib/scylla_docker_cluster.py
```diff
@@ -352,9 +378,6 @@ def chmod(self, file_path, permissions):
             stdout=PIPE, stderr=PIPE)
 
 
-import subprocess
```
Orenef11, line 355: good job!

ccmlib/scylla_docker_cluster.py
```diff
@@ -365,5 +388,4 @@ def _logger_cmd(self) -> str:
         return f'docker logs -f {self._node.pid} >> {self._target_log_file} 2>&1 &'
 
     def start(self):
-        subprocess.check_call(['bash', '-c', self._logger_cmd])
```
Orenef11, line 368: good job!

ccmlib/scylla_docker_cluster.py
```diff
@@ -17,6 +17,8 @@ class ScyllaDockerCluster(ScyllaCluster):
     def __init__(self, *args, **kwargs):
         super(ScyllaDockerCluster, self).__init__(*args, **kwargs)
         self.docker_image = kwargs['docker_image']
+        self.ipprefix = '172.17.0.'
```
fgelcer, line 20: this will need to be updated, once we have our IP addresses

ccmlib/scylla_docker_cluster.py
```diff
@@ -44,6 +46,15 @@ def _update_config(self):
                 'docker_image': docker_image
             }, f)
 
+    def get_node_ip(self, nodeid):
```
fgelcer, line 49: i don't see this function being called

ccmlib/scylla_docker_cluster.py
```diff
@@ -99,12 +110,13 @@ def update_yaml(self):
         with open(conf_file, 'w') as f:
             yaml.safe_dump(data, f, default_flow_style=False)
 
-    def create_docker(self):
+    def create_docker(self, args=None):
```
fgelcer, line 113: so, to avoid what @Orenef11 mentions, we could start it as dict() or even an empty string

fgelcer (changes requested):
2 important changes that @Orenef11 pointed out

ccmlib/scylla_docker_cluster.py
Orenef11: @fgelcer @ShlomiBalalis 
Please dont method with `dict` or `list` variable because it's wrong in Python:
```
def blablab(a=[]):
   ...:     a.append(1)
   ...:     print(a)
   ...: 
blablab()
[1]
blablab()
[1, 1]
blablab()
[1, 1, 1]
blablab()
[1, 1, 1, 1]
```

It's the correct way:
```
     def create_docker(self, args=None):
        args = args or {}
```

ccmlib/scylla_docker_cluster.py
aleksbykov: I agree, them better. just save legacy pattern and forgot to change.

ccmlib/scylla_docker_cluster.py
```diff
@@ -44,6 +46,15 @@ def _update_config(self):
...
             }, f)
 
+    def get_node_ip(self, nodeid):
+        try:
+            node = self.nodelist()[nodeid - 1]
+            return node.address()
+        except IndexError:
+            if nodeid > len(self.nodelist()):
+                nodeid += len(self.nodelist())
+            return self.get_ipformat() % nodeid
```
fruch, line 56: why do we need this fallback ? I don't think we should supply those formatted address, since they aren't going to be true, we'll know the address only once we run the container,
any code that assume to know the address in advance is prone to error, and we should refactor it.

PR #278, opened 2020-11-10
ccmlib/scylla_repository: run_scylla_install_script: undo sed $prefix hack

bhalevy:
This hack didn't work when I added support in install.sh
for ubsan-suppressions.supp file as it looked for a `"$prefix` string
to substitute (with a leading double quotation mark).

As Avi explained in https://github.com/scylladb/scylla/issues/4949#issuecomment-724786194,
the hack is not required, so undo it by passing the right
`--prefix` option to install.sh, with no `--root` option, as
we're installing it relative to the host real root directory.

Refs scylladb/scylla#4949

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -238,9 +238,6 @@ def __get_dir():
 def run_scylla_install_script(install_dir, target_dir, package_version):
     scylla_target_dir = os.path.join(target_dir, 'scylla')
 
-    # FIXME: remove this hack once scylladb/scylla#4949 is fixed and merged
-    run('''sed 's|"$prefix|"$root/$prefix|' -i install.sh''', cwd=install_dir)
-
     if package_version <= packaging.version.parse('2'):
         # leave this for compatibility
         run('''sed -i '/systemctl --user/d' install.sh''', cwd=install_dir)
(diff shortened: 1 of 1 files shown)


bhalevy:
BYO test: https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/320

ccmlib/scylla_repository.py
```diff
@@ -249,9 +246,9 @@ def run_scylla_install_script(install_dir, target_dir, package_version):
         # on relocatable package format 2.1 or later, use --packaging instead of sed
         install_opt = ' --packaging'
 
-    run('''{0}/install.sh --root {1} --prefix {1} --prefix /opt/scylladb --nonroot{2}'''.format(
+    run('''{0}/install.sh --prefix {1} --nonroot{2}'''.format(
```
avikivity, line 249: You may want to add --packaging (maybe in a separate patch) to avoid installation of a scylla-server.service under the user's systemd configuration.

bhalevy: It's added based on the package version.
See https://github.com/scylladb/scylla-ccm/blob/74547ec5a3b92201fd6569c388eadf762d8c58a5/ccmlib/scylla_repository.py#L244-L253

fruch:
> BYO test: https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/320

seem like you'll need to rebase and rerun it

bhalevy:
> > BYO test: https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/320
> 
> seem like you'll need to rebase and rerun it

@fruch it didn't make byo run fail.  let's merge this pr since Avi already merged the scylla patch that depends on it to work (it should pass in release but will fail in debug since it won't find the suppressions file without this patch)

PR #281, opened 2020-12-14
Support node upgrade procedure

juliayakovlev:
Added support of node upgrade and method that upgrade
whole cluster

I opened this PR as Draft because of I want to run byo dtest and validate that it doesn't broke other tests

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -21,9 +21,11 @@ from six import print_
 from six.moves import xrange
 
 from ccmlib import common
-from ccmlib.node import Node
+from ccmlib.node import Node, NodeUpgradeError
+from ccmlib.node import Status
 from ccmlib.node import NodeError
 from ccmlib.node import TimeoutError
+from ccmlib.scylla_repository import setup
 
 
 def wait_for(func, timeout, first=0.0, step=1.0, text=None):
@@ -66,6 +68,8 @@ class ScyllaNode(Node):
     def __init__(self, name, cluster, auto_bootstrap, thrift_interface,
                  storage_interface, jmx_port, remote_debug_port, initial_token,
                  save=True, binary_interface=None, scylla_manager=None):
+        self._node_install_dir = None
+        self._node_scylla_version = None
         super(ScyllaNode, self).__init__(name, cluster, auto_bootstrap,
                                          thrift_interface, storage_interface,
                                          jmx_port, remote_debug_port,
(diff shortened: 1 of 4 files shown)


ccmlib/scylla_node.py
```diff
@@ -1141,3 +1178,97 @@ def no_pending_flushes():
...
+
+    def upgrade(self, upgrade_to_version):
+        self.upgrader.upgrade(upgrade_version=upgrade_to_version)
+
+class NodeUpgrader:
+
+    """
+    Upgrade node is supported when uses relocatable packages only
+    """
+    copy_suffix = '.org'
```
fruch, line 1205: where this is used ? and for what ?

ccmlib/scylla_node.py
```diff
@@ -795,52 +821,62 @@ def import_dse_config_files(self):
...
+            os.remove(dst)
+
         try:
             os.link(src, dst)
         except OSError as oserror:
             if oserror.errno == errno.EXDEV or oserror.errno == errno.EMLINK:
                 do_copy(src, dst, extra_perms)
             else:
-                raise RuntimeError("Unable to create hard link from %s to %s: %s" % (src, dst, oserror))
+                 raise RuntimeError("Unable to create hard link from %s to %s: %s" % (src, dst, oserror))
```
fruch, line 841: why the extra space here ?

fruch (approved):
LGTM

ccmlib/scylla_node.py
juliayakovlev: removed

ccmlib/scylla_node.py
juliayakovlev: fixed

fruch:
@juliayakovlev you got it working in BYO  ? (would be nice if you have a link, just for the confirmation)

ccmlib/scylla_node.py
```diff
@@ -1141,3 +1178,95 @@ def no_pending_flushes():
...
+        except Exception as exc:
+            self.node.node_install_dir = self.orig_install_dir
+            raise NodeUpgradeError("Failed to import executables files. %s" % exc)
+
+    def upgrade(self, upgrade_version: str):
+        """
+        :param upgrade_version: relocatables folder. Example: unstable/master:2020-11-18T08:57:53Z
+        """
+        self.scylla_version_for_upgrade = upgrade_version
+        cdir = self._setup_relocatable_packages()
```
amoskong, line 1246: do you plan to add a rollback()?

juliayakovlev: yes, I have such task

juliayakovlev:
> @juliayakovlev you got it working in BYO ? (would be nice if you have a link, just for the confirmation)

Yes, I ran byo to validat, that it doesn't broke dtest
Byo test: https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/363/

One test failed but it's not because of this CCM change

juliayakovlev:
@fruch  Can we merge it?

PR #282, opened 2020-12-20
fix: Fixing all ResourceWarning messages

Orenef11:
* Convert all `open` to` with open` to ensure the file will close
   after the used

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -187,13 +187,11 @@ class ScyllaNode(Node):
                 '-jar',
                 jmx_jar]
         log_file = os.path.join(self.get_path(), 'logs', 'system.log.jmx')
-        jmx_log = open(log_file, 'a')
         env_copy = os.environ
         env_copy['SCYLLA_HOME'] = self.get_path()
-        self._process_jmx = subprocess.Popen(args, stdout=jmx_log,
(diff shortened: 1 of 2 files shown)


ccmlib/scylla_node.py
```diff
@@ -187,13 +187,11 @@ def _start_jmx(self, data):
...
         log_file = os.path.join(self.get_path(), 'logs', 'system.log.jmx')
-        jmx_log = open(log_file, 'a')
         env_copy = os.environ
         env_copy['SCYLLA_HOME'] = self.get_path()
-        self._process_jmx = subprocess.Popen(args, stdout=jmx_log,
-                                             stderr=jmx_log,
-                                             close_fds=True,
-                                             env=env_copy)
+
+        with open(log_file, 'a') as jmx_log:
```
fruch, line 193: doesn't this mean or log file would be closed, right after this call ?
did you checked the the logs of jmx, and scylla are properly written when this context manager is used like that ?

fruch: FYI, `Popen(close_fds=True)` already close those files, regardless of the context manager.

fruch: in cases of errors, yep it might be left open

fruch (approved):
LGTM

PR #283, opened 2020-12-21
Add support to download of released version and latest release candidate

juliayakovlev:


ccmlib/common.py
```diff
@@ -692,3 +692,16 @@ def assert_jdk_valid_for_cassandra_version(cassandra_version):
...
         exit(1)
+
+
+def run_cmd_with_output(url):
+    output = subprocess.run(url.split(), capture_output=True, text=True)
+    return output.stdout, output.stderr, output.returncode
+
+
+def aws_bucket_ls(s3_url):
+    url = s3_url.replace('https://s3.amazonaws.com', 'aws s3 ls s3:/')
```
fruch, line 703: this mean that awscli need to be installed for this to work.
also this assume you have credentials to AWS, please make sure it works also without credentials (there are places we'll want to run it, that we don't have credentials)

juliayakovlev: changed to use boto3

ccmlib/common.py
```diff
@@ -692,3 +693,21 @@ def assert_jdk_valid_for_cassandra_version(cassandra_version):
     if cassandra_version >= '3.0' and get_jdk_version() < '1.8':
         print_('ERROR: Cassandra 3.0+ requires Java >= 1.8, found Java {}'.format(get_jdk_version()))
         exit(1)
+
+
+def aws_bucket_ls(s3_url):
+    bucket_object = s3_url.replace('https://s3.amazonaws.com/', '').split('/')
+    prefix = '/'.join(bucket_object[1:])
+    s3_conn = client('s3')
```
fruch, line 701: we still need to be able to use it without signing, and we need boto3 added to `setup.py` dependency  list 
https://stackoverflow.com/questions/34865927/can-i-use-boto3-anonymously

juliayakovlev: changed

fruch: @juliayakovlev we still need boto3 define here: https://github.com/scylladb/scylla-ccm/blob/next/setup.py#L27

fruch (approved):
LGTM

PR #284, opened 2020-12-24
Add method to clean runtime file after upgrade

juliayakovlev:
cassandra.in.sh file is created during ckuster creation and need be removed after upgrade


fruch (approved):
LGTM

PR #285, opened 2020-12-30
Raise priority of SCYLLA_VERSION in presence of packages in environment variable

juliayakovlev:
This PR will change download of relocatable's behavior.

Now if SCYLLA_JMX_PACKAGE / SCYLLA_TOOLS_JAVA_PACKAGE or SCYLLA_JAVA_TOOLS_PACKAGE / SCYLLA_CORE_PACKAGE environment variables are supplied (it happens when running dtest from Jenkins, packages are downloaded before the tests are started and Scylla version is some random and long string - SHA - that not include ":"), those packages will be used.
But this mechanism is wrong in case upgrade test: during the test we need to download relocatable's for different versions. But if  packages environment variables for tested version were supplied, the upgrade version won't be downloaded. It will use packages environment variables instead of version number.

This PR won't take packages environment variables if version with ":" is supplied

Running test with this code: [dtest-byo](https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/lastBuild/console)

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -144,31 +144,40 @@ def setup(version, verbose=True):
                 s3_url = ENTERPRISE_RELOCATABLE_URLS_BASE.format(type_n_version[0], s3_version)
             else:
                 s3_url = RELOCATABLE_URLS_BASE.format(type_n_version[0], s3_version)
+            packages = RelocatablePackages(scylla_jmx_package=os.path.join(s3_url,
+                                                                           f'{scylla_product}-jmx-package.tar.gz'),
+                                           scylla_tools_package=os.path.join(s3_url,
+                                                                             f'{scylla_product}-tools-package.tar.gz'),
+                                           scylla_package=os.path.join(s3_url,
+                                                                       f'{scylla_product}-package.tar.gz'))
+
         version = os.path.join(*type_n_version)
 
     cdir = version_directory(version)
 
     if cdir is None:
-        if not packages:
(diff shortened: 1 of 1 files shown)


ccmlib/scylla_repository.py
```diff
@@ -144,31 +144,40 @@ def setup(version, verbose=True):
...
-        if not packages:
-            packages = RelocatablePackages(scylla_jmx_package=os.path.join(s3_url, f'{scylla_product}-jmx-package.tar.gz'),
-                                           scylla_tools_package=os.path.join(s3_url, f'{scylla_product}-tools-package.tar.gz'),
-                                           scylla_package=os.path.join(s3_url, f'{scylla_product}-package.tar.gz'))
+        if not packages and not s3_url:
+            packages = RelocatablePackages(scylla_jmx_package=os.environ.get('SCYLLA_JMX_PACKAGE'),
+                                           # Try the old name for backward compatibility
+                                           scylla_tools_package=os.environ.get("SCYLLA_TOOLS_JAVA_PACKAGE") or
+                                                                os.environ.get("SCYLLA_JAVA_TOOLS_PACKAGE"),
+                                           scylla_package=os.environ.get('SCYLLA_CORE_PACKAGE')
```
fruch, line 164: you are not handling the `packages.scylla_tools_package` i.e. the commandline that would pass this

fruch: nevermind, now I understand where it's going

fruch (approved):
LGTM

PR #286, opened 2021-01-04
docker-ccm support

fruch:
merge of the branch that was used for the `docker-ccm` hackathon effort


fruch:
took this branch for a test drive: 
https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/384

PR #297, opened 2021-01-24
fix(scylla_extract_install_dir_and_mode): never to return a None scylla_mode

fruch:
in some cases when used a specific local files, it would return None, and not
default `release`


fruch:
@bhalevy hopefully soon we'll have a proper way, using `scylla version --mode`

bhalevy (approved):

PR #298, opened 2021-01-24
feature(actively_watch_logs_for_error): backport from cassandra-dtest

fruch:
As part of moving scylla-dtest to pytest, we want this function
so we can have closer feedback in logs of test when there are errors
in the scylla logs.

Ref: https://github.com/riptano/ccm/commit/24e14bc81109e3c9f5260aa46e0f57f85d548dc9
Ref: https://github.com/riptano/ccm/commit/50eb217ef1b5b4242e8526c350622691f12931e2

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -3,9 +3,11 @@
 import os
 import random
 import shutil
-import subprocess
+import threading
 import time
 from pkg_resources import parse_version
+from collections import OrderedDict, defaultdict
+from concurrent.futures import ThreadPoolExecutor
 
 import yaml
 from six import iteritems, print_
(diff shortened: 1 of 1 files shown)


ccmlib/cluster.py
```diff
@@ -121,6 +123,91 @@ def set_install_dir(self, install_dir=None, version=None, verbose=False):
...
+                self.cluster = cluster
+                self.req_stop_event = threading.Event()
+                self.done_event = threading.Event()
+                self.log_positions = defaultdict(int)
+
+            def start(self):
+                self.thread = self.executor.submit(self.run)
+
+            def scan(self):
+                errordata = OrderedDict()
```
Orenef11, line 154: There is no need to use `OrderedDict` because starting with `Python3.6`, the key insertion order is saved.

```
result = {key:key for key in range(10)}
list(range(10)) == list(result.keys())
True
```

ccmlib/cluster.py
```diff
@@ -121,6 +123,91 @@ def set_install_dir(self, install_dir=None, version=None, verbose=False):
...
+                        self.log_positions[node.name] = next_time_scan_from_mark
+                        if errors:
+                            errordata[node.name] = errors
+                except IOError as e:
+                    if 'No such file or directory' in str(e.strerror):
+                        pass  # most likely log file isn't yet written
+
+                    # in the case of unexpected error, report this thread to the callback
+                    else:
+                        errordata['log_scanner'] = [[str(e)]]
```
Orenef11, line 174: Why need two level of nesting?

Orenef11 (changes requested):

Orenef11:
LGTM

ccmlib/cluster.py
fruch: in theory CCM support python versions lower then 3.6 (in practice, we probably broken it long time ago)

ccmlib/cluster.py
fruch: I have now idea, It's a copy-paste from upstream

fruch: seem like it's goes hand with their implementation of `grep_log_for_errors`
https://github.com/riptano/ccm/blob/master/ccmlib/node.py#L2199
and also to ours...

the reason isn't clear to me, but not going to change that now.

PR #300, opened 2021-01-28
fix(scylla_repository): logic for setup of manager wasn't correct

fruch:
seems like the fix in 4571bc7c3835e8a1632d571f7beff707c96b4563 wasn't enough
and not accurate, and was breaking the manager dtest jobs
since `SCYLLA_EXT_OPTS` was defined, while the logic assume it would be defined
only when manger is used locally


fruch:
Jenkins staging job running with it:
https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/manager-dtest/3/

ShlomiBalalis (approved):

PR #301, opened 2021-02-02
fix cassandra bin path in node.run_sstablesplit

juliayakovlev:
The issue here is wrong path to cassandra/bin folder
It affects on the sstablesplit_test only. I didn't find other test that use it.
It's not connected to pytest.
Actually this test is not running, not part of the scylla_tests file

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -990,7 +990,7 @@ class Node(object):
             if debug:
                 cmd.append('--debug')
             cmd.append(f)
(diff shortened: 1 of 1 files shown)


roydahan:
@juliayakovlev what's the issue here?
Is it affecting the pytest branch only or in general?

juliayakovlev:
> @juliayakovlev what's the issue here?
> Is it affecting the pytest branch only or in general?

The issue here is wrong path to cassandra/bin folder
It affects on the sstablesplit_test only. I didn't find other test that use it. 
It's not connected to pytest. 
Actually this test is not running, not part of the scylla_tests file

bentsi (approved):

PR #302, opened 2021-02-03
fix(scylla_node): covering cases when network_interfaces records could be None

dkropachev:
Since cassandra dropped thrift support it is possible that at some point ccm
can get request to deploy nodes/cluster with thrift_interface=None
which is currently results in python errors

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -97,6 +97,12 @@ class ScyllaNode(Node):
         self._create_directory()
 
     @property
+    def ip_addr(self):
+        for ip_bucket in self.network_interfaces.values():
+            if ip_bucket and ip_bucket[0]:
+                return ip_bucket[0]
+
+    @property
     def node_install_dir(self):
         if not self._node_install_dir:
             self._node_install_dir = self.get_install_dir()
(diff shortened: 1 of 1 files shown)


fruch:
@dkropachev can you expend a bit more, when this case is happening ? also take a look at integration test in github actions, it's failing (not sure if related or not)

dkropachev:
> @dkropachev can you expend a bit more, when this case is happening ? also take a look at integration test in github actions, it's failing (not sure if related or not)

Example is auth tests:
https://github.com/scylladb/scylla-dtest/pull/1924/files

I work it around with following code in new_node:

```
    if cluster.__class__ in (Cluster, DseCluster):
        if cluster.cassandra_version() < '4':
            thrift = (ipformat % i, 9160)
    else:
        thrift = (ipformat % i, 9160)
```

dkropachev:
> @dkropachev can you expend a bit more, when this case is happening ? also take a look at integration test in github actions, it's failing (not sure if related or not)

Checked intergration tests, did not hapen localy, don't know how to restart it.

fruch (approved):
LGTM

fruch:
> > @dkropachev can you expend a bit more, when this case is happening ? also take a look at integration test in github actions, it's failing (not sure if related or not)
> 
> Checked intergration tests, did not hapen localy, don't know how to restart it.

I think I've found the root cause, I wasn't caching correctly
https://github.com/scylladb/scylla-ccm/pull/304/commits/87bfe729d1870b6e123cc405c5fd8d3e72fdf6b1

PR #308, opened 2021-02-14
Adjust scylla start timeout

bhalevy:
Currently we need to wait the full `node.watch_log_for` timeout
(600 seconds) to detect that a node failed during startup.
Pass the last_node process to `watch_log_for` so it
can fail as as it detects that the node terminated.

And in scylla_start: increase timeout in debug mode
to prevent test failures due to slow start when scylla is resharding sstables when started.

Test: materialized_views_test:TestMaterializedViews.{interrupt_build_process_with_resharding_low_to_half_test,interrupt_build_process_and_resharding_low_to_half_test}

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -109,7 +109,8 @@ class ScyllaCluster(Cluster):
                 if started:
                     last_node, _, last_mark = started[-1]
                     last_node.watch_log_for("node is now in normal status|Starting listening for CQL clients",
-                                            verbose=verbose, from_mark=last_mark)
+                                            verbose=verbose, from_mark=last_mark,
+                                            process=last_node._process_scylla)
                 mark = 0
                 if os.path.exists(node.logfilename()):
                     mark = node.mark_log()
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -267,7 +267,7 @@ class ScyllaNode(Node):
         return bool(self.grep_log("{}|{}".format(bootstrap_message, resharding_message), from_mark=from_mark))
 
     def _start_scylla(self, args, marks, update_pid, wait_other_notice,
-                      wait_for_binary_proto, ext_env):
+                      wait_for_binary_proto, ext_env, timeout=None):
         log_file = os.path.join(self.get_path(), 'logs', 'system.log')
         # In case we are restarting a node
         # we risk reading the old cassandra.pid file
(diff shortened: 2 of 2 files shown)


fruch:
@bhalevy I'm not sure I like the approach of those default being handled in CCM.
I would rather see it originating from the tests themselves.

and also the fact we are scattering those defaults across multiple places, is something that would make it a bit hard to follow.

I'm not sure on the effect it would have across the board, i.e. are we sure all the test would pass if the timeout is now 7min ?

bhalevy:
> @bhalevy I'm not sure I like the approach of those default being handled in CCM.
> I would rather see it originating from the tests themselves.
> 
> and also the fact we are scattering those defaults across multiple places, is something that would make it a bit hard to follow.
> 
> I'm not sure on the effect it would have across the board, i.e. are we sure all the test would pass if the timeout is now 7min ?

These defaults extend the existing default timeouts (that are present in ccm, like it or not).
It's possible to override the default timeout from the test, but it will require a more complicated change since the current cluster/node start api does not expose a timeout parameter.

nyh:
> @bhalevy I'm not sure I like the approach of those default being handled in CCM.
> I would rather see it originating from the tests themselves.
> 
> and also the fact we are scattering those defaults across multiple places, is something that would make it a bit hard to follow.
> 
> I'm not sure on the effect it would have across the board, i.e. are we sure all the test would pass if the timeout is now 7min ?

In my opinion:
1. Like Benny said, dtest or ccm should detect *immediately* when Scylla failed to boot and exited, and not wait for any timeout.
2. After doing this, there doesn't really need to be any separate timeout for booting Scylla as part of a test! We already have a timeout for **entire** test - starting Scylla is just one of many parts of the test. If this part takes so long that the entire test times out, we'll know that. Anyway, if CCM insists on a timeout for starting Scylla, you can use 1 hour. Or one month :-) You don't even need to think it as a configurable default, just think of it as "infinity that nobody will ever need to configure".

bhalevy:
> In my opinion:
> 
>     1. Like Benny said, dtest or ccm should detect _immediately_ when Scylla failed to boot and exited, and not wait for any timeout.
> 
>     2. After doing this, there doesn't really need to be any separate timeout for booting Scylla as part of a test! We already have a timeout for **entire** test - starting Scylla is just one of many parts of the test. If this part takes so long that the entire test times out, we'll know that. Anyway, if CCM insists on a timeout for starting Scylla, you can use 1 hour. Or one month :-) You don't even need to think it as a configurable default, just think of it as "infinity that nobody will ever need to configure".

I agree with you on the theoretical side, but practically speaking, we use dtest+ccm on jenkins and we want individual tests to fail in a reasonable time if intermediate steps they go through time out, otherwise when we have a systematic problem causes timeout of a large number of tests after a "sanity" timer (of 2 hours by default), the whole job will eventually timeout and coverage will be low since we weren't able to run all tests.

Hence we do you shorter timeouts for individual steps, like waiting for a node to be marked UP on other nodes when it starts, or waiting for it to start listening for CQL, or timeout for nodetool commands, etc.

bhalevy:
@fruch ping

fruch:
> > @bhalevy I'm not sure I like the approach of those default being handled in CCM.
> > I would rather see it originating from the tests themselves.
> > and also the fact we are scattering those defaults across multiple places, is something that would make it a bit hard to follow.
> > I'm not sure on the effect it would have across the board, i.e. are we sure all the test would pass if the timeout is now 7min ?
> 
> These defaults extend the existing default timeouts (that are present in ccm, like it or not).
> It's possible to override the default timeout from the test, but it will require a more complicated change since the current cluster/node start api does not expose a timeout parameter.

I'll merge it for, but I still think this logic is currently hidden inside CCM, and should be revisited to be controlled by the test.

PR #313, opened 2021-03-01
scylla_cluster: stop_nodes: ignore node.is_running

bhalevy:
In addition to 23880da688cc8f3bc2b0ba5d17363c4edd42a138
cluster.stop_nodes also needs to ignore node.is_running
to kill scylla-jmx, e.g. when the cluster is stopped
in the end of a dtest.

This change calls node.do_stop on all nodes to be stopped,
regardless if they are running or not (as it is possible
that the scylla process would have already exited, or
failed to start), and then call node.wait_until_stopped
if wait=True on all the above nodes.  wait_until_stopped
makes sure to kill all processes, and to get a core dump
from scylla if dump_core=True (derived from gently).

Dtest: secondary_indexes_test.py:TestSecondaryIndexes.test_insert_data_after_recreating_ks

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -163,14 +163,13 @@ class ScyllaCluster(Cluster):
             marks = [(node, node.mark_log()) for node in other_nodes if node.is_live()]
 
         # stop all nodes in parallel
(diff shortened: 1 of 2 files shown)


fruch (approved):
LGTM

PR #315, opened 2021-03-04
fix(ccm add): support `ccm add` for relocatable package

fruch:
Seems that since relocatable was introduced, not many people were using
the ccm commandline for it, and this was broken for relocatable packages
now the `scylla_version` is persisted and any other following ccm commands
now would know it's relocatable version, and could work correctly.


fgelcer (approved):
LGTM

PR #317, opened 2021-03-22
node: use the kernel page cache for ccm scylla nodes

avikivity:
Scylla nodes started by ccm run in an overprovisioned environment
with limited performance disks. To reduce load on the disks, use
the new --kernel-page-cache option. This will reduce the likelihood
of timeouts on more intensive tests.


avikivity:
@scylladb/qa-maintainers ping

roydahan (approved):

Issue #318, opened 2021-03-31
Starting scylla with --kernel-page-cache is not supported by older scylla versions


bhalevy:
PR #317 caused a regression in rolling_upgrade_test.

https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/780/artifact/logs-all.release.2/1617166227559_rolling_upgrade_test.RollingUpgradeTest.test_rolling_upgrade/node1.log:
```
Scylla version 4.3.0-0.20210110.000585522 with build-id 7fab6c35a5d403f3416292b37c9a9aa94fe4db1f starting ...
command used: "/jenkins/workspace/scylla-master/dtest-release/scylla/.dtest/dtest-5ns8k1n5/test/node1/bin/scylla --options-file /jenkins/workspace/scylla-master/dtest-release/scylla/.dtest/dtest-5ns8k1n5/test/node1/conf/scylla.yaml --log-to-stdout 1 --api-address 127.0.51.1 --collectd-hostname d8061d688155.node1 --smp 2 --memory 1024M --abort-on-seastar-bad-alloc --abort-on-lsa-bad-alloc 1 --abort-on-internal-error 1 --developer-mode true --default-log-level info --collectd 0 --overprovisioned --prometheus-address 127.0.51.1 --unsafe-bypass-fsync 1 --kernel-page-cache 1"
parsed command line options: [options-file: /jenkins/workspace/scylla-master/dtest-release/scylla/.dtest/dtest-5ns8k1n5/test/node1/conf/scylla.yaml, log-to-stdout: 1, api-address: 127.0.51.1, collectd-hostname: d8061d688155.node1, smp: 2, memory: 1024M, abort-on-seastar-bad-alloc, abort-on-lsa-bad-alloc: 1, abort-on-internal-error: 1, developer-mode: true, default-log-level: info, collectd: 0, overprovisioned, prometheus-address: 127.0.51.1, unsafe-bypass-fsync: 1, kernel-page-cache, (positional) 1]
error: unrecognised option '--kernel-page-cache'
```

The option was recently introduced in scylladb/scylla@8785dd62cb740522d80eb12f8272081f85be9b7e
and is not available in any release up to and including 4.4.0.

bhalevy:
Hit the same issue also in:
[...]

PR #321, opened 2021-04-04
Starting scylla with --kernel-page-cache from 4.5 version and above only

juliayakovlev:
This commit fixes the issue: https://github.com/scylladb/scylla-ccm/issues/318

Start Scylla node with --kernel-page-cache argument if the version is 4.5 or higher

Upgrade tests are passed: https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/502/console

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -214,7 +214,7 @@ class Node(object):
         """
         Returns the path to the cassandra source directory used by this node.
         """
-        if self.__install_dir is None:
+        if not self.__install_dir:
             return self.cluster.get_install_dir()
         else:
             common.validate_install_dir(self.__install_dir)
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -11,6 +11,7 @@ import stat
 import subprocess
 import time
 import threading
+from distutils.version import LooseVersion
 
 import psutil
 import yaml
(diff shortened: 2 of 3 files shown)


fruch:
@juliayakovlev can you take a look at the unit tests running in github actions ? seem like this new code is failing the docker-ccm, and would need a different way to get the version for it.

juliayakovlev:
> @juliayakovlev can you take a look at the unit tests running in github actions ? seem like this new code is failing the docker-ccm, and would need a different way to get the version for it.

@fruch I am checking it now. 
It failed on `common.isDse` function, that has been called from old  `ccmlib.node.Node.get_install_dir` without any parameter. For now it's not clear why it fails

fruch (approved):
LGTM, once it would pass the test, I'll merge it.

juliayakovlev:
> LGTM, once it would pass the test, I'll merge it.

Upgrade and rolling tests are passed:
https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/504/console

fruch:
> > LGTM, once it would pass the test, I'll merge it.
> 
> @fruch
> Upgrade and rolling tests are passed:
> https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/504/console

I'm fixing one more small thing cause the test in the PR were failing (they were using 666.development version)

kbr-:
This broke my test workflow :(
getting these errors:
```
'/home/kbraun/dev/scylla/scylla-core-package/SCYLLA-VERSION-FILE' wasn't found
```
I guess I can reverse engineer the code and check what I need to put in `SCYLLA_VERSION_FILE` to make it work, but it would be nice if such breaking changes were documented.

PR #322, opened 2021-04-12
Get major release scylla version

juliayakovlev:
Get major release scylla version and start Scylla with '--kernel-page-cache'
parameter if major release >= 4.5

Fix for commit https://github.com/scylladb/scylla-ccm/commit/86f5f5b335ed2ad079f2b2d3d2dbdefcc187dc53#commitcomment-49399513


bhalevy (changes requested):
Nack, sorry.
This tries to reinvent the wheel.
We don't compare versions using plain strings nor floating point numbers
since e.g. version '4.20' should be greater than version '4.3'.
`parse_version` does that correctly.

ccmlib/scylla_node.py
```diff
@@ -577,8 +576,9 @@ def start(self, join_ring=True, no_wait=False, verbose=False,
 
         # The '--kernel-page-cache' was introduced by
         # https://github.com/scylladb/scylla/commit/8785dd62cb740522d80eb12f8272081f85be9b7e from 4.5 version
-        current_node_version = self.node_install_dir_version
-        if current_node_version and LooseVersion('666.development') < LooseVersion(current_node_version) >= LooseVersion('4.5'):
+        major_node_scylla_version = self.get_release_major_scylla_version()
+
+        if major_node_scylla_version and float(major_node_scylla_version) >= 4.5:
```
nyh, line 581: If you add this get_release_major_scylla_version() function, I suggest you simplify its usage by ensuring it always returns a float, then you can just have "if major_node_scylla_version >= 4.5".

However, see also below my question about enterprise version, which I'm not sure how this code (or the previous code it replaces) handles....

ccmlib/scylla_node.py
```diff
@@ -1240,6 +1240,12 @@ def flush(self):
         self.nodetool("flush")
         self._wait_no_pending_flushes()
 
+    def get_release_major_scylla_version(self):
```
nyh, line 1243: Would be useful to have a comment on this, saying that this returns something like "4.5".
Is "major version" the official name we use for something like "4.5"?

A more serious issue: how does this work for enterprise versions?

nyh:
> Nack, sorry.
> This tries to reinvent the wheel.
> We don't compare versions using plain strings nor floating point numbers
> since e.g. version '4.20' should be greater than version '4.3'.
> `parse_version` does that correctly.

By the way, how do we handle enterprise versions? Is 2020.1 more or less than 4.5? :-(

juliayakovlev:
> Nack, sorry.
> This tries to reinvent the wheel.
> We don't compare versions using plain strings nor floating point numbers
> since e.g. version '4.20' should be greater than version '4.3'.
> `parse_version` does that correctly.

Ok, will change

ccmlib/scylla_node.py
```diff
@@ -1240,6 +1242,12 @@ def flush(self):
...
+        version = self.get_node_scylla_version()
+        find = re.search(r"^(\d+.\d+).*", version)
+
+        return find.groups()[0] if find else ''
```
nyh, line 1249: I think this interface could have been less clunky if this function returned the result of parse_version(), and also fail if it couldn't find the version (I don't see what's the point in returning '' and hiding the error).

bhalevy:
> > Nack, sorry.
> > This tries to reinvent the wheel.
> > We don't compare versions using plain strings nor floating point numbers
> > since e.g. version '4.20' should be greater than version '4.3'.
> > `parse_version` does that correctly.
> 
> By the way, how do we handle enterprise versions? Is 2020.1 more or less than 4.5? :-(

Unfortunately yes, according to the versioning rules we adopted (https://www.python.org/dev/peps/pep-0440/).
That was one of the reason to change the version of master/enterprise from `666-development` and `9999.something` to e.g. `4.6.dev` and `2022.1.dev` respectively.

Also, I proposed https://github.com/scylladb/scylla-enterprise/issues/1486 for getting the base OSS version from the scylla-enterprise binary so it can be compared against it.  That said, I'm not sure that relying on the version strings is the best way to control the test features anyhow :)

Regarding the command line option at hand, the test could probe its support in other ways, e.g. by running `scylla --help` and looking for it in the output, or by handling the error scylla produces when it's started with the unsupported flags.

For other features to be tested, we could have a formal test api that can be used to probe the binary what features it supports and what version _of the feature_ so the test can tweak itself accordingly, rather than guessing what flavor of the feature is supported based on the scylla/scylla-enterprise version.

ccmlib/scylla_node.py
bhalevy: Agreed.

Also, the regular expression uses a `.` wildcard, and you probably want a `\.` instead.

You can refactor `_get_scylla_version` to reuse the regular it uses to match the PEP-440 conforming version part.
For example:
```python
>>> s='4.6.dev-0.20210412.d1ce1b971'
>>> pattern=r'^((\d+!)?\d+([.-]\d+)*([a-z]+\d*)?([.-](post|dev|rc)\d*)*)(-.*)$'
>>> m=re.match(pattern, s)
>>> m.group(1)
'4.6.dev'
```

bhalevy (changes requested):

bhalevy:
ping @juliayakovlev

fruch:
fixed by:
#338

PR #325, opened 2021-04-21
feature(manager-agent): function to execute download-files command

ShlomiBalalis:
Added agent_download_files function.
The function receives a node object, bucket location list and a snapshot tag
and afterwards uses the agent's download-files command to download the snapshot
files to the node's upload directory.
optional -  keyspace_filter_list: list of glob strings. only the files of keyspaces
that match the glob filters will be downloaded.

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -457,3 +457,28 @@ class ScyllaManager:
         if exit_status != 0:
             raise Exception(" ".join(args), exit_status, stdout, stderr)
         return stdout, stderr
+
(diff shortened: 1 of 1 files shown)


roydahan:
Please fix commit message - length of title and missing description

roydahan:
I still don't see a meaningful commit message that explain something

roydahan:
@ShlomiBalalis I didn't ask you to copy the function descritption to the commit message.
The commit message should include things like:
The reason we need to add this code, why we need it now and not previously, which feature it's going to test and since what release, which code is using it in the dtest, etc....

roydahan (approved):

roydahan:
@ShlomiBalalis please ping here for backport once it passes on master.

PR #331, opened 2021-06-16
increment supported relocatable package version

syuu1228:
Latest version is 3.
We don't need fixes for scylla-ccm on this version.


fruch (approved):
LGTM

PR #332, opened 2021-06-18
fix(scylla_node): fix upgrade from local installation

michoecho:
import_bin_files() doesn't pass replace=replace to hard_link_or_copy
when copying scylla-jmx from a local installation, so node upgrade from
a local installation fails with a "file exists" error. Fix that.

Discovered when running upgrade_test.py with a local Scylla build.


fruch (approved):
LGTM

PR #335, opened 2021-07-28
scylla_node: start: set --max-networking-io-control-blocks=100 by default

### LINKED ISSUE #334: adjust max-networking-io-control-blocks=100 by default
scylladb/scylla@2cfc517874e98c5780c1b1b4c38440a8123f86f6 sets `max_networking_aio_io_control_blocks` to 50000 by default, causing us to run out of aio iocb slots and fail starting in `setup_aio_context`.

The signature for the error is:
```
ERROR 2021-07-20 15:54:51,154 [shard 7] seastar - Could not setup Async I/O: Resource temporarily unavailable. The most common cause is not enough request capacity in /proc/sys/fs/aio-max-nr. Try increasing that number or reducing the amount of logical CPUs available for your application
```

bhalevy:
scylladb/scylla@2cfc517 sets max_networking_aio_io_control_blocks to
50000 by default, causing us to run out of aio iocb slots and fail
starting in setup_aio_context.

Reduce it by default to 100 that is sufficient for testing
if not given a different value via jvm_args.

Fixes #334

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -595,6 +595,8 @@ class ScyllaNode(Node):
         if replace_address:
             args += ['--replace-address', replace_address]
         args += ['--unsafe-bypass-fsync', '1']
+        if '--max-networking-io-control-blocks' not in args:
+            args += ['--max-networking-io-control-blocks', '100']
 
         # The '--kernel-page-cache' was introduced by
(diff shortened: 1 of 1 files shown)


bhalevy:
@fruch it's unclear why the CI test timed out.
FWIW, the change passed locally for me with nosetests.

bhalevy:
@fruch ping

fruch (approved):
LGTM

PR #337, opened 2021-08-04
ccm: explicitly use Python3 interpreter

### LINKED ISSUE #336: 'ccm list' fails with "SyntaxError: invalid syntax" on Ubuntu 20 box
HEAD: f1a1b775b7d6fddf6adf458dd179e38e70b18c07

*Description*
```
$ ccm list                                                               
  File "/home/vladz/work/scylla-ccm/ccm", line 7, in <module>                                                                  
    from ccmlib import common                                                                                                  
  File "/home/vladz/work/scylla-ccm/ccmlib/common.py", line 489                                                                
    raise TimeoutError(f"Relocatables download still runs in parallel from another test after 60 min. "
                                                                                                      ^
SyntaxError: invalid syntax
```

The reason is the first line in `ccm`:
```
#!/usr/bin/env python
```

And `/usr/bin/python` points to `python2.7`.
And f-strings are a Python3-only feature.

vladzcloudius:
On some (all?) modern Linux distros /usr/bin/python
still points to python2.

At the same time our code uses some Python3-specific features
like f-strings.

Hence we (still) need to be explicit about Python3.

Fixes #336

### DIFF
--- a/ccm
+++ b/ccm
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import sys
 


nyh (approved):
Looks good. It's not true that in all modern Linux distros python still points to python2 - it points to python3 on recent Fedora. But nevertheless, "python3" exists on every Linux distro, and all of the Python scripts and tests in scylla.git also use "python3" to run Python, so ccm can do the same.

bhalevy (approved):

bhalevy:
@scylladb/ccm-maint ping

fruch (approved):
LGTM

PR #338, opened 2021-08-09
scylla_node: start: do not rely on SCYLLA-VERSION-FILE

### LINKED ISSUE #318: Starting scylla with --kernel-page-cache is not supported by older scylla versions

bhalevy:
Pull #321 introduced code to check the node version using
SCYLLA-VERSION-FILE (and LooseVersion). However,
when testing scylla compiled from source,
SCYLLA-VERSION-FILE is not available.

Try to get the node version using node_install_dir_version
and if not successful, fallback to cluster.version()
to decide whether to add the --kernel-page-cache command
line option.

Also, it uses parse_version rather than LooseVersion
for version comparison as we already changed version
checking code everywhere to use the latter that is
suitable for our versioning conventions, unlike
LooseVersion.

`node_install_dir_version` was changed from `@property`
to a regular method so it can call self.debug in
case SCYLLA_VERSION_FILE wasn't found.

Refs #321
Fixes #318

Dtest: bootstrap_test.py:TestBootstrap.start_stop_test (dev, against scylla 4.4, 4.5, 4.6.dev)

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -11,7 +11,7 @@ import stat
 import subprocess
 import time
 import threading
-from distutils.version import LooseVersion
+from pkg_resources import parse_version
 
 import psutil
 import yaml
@@ -429,15 +429,14 @@ class ScyllaNode(Node):
 
         return java_up
 
-    @property
     def node_install_dir_version(self):
         if not self.node_install_dir:
-            return ""
+            return None
 
         scylla_version_file_path = os.path.join(self.node_install_dir, CORE_PACKAGE_DIR_NAME, SCYLLA_VERSION_FILE)
         if not os.path.exists(scylla_version_file_path):
-            print(f"'{scylla_version_file_path}' wasn't found")
-            return ""
+            self.debug(f"'{scylla_version_file_path}' wasn't found")
+            return None
 
         with open(scylla_version_file_path, 'r') as f:
             version = f.readline()
@@ -598,8 +597,8 @@ class ScyllaNode(Node):
 
         # The '--kernel-page-cache' was introduced by
         # https://github.com/scylladb/scylla/commit/8785dd62cb740522d80eb12f8272081f85be9b7e from 4.5 version
-        current_node_version = self.node_install_dir_version
-        if current_node_version and LooseVersion('666.development') < LooseVersion(current_node_version) >= LooseVersion('4.5'):
+        current_node_version = self.node_install_dir_version() or self.cluster.version()
+        if parse_version(current_node_version) >= parse_version('4.5.dev'):
             args += ['--kernel-page-cache', '1']
 
         ext_env = {}


bhalevy:
ping @juliayakovlev @fruch

ccmlib/scylla_node.py
```diff
@@ -598,8 +585,7 @@ def start(self, join_ring=True, no_wait=False, verbose=False,
 
         # The '--kernel-page-cache' was introduced by
         # https://github.com/scylladb/scylla/commit/8785dd62cb740522d80eb12f8272081f85be9b7e from 4.5 version
-        current_node_version = self.node_install_dir_version
-        if current_node_version and LooseVersion('666.development') < LooseVersion(current_node_version) >= LooseVersion('4.5'):
+        if parse_version(self.cluster.version()) >= parse_version('4.5.dev'):
```
juliayakovlev, line 588: `self.cluster.version()` function runs `nodetool`. But in this step the Scylla node is not started yet. So the nodetool can't run. This is the reason why property `node_install_dir_version` has been added.
How did you get it working?

bhalevy: I believe that cluster.version() runs https://github.com/scylladb/scylla-ccm/blob/05dc1327966bffcc5bed144f1a7437c2ab85c862/ccmlib/cluster.py#L221-L222

If you refer to https://github.com/scylladb/scylla-ccm/blob/next/ccmlib/scylla_cluster.py#L266-L270, it's `ScyllaManager.version`, is that what you meant?

bhalevy: @juliayakovlev ping

juliayakovlev: Sorry for delay.
No, this is version of the node, not cluster:
https://github.com/scylladb/scylla-ccm/blob/f4f1f5c11b95c6eef3d4957e162953dc11956ba0/ccmlib/node.py#L1328

bhalevy: @juliayakovlev but I changed the call to `self.cluster.version()`

juliayakovlev: And which cluster version will be during upgrade cluster, when one node will run different version?

bhalevy:
In v2:
- try node_install_dir_version first and fall back to cluster.version() if node version not found.
  - this should work when running upgrade test with installed relocatable packages.

Cc @juliayakovlev

ccmlib/scylla_node.py
bhalevy: Ok, fair enough, I pushed a new version that tries node_install_dir_version first and
falls back to cluster.version() only if the former could not find the version.

juliayakovlev (approved):

bhalevy:

roydahan (approved):

PR #339, opened 2021-08-09
fix(manager): fixing myriad of issues that came with 2.5

ShlomiBalalis:


ccmlib/scylla_repository.py
```diff
@@ -316,7 +316,10 @@ def setup_scylla_manager(scylla_manager_package=None):
                     mkdir -p dist/etc
                     mkdir schema
                     cp -r ./etc/scylla-manager/* ./dist/etc/
-                    cp -r ./etc/scylla-manager/cql ./schema/
+                    if [ -d "./etc/scylla-manager/cql" ]
```
juliayakovlev, line 319: What may happen if schema won't be copied? Does it has affect?

juliayakovlev (commented):
one question

ccmlib/scylla_repository.py
ShlomiBalalis: Not at all, according to michal they moved it into the binary. That's also why we don't config the migrate_dir parameter anymore

juliayakovlev (approved):

bentsi (approved):

roydahan:
@ShlomiBalalis Do we need to backport it to manager-2.5 branch?
Do we even have branches for manager in ccm?

ShlomiBalalis:
> @ShlomiBalalis Do we need to backport it to manager-2.5 branch?
> Do we even have branches for manager in ccm?

@roydahan We do not, we only use master

PR #340, opened 2021-08-12
fix(scylla_version): modified to be able to use master builds

ShlomiBalalis:
Master builds use a naming scheme that is a bit different from the next builds,
and using such build will cause an exception because of it.
Specifically, the scylla_package is called scylla-x86_64-package.tar.gz instead
of scylla-package.tar.gz.
I modified the setup function to support both options.


yaronkaikov (approved):
LGTM

roydahan (approved):

PR #346, opened 2021-11-18
test_scylla_cmds.py: use new manager packages

fruch:
since ddd36ff003781142443f59159a683fa7adb4f55e the url in the test
doesn't work anymore, and needed to change.


fruch:
failing cause of #335 was merged, fix in #347

PR #347, opened 2021-11-18
Fix integration suite to pass

fruch:
since b50e43f9325de43a90105346ae38437ecf178cc1 was introduced
the old scylla used isn't working, moving to newer version of scylla

moving to newer version of the docker image has other issues, disabling the
tests for now

### DIFF
--- a/tests/test_scylla_cmds.py
+++ b/tests/test_scylla_cmds.py
@@ -14,7 +14,7 @@ LOGGER = logging.getLogger(__name__)
 
 cluster_params = pytest.mark.parametrize(
     'cluster_under_test',
-    (pytest.param('ccm_docker_cluster', marks=pytest.mark.docker),
+    (pytest.param('ccm_docker_cluster', marks=[pytest.mark.docker, pytest.mark.skip(reason="ccm docker support broke in master, need to fix")]),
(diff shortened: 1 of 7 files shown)


tests/test_scylla_cmds.py
```diff
@@ -14,7 +14,7 @@
 
 cluster_params = pytest.mark.parametrize(
     'cluster_under_test',
-    (pytest.param('ccm_docker_cluster', marks=pytest.mark.docker),
+    (pytest.param('ccm_docker_cluster', marks=[pytest.mark.docker, pytest.mark.skip(reason="ccm docker support broke in master, need to fix")]),
```
fgelcer, line 17: aren't you fixing it in this PR?

tests/test_scylla_docker_cluster.py
```diff
@@ -3,6 +3,7 @@
 
 
 @pytest.mark.docker
+@pytest.mark.skip(reason="ccm docker support broke in master, need to fix")
```
fgelcer, line 6: and here?

fgelcer (approved):
LGTM

tests/test_scylla_cmds.py
fruch: Nope, too many issues

PR #349, opened 2021-11-26
add padding zero to node names

soyacz:
when there's more than 9 nodes in cluster,
nodes with number >9 are improperly sorted by nodelist() method
this causes errors in some tests when test relies on last node
in nodelist is the one with highest number

This commit adds padding zero to node number. E.g:
node01, node09, node10, node11

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -293,7 +293,7 @@ class Cluster(object):
         binary = None
         if parse_version(self.version()) >= parse_version('1.2'):
             binary = self.get_binary_interface(i)
-        node = self.create_node(name='node{}'.format(i),
+        node = self.create_node(name='node{0:02d}'.format(i),
                                 auto_bootstrap=auto_bootstrap,
                                 thrift_interface=self.get_thrift_interface(i),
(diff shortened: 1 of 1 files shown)


bhalevy:
And what if we create a cluster of 100 nodes or more?
It'd be better to pass a parameter like the number of digits or maybe the whole name, since the cluster.populate levels knows in advance how many nodes to provision.

fruch:
@soyacz this change is gonna break a bunch of tests in dtest that assume the name of the first node is `node1`

You might change the `self.nodes` to OrderedDict, and the order would be preserved by the order of creation of the nodes. (i.e. remove the sorted from `nodelist`)

soyacz:
> @soyacz this change is gonna break a bunch of tests in dtest that assume the name of the first node is `node1`
> 
> You might change the `self.nodes` to OrderedDict, and the order would be preserved by the order of creation of the nodes. (i.e. remove the sorted from `nodelist`)

Good idea @fruch - I'll close this PR and open new one with OrderedDict

soyacz:
https://github.com/scylladb/scylla-ccm/pull/351

PR #351, opened 2021-11-29
fix(cluster.py): fix nodelist ordering

soyacz:
When there's more than 9 nodes in cluster,
nodes with number >9 are improperly sorted by nodelist() method
this causes errors in some tests when test relies on last node
in nodelist is the one with highest number

This fix converts node's dictionary to OrderedDict


bhalevy (approved):
LGTM

fruch (approved):
LGTM

PR #353, opened 2021-12-06
Increase the max-networking-io-control-blocks parameter

### LINKED ISSUE #352: Increase the max-networking-io-control-blocks parameter
It was reduced to 100 in b50e43f9325de43a90105346ae38437ecf178cc1 to deal with #334, but apparently the low value causes https://github.com/scylladb/seastar/issues/976.
We should bump it back up to a reasonable value (10000?)

bhalevy:
It was reduced to 100 in b50e43f to deal with #334,
and then in 17567b52e it was restricted to version 4.7.dev and up.
But apparently the low value causes scylladb/seastar#976

This change increases it to a more reasonable value of 1000
and changes the version restriction to 4.6, since the bump
up to 50000 happened in scylladb/scylla@2cfc517874
which is in branch-4.6

Fixes #352

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -601,8 +601,8 @@ class ScyllaNode(Node):
         if parse_version(current_node_version) >= parse_version('4.5.dev'):
             args += ['--kernel-page-cache', '1']
 
-        if parse_version(current_node_version) >= parse_version('4.7.dev') and '--max-networking-io-control-blocks' not in args:
-            args += ['--max-networking-io-control-blocks', '100']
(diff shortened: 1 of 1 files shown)


bhalevy:
@scylladb/scylla-maint ping
We're hitting https://github.com/scylladb/seastar/issues/976 frequently in master dtest-release

denesb:
This repo is maintained by the dtest developers, you will have better luck pinging them maybe.

bhalevy:
ping @scylladb/dtest-maintainers

fruch (approved):
LGTM

PR #354, opened 2021-12-09
fix(scylla_repository.py): make it work properly with arch

dkropachev:
https://trello.com/c/Hc4c6Cqy/4354-fix-ccm-to-support-new-relocatible-files-layout

From 4.6 we started releasing arm arch along with x86_64.
Because of that some of the package names had changed from `scylla-package.tar.gz` to `scylla-x86_64-package.tar.gz`
on both unstable and releases.

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -55,7 +55,7 @@ class RelocatablePackages(NamedTuple):
     scylla_jmx_package: str
 
 
-def release_packages(s3_url, version):
+def release_packages(s3_url, version, arch='x86_64'):
     """
     Choose RELEASE relocatable packages for download.
     It covers 3 cases:
@@ -78,7 +78,14 @@ def release_packages(s3_url, version):
         raise RuntimeError(
             f"Failed to get release packages list for version {major_version}. URL: {s3_url}.")
 
-    all_packages = [name for name in files_in_bucket if 'jmx' in name or 'tools' in name or 'scylla-package' in name]
+    scylla_package_mark = 'scylla-package'
+
+    for file_name in files_in_bucket:
+        if f'scylla-{arch}-package' in file_name:
+            scylla_package_mark = f'scylla-{arch}-package'
+            break
+
+    all_packages = [name for name in files_in_bucket if 'jmx' in name or 'tools' in name or scylla_package_mark in name]
 
     candidates = []
     # Search for released version packages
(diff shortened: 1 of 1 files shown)


ccmlib/scylla_repository.py
```diff
@@ -155,6 +165,7 @@ def setup(version, verbose=True):
     s3_url = ''
     type_n_version = version.split(':', 1)
     scylla_product = os.environ.get('SCYLLA_PRODUCT', 'scylla')
+    scylla_arch = os.environ.get('SCYLLA_ARCH', 'x86_64')
```
aleksbykov, line 168: is this env variable set? Should it be passed/set to dtest docker?

ccmlib/scylla_repository.py
```diff
@@ -117,11 +124,14 @@ def release_packages(s3_url, version):
     for package in release_packages:
         # Expected packages names (examples):
         #  'scylla-jmx-package-4.3.0-0.20210110.000585522.tar.gz'
-        #  'scylla-package-4.3.0-0.20210110.000585522.tar.gz'
+        #  'scylla-package-4.3.0-0.20210110.000585522.tar.gz' or
+        #   'scylla-x86_64-package-4.3.0-0.20210110.000585522.tar.gz'
         #  'scylla-tools-package-4.3.0-0.20210110.000585522.tar.gz'
```
aleksbykov, line 129: lets change to 4.6.0 scylla version

ccmlib/scylla_repository.py
dkropachev: Not necessarily, currently we are targeting only x86_64, but when we would want to target arm, we should add it there.

ccmlib/scylla_repository.py
dkropachev: done

ccmlib/scylla_repository.py
```diff
@@ -78,7 +78,14 @@ def release_packages(s3_url, version):
         raise RuntimeError(
             f"Failed to get release packages list for version {major_version}. URL: {s3_url}.")
 
-    all_packages = [name for name in files_in_bucket if 'jmx' in name or 'tools' in name or 'scylla-package' in name]
+    scylla_package_mark = 'scylla-package'
+
+    for file_name in files_in_bucket:
+        if f'scylla-{arch}-package' in file_name:
```
fgelcer, line 84: this change won't allow us to run builds from 2 weeks ago or older

juliayakovlev (approved):
It looks ok. Did you try this with different versions?

dkropachev:
> It looks ok. Did you try this with different versions?

Yes, on both releases and unstable relocatables

PR #355, opened 2021-12-12
common: fix `scylla_extract_mode` to handle new package name

fruch:
since now package filename has arch where the mode was, this
function was return wrong the arch in some cases

### DIFF
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -1,0 +1,19 @@
+from ccmlib.common import scylla_extract_mode
+
+
+def test_scylla_extract_mode():
+    assert scylla_extract_mode("build/dev") == 'dev'
+    assert scylla_extract_mode("../build/release") == 'release'
+    assert scylla_extract_mode("../build/release/scylla") == 'release'
+    assert scylla_extract_mode("/home/foo/scylla/build/debug") == 'debug'
+    assert scylla_extract_mode("url=../scylla/build/debug/scylla-package.tar.gz") == 'debug'
+
+    assert scylla_extract_mode("url=./scylla-debug-x86_64-package.tar.gz") == 'debug'
+    assert scylla_extract_mode("url=./scylla-x86_64-package.tar.gz") == 'release'
+    assert scylla_extract_mode("url=./scylla-debug-aarch64-package.tar.gz") == 'debug'
+    assert scylla_extract_mode("url=./scylla-package.tar.gz") == 'release'
+    assert scylla_extract_mode("url=./scylla-debug-package.tar.gz") == 'debug'
+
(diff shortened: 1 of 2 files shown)


bhalevy:
There's also this format in http://downloads.scylladb.com/downloads/scylla/relocatable/scylladb-4.5/
`scylla-debug-package-4.5.2.0.20211114.26aca7b9f.tar.gz`

tests/test_common.py
```diff
@@ -0,0 +1,12 @@
+from ccmlib.common import scylla_extract_mode
+
+
+def test_scylla_extract_mode():
+    assert scylla_extract_mode("build/dev") == 'dev'
+    assert scylla_extract_mode("../build/release") == 'release'
+    assert scylla_extract_mode("../build/release/scylla") == 'release'
+    assert scylla_extract_mode("/home/foo/scylla/build/debug") == 'debug'
+    assert scylla_extract_mode("url=../scylla/build/debug/scylla-package.tar.gz") == 'debug'
+    assert scylla_extract_mode("url=./scylla-x86_64-debug-package.tar.gz") == 'debug'
```
bhalevy, line 10: I'm not sure about that...

I think the mode comes before the arch, e.g. http://downloads.scylladb.com.s3.amazonaws.com/unstable/scylla/branch-4.6/relocatable/2021-12-08T10%3A13%3A35Z/scylla-debug-aarch64-package.tar.gz

bhalevy (changes requested):

tests/test_common.py
bhalevy: Cc @yaronkaikov @benipeled

benipeled: Yup, ex. `scylla-debug-aarch64-package.tar.gz` or `scylla-debug-x86_64-unified-package-4.6.rc0.0.20211130.afc18d507.tar.gz`

PS> IMO the arch-name convention is wrong, should be at the end ex. `scylla-debug-package-aarch64.tar.gz` but that's another story.

fruch:
@benipeled @bhalevy , I've handled both cases (formal release with version string in them, and correct order of mode vs. arch)

tests/test_common.py
```diff
@@ -0,0 +1,19 @@
+from ccmlib.common import scylla_extract_mode
+
+
+def test_scylla_extract_mode():
+    assert scylla_extract_mode("build/dev") == 'dev'
+    assert scylla_extract_mode("../build/release") == 'release'
+    assert scylla_extract_mode("../build/release/scylla") == 'release'
+    assert scylla_extract_mode("/home/foo/scylla/build/debug") == 'debug'
+    assert scylla_extract_mode("url=../scylla/build/debug/scylla-package.tar.gz") == 'debug'
```
bhalevy, line 9: how does that work?

bhalevy (approved):

PR #359, opened 2021-12-22
node: wait_for_compactions: add idle timeout

bhalevy:
Currently wait_for_compactions has no timeout
so it can loop forever if compaction gets stuck
and never completes.

To detect that earlier provide a timeout
(300 seconds by default) after which, if
compactionstats show no progress, the function
raises a TimeoutError.

DTest: compaction_additional_test.py::TestCompactionAdditional::test_compaction_removes_ttld_data_by_time_windows

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -735,16 +735,30 @@ class Node(object):
         else:
             return False
 
-    def wait_for_compactions(self):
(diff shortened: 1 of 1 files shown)


fruch:
@Orenef11 was also trying to address that in #357, but also had lots of other changes there.

fruch (approved):
LGTM

PR #362, opened 2021-12-23
scylla_cluster: start with no_wait=False by default

bhalevy:
This effectively reverts scylla_cluster: handle no_wait flag
since we need the default start to be safe and set `no_wait=True`
explicitly from tests that are proved to tolerate it.
    
Scylla start became faster since
scylladb/scylla@78d0cc4ab5ebd47f101ce78b79a30a48ee9f4b9b
and without this change, multi-node dtests tend to fail.
    
DTest: consistent_bootstrap_test.py::TestBootstrapConsistency::test_consistent_reads_after_bootstrap
(with --cassandra-dir)
    
Validating the debug log:
```
21:13:06,524 3398130 ccm                            DEBUG    cluster.py          :694  | test_consistent_reads_after_bootstrap: start_nodes: no_wait=False wait_for_binary_proto=True wait_other_notice=True force_wait_for_cluster_start=True
21:13:06,528 3398130 ccm                            DEBUG    cluster.py          :694  | test_consistent_reads_after_bootstrap: node1: Starting scylla: args=['/home/bhalevy/.dtest/dtest-_3m0rc17/test/node1/bin/scylla', '--options-file', '/home/bhalevy/.dtest/dtest-_3m0rc17/test/node1/conf/scylla.yaml', '--log-to-stdout', '1', '--api-address', '127.0.72.1', '--collectd-hostname', 'localhost.localdomain.node1', '--developer-mode', 'true', '--smp', '1', '--memory', '512M', '--default-log-level', 'info', '--collectd', '0', '--overprovisioned', '--prometheus-address', '127.0.72.1', '--unsafe-bypass-fsync', '1', '--kernel-page-cache', '1', '--max-networking-io-control-blocks', '1000'] wait_other_notice=True wait_for_binary_proto=True
```
    
Refs and complements https://github.com/scylladb/scylla-dtest/pull/2490

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -83,13 +83,14 @@ class ScyllaCluster(Cluster):
         for node, p, _ in started:
             node._update_pid(p)
 
-    def start_nodes(self, nodes=None, no_wait=True, verbose=False, wait_for_binary_proto=None,
+    def start_nodes(self, nodes=None, no_wait=False, verbose=False, wait_for_binary_proto=None,
               wait_other_notice=None, jvm_args=None, profile_options=None,
               quiet_start=False):
         if wait_for_binary_proto is None:
             wait_for_binary_proto = self.force_wait_for_cluster_start
         if wait_other_notice is None:
             wait_other_notice = self.force_wait_for_cluster_start
+        self.debug(f"start_nodes: no_wait={no_wait} wait_for_binary_proto={wait_for_binary_proto} wait_other_notice={wait_other_notice} force_wait_for_cluster_start={self.force_wait_for_cluster_start}")
         self.started=True
 
         p = None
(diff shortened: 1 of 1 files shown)


bhalevy:
Cc @nyh

nyh:
@bhalevy we need to understand why this is needed... Do you need this fix even **after** your fix https://github.com/scylladb/scylla-dtest/pull/2490 was merged?

The point is, what this patch changes is to have ccm wait a bit longer **between** starting each node (it will wait until each node reaches "normal" state, not only "listening" state). But it doesn't change anything on what ccm or dtest waits for after the entire cluster is up. I suspect (but maybe my suspicious is wrong) that we don't need to sleep more between nodes - we just need to sleep more after the entire cluster is up before starting the actual test, and this is what your PR 2490 fixed. The question is whether this patch is needed **as well**.

bhalevy:
@nyh I think that the rule of thumb should be to use safe defaults and to override them only where it safe, not to gamble and try unsafe defaults and see what breaks...

#2490 fixed more than waiting after the entire cluster is up, but rather it fixed the implicit defaults for `wait_for_binary_proto` and `wait_other_notice` when they aren't specified by the caller.

`no_wait=True` should be used when the test starts a node asynchronously without waiting for anything.
Typically the test then calls something like `node.watch_log_for` or some other way to poll the node's state and determine when it has reached a state that the test can proceed.  It's not about performance.

fruch (approved):
LGTM

PR #363, opened 2021-12-26
_get_scylla_version: fallback to `scylla --version` when `SCYLLA-VERS…

fruch:
…ION-FILE` isn't available

Since in some cases we won't have `SCYLLA-VERSION-FILE` in place, we fallback to using
`scylla --version` commandline insted to get the version

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -659,8 +659,8 @@ def get_version_from_build(install_dir=None, node_path=None):
 
 
 def _get_scylla_version(install_dir):
-    scylla_version_files = [ os.path.join(install_dir, 'build', 'SCYLLA-VERSION-FILE'),
-                                os.path.join(install_dir, 'scylla-core-package', 'SCYLLA-VERSION-FILE') ]
+    scylla_version_files = [os.path.join(install_dir, 'build', 'SCYLLA-VERSION-FILE'),
+                            os.path.join(install_dir, 'scylla-core-package', 'SCYLLA-VERSION-FILE')]
     for version_file in scylla_version_files:
         if os.path.exists(version_file):
             with open(version_file) as file:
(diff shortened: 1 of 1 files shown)


ccmlib/common.py
```diff
@@ -670,7 +670,24 @@ def _get_scylla_version(install_dir):
...
+    for scylla_exec in [
+        os.path.join(install_dir, 'scylla'),
+        os.path.join(install_dir, 'build', 'debug', 'scylla'),
+        os.path.join(install_dir, 'build', 'dev', 'scylla'),
+        os.path.join(install_dir, 'build', 'release', 'scylla'),
+        os.path.join(install_dir, 'bin', 'scylla'),
+        os.path.join(install_dir, 'scylla-core-package', 'bin', 'scylla'),
+    ]:
+        try:
+            scylla_version = subprocess.run(f"{scylla_exec} --version", shell=True, capture_output=True, text=True)
```
bhalevy, line 683: That won't work when scylla is built in dbuild environement.
It must be installed first and patchelf'ed to point to the loadable libs dir.
Otherwise you get:
```
/home/bhalevy/dev/scylla-2/build/dev/scylla: error while loading shared libraries: libthrift-0.13.0.so: cannot open shared object file: No such file or directory
```

bhalevy (changes requested):

bhalevy:
I'm not sure that this PR is required.
We get the right scylla version eventually when finding `SCYLLA-VERSION-FILE` under `build/`

nyh:
@fruch I think you had the right idea. The `SCYLLA-VERSION-FILE` is in build/, **not** in build/dev which is where I tell dtest that Scylla lives! So will it, or won't it, find SCYLLA-VERSION-FILE?

I think running `scylla --version` is exactly the right thing to do if we need the version. It's also deliberately very fast - it doesn't start Seastar, allocate memory, and so on.

bhalevy:
> @fruch I think you had the right idea. The `SCYLLA-VERSION-FILE` is in build/, **not** in build/dev which is where I tell dtest that Scylla lives! So will it, or won't it, find SCYLLA-VERSION-FILE?
> 
> I think running `scylla --version` is exactly the right thing to do if we need the version. It's also deliberately very fast - it doesn't start Seastar, allocate memory, and so on.

@nyh we's definitely like to do that, but it will require a much bigger change since to run scylla you have to install the binary and the lodable objects and use `patchelf` to point the loader to where the loadable objects are.  This is done only later in the process but we currently look for the version before scylla is installed from the build directory.

PR #365, opened 2021-12-27
node: flush: add ks and table optional args

bhalevy:
Allow flushing a specified keyspace / table.

DTest: (modified) test_disable_autocompaction_schema


fruch (approved):
LGTM

PR #367, opened 2021-12-28
style(common.py): Moving wait_for from scylla_node

Orenef11:
The method is generic and can be "call" from each file in the package

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -461,6 +461,37 @@ def scylla_extract_install_dir_and_mode(install_dir):
     return install_dir, scylla_mode
 
 
+def wait_for(func, timeout, first=0.0, step=1.0, text=None):
+    """
+    Wait until func() evaluates to True.
+
+    If func() evaluates to True before timeout expires, return the
+    value of func(). Otherwise return None.
+
+    :param func: Function that will be evaluated.
+    :param timeout: Timeout in seconds
+    :param first: Time to sleep before first attempt
+    :param step: Time to sleep between attempts in seconds
+    :param text: Text to print while waiting, for debug purposes
+    """
+    start_time = time.time()
+    end_time = time.time() + timeout
+
+    time.sleep(first)
+
+    while time.time() < end_time:
+        if text:
+            print_("%s (%f secs)" % (text, (time.time() - start_time)))
+
+        output = func()
+        if output:
+            return output
+
+        time.sleep(step)
+
+    return None
+
+
 def check_file_exists(path_to_file):
     if os.path.exists(path_to_file):
         return True
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -548,19 +548,15 @@ def check_socket_available(itf):
 
 
 def check_socket_listening(itf, timeout=60):
-    end = time.time() + timeout
-    while time.time() <= end:
-        try:
-            sock = socket.socket()
-            sock.connect(itf)
-            sock.close()
-            return True
-        except socket.error:
-            # Try again in another 200ms
-            time.sleep(.2)
-            continue
+    def _check_socket_listening() -> bool:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as _socket:
+            try:
+                _socket.connect(itf)
+                return True
+            except socket.error:
+                return False
 
-    return False
+    return wait_for(func=_check_socket_listening, timeout=timeout, step=0.2)
 
 
 def interface_is_ipv6(itf):
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -14,6 +14,7 @@ import sys
 import time
 import tempfile
 import logging
+from typing import Callable
 
 import yaml
 from boto3.session import Session
@@ -461,18 +462,18 @@ def scylla_extract_install_dir_and_mode(install_dir):
     return install_dir, scylla_mode
 
 
-def wait_for(func, timeout, first=0.0, step=1.0, text=None):
+def wait_for(func: Callable, timeout: int, first: float = 0.0, step: float = 1.0, text: str = ""):
     """
     Wait until func() evaluates to True.
 
-    If func() evaluates to True before timeout expires, return the
-    value of func(). Otherwise return None.
+    If func() evaluates to True before timeout expires, return True.
+    Otherwise, return False.
 
-    :param func: Function that will be evaluated.
-    :param timeout: Timeout in seconds
-    :param first: Time to sleep before first attempt
-    :param step: Time to sleep between attempts in seconds
-    :param text: Text to print while waiting, for debug purposes
+    param func: Function that will be evaluated.
+    param timeout: Timeout in seconds.
+    param first: Time to sleep in seconds before first attempt.
+    param step: Time to sleep in seconds between attempts in seconds.
+    param text: Text to print while waiting, for debug purposes.
     """
     start_time = time.time()
     end_time = time.time() + timeout
@@ -482,14 +483,10 @@ def wait_for(func, timeout, first=0.0, step=1.0, text=None):
     while time.time() < end_time:
         if text:
             print_("%s (%f secs)" % (text, (time.time() - start_time)))
-
-        output = func()
-        if output:
-            return output
-
+        if func():
+            return True
         time.sleep(step)
-
-    return None
+    return False
 
 
 def check_file_exists(path_to_file):
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -489,33 +489,10 @@ def wait_for(func: Callable, timeout: int, first: float = 0.0, step: float = 1.0
     return False
 
 
-def check_file_exists(path_to_file):
-    if os.path.exists(path_to_file):
-        return True
-    return False
-
-
-def _wait_until_false(wait_seconds, func, kwargs):
-    start_time = time.time()
-    wait_time_sec = 5
-    while True:
-        if not func(**kwargs):
-            return True
-        elapsed = time.time() - start_time
-        if elapsed >= wait_seconds:
-            return False
-        time.sleep(wait_time_sec)
-        if elapsed + wait_time_sec > wait_seconds:
-            wait_time_sec = wait_seconds - elapsed
-        elif wait_time_sec <= 16:
-            wait_time_sec *= 2
-
-
 def wait_for_parallel_download_finish(placeholder_file):
-    if check_file_exists(placeholder_file):
-        if not _wait_until_false(wait_seconds=3600, func=check_file_exists, kwargs={'path_to_file': placeholder_file}):
-            raise TimeoutError(f"Relocatables download still runs in parallel from another test after 60 min. "
-                               f"Placeholder file exists: {placeholder_file}")
+    if not wait_for(func=lambda: not os.path.exists(placeholder_file), timeout=3600):
+        raise TimeoutError(f"Relocatables download still runs in parallel from another test after 60 min. "
+                           f"Placeholder file exists: {placeholder_file}")
 
 
 def validate_install_dir(install_dir):
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -379,16 +379,16 @@ class ScyllaNode(Node):
                 pass
 
     def _wait_java_up(self, ip_addr, jmx_port):
-        start_time = time.time()
-        while (time.time() - start_time) < 30:
+        def is_java_up():
             with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as _socket:
                 _socket.settimeout(1.0)
                 try:
                     _socket.connect((ip_addr, jmx_port))
                     return True
                 except (socket.timeout, ConnectionRefusedError):
-                    time.sleep(0.05)
-        return False
+                    return False
+
+        return wait_for(func=is_java_up, timeout=30, step=0.05)
 
     def node_install_dir_version(self):
         if not self.node_install_dir:
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -18,7 +18,7 @@ import yaml
 import glob
 import re
 
-from ccmlib.common import CASSANDRA_SH, BIN_DIR
+from ccmlib.common import CASSANDRA_SH, BIN_DIR, wait_for
 from six import print_
 from six.moves import xrange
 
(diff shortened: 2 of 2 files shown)


Orenef11:
@bhalevy 
Each change is in a separate commit.

ccmlib/common.py
```diff
@@ -461,6 +461,37 @@ def scylla_extract_install_dir_and_mode(install_dir):
     return install_dir, scylla_mode
 
 
+def wait_for(func, timeout, first=0.0, step=1.0, text=None):
```
Orenef11, line 464: I only move the method from `ccmlib/scylla_node.py` to here

ccmlib/common.py
```diff
@@ -492,33 +492,10 @@ def wait_for(func, timeout, first=0.0, step=1.0, text=None):
...
-        elif wait_time_sec <= 16:
-            wait_time_sec *= 2
-
-
 def wait_for_parallel_download_finish(placeholder_file):
-    if check_file_exists(placeholder_file):
-        if not _wait_until_false(wait_seconds=3600, func=check_file_exists, kwargs={'path_to_file': placeholder_file}):
-            raise TimeoutError(f"Relocatables download still runs in parallel from another test after 60 min. "
-                               f"Placeholder file exists: {placeholder_file}")
+    if not wait_for(func=lambda: not os.path.exists(placeholder_file), timeout=3600):
```
Orenef11, line 496: I have checked via dtest and it was worked

ccmlib/node.py
```diff
@@ -741,24 +742,26 @@ def wait_for_compactions(self, idle_timeout=300):
...
-        pending_tasks = None
-        last_change = None
+        prev_task_size, next_task_size = -1, -1
         pattern = re.compile(r"pending tasks:\s*(?P<tasks>\d+)")
-        while not last_change or time.time() - last_change < idle_timeout:
-            output, err = self.nodetool("compactionstats", capture_output=True)
-            m = pattern.search(output)
-            if not m:
+
+        def _wait_for_compactions():
```
Orenef11, line 748: I have checked via dtest and the logic is worked well.

ccmlib/scylla_node.py
```diff
@@ -729,19 +729,9 @@ def do_stop(self, gently=True):
...
-                return True
-            elapsed = time.time() - start_time
-            if elapsed >= wait_seconds:
-                return False
-            time.sleep(wait_time_sec)
-            if elapsed + wait_time_sec > wait_seconds:
-                wait_time_sec = wait_seconds - elapsed
-            elif wait_time_sec <= 16:
-                wait_time_sec *= 2
+        if wait_for(func=self.is_running, timeout=wait_seconds):
```
Orenef11, line 732: When the time runs out, the value returned is `None`.

Orenef11:
> @bhalevy Each change is in a separate commit.

ccmlib/node.py
```diff
@@ -741,24 +742,26 @@ def wait_for_compactions(self, idle_timeout=300):
...
-        raise TimeoutError(f"Waiting for compactions timed out after {idle_timeout} seconds with {pending_tasks} pending tasks remaining.")
+            next_task_size = int(match.group('tasks'))
+            if next_task_size == 0:
+                return True
+            if next_task_size > prev_task_size:
+                self.warning(f"Pending compaction tasks increased from {prev_task_size} to "
+                             f"{next_task_size} while waiting for compactions.")
+            prev_task_size = next_task_size
+
+        if not wait_for(func=_wait_for_compactions, timeout=idle_timeout, step=1):
```
bhalevy, line 762: This changed the behavior of `idle_timeout` to be "total timeout".
The idea was that if `next_task_size != 0` after `idle_timeout` but `next_task_size != prev_task_size`, let the loop continue for another `idle_timeout` until either `next_task_size == 0` or there is no progress (`next_task_size == prev_task_size`)

The reason is that we want to support long running compactions (previously there was no timeout at all)

ccmlib/scylla_node.py
```diff
@@ -1204,19 +1204,18 @@ def _write_agent_log4j_properties(self, agent_dir):
...
-            stdout, _ = self.nodetool('cfstats', timeout=wait_timeout)
+        def no_pending_flushes() -> bool:
+            stdout = self.nodetool('cfstats', timeout=wait_timeout)[0]
             pending_flushes = False
             for line in stdout.splitlines():
                 line = line.strip()
                 if line.startswith('Pending flushes'):
-                    _, pending_flushes_str = line.split(':')
+                    pending_flushes_str = line.split(':', maxsplit=1)[1]
                     pending_flushes_count = int(pending_flushes_str.strip())
```
bhalevy, line 1204: I'd just use a regex to extract the field.  It's simpler.  For example:

```Python
m = re.match(r'Pending flushes:\s*(?P<count>\d+)\s*', line)
if m:
    pending_flushes_count = int(m.group('count'))
```

ccmlib/scylla_node.py
```diff
@@ -1204,19 +1204,18 @@ def _write_agent_log4j_properties(self, agent_dir):
...
                     if pending_flushes_count > 0:
                         pending_flushes = True
             return not pending_flushes
-        result = wait_for(no_pending_flushes, timeout=wait_timeout, step=1.0)
-        if result is None:
+        if wait_for(no_pending_flushes, timeout=wait_timeout, step=1.0) is None:
```
bhalevy, line 1208: why not `if not wait_for`?

bhalevy (changes requested):

ccmlib/common.py
```diff
@@ -461,6 +461,37 @@ def scylla_extract_install_dir_and_mode(install_dir):
     return install_dir, scylla_mode
 
 
+def wait_for(func, timeout, first=0.0, step=1.0, text=None):
+    """
+    Wait until func() evaluates to True.
+
+    If func() evaluates to True before timeout expires, return the
+    value of func(). Otherwise return None.
```
bhalevy, line 469: BTW, It's odd that `wait_for` returns `None` rather than `False`, if it's defined to wait until `func` returns `True`.

It would make sense if we defined it as `wait_until_value` where it would wait until `func` return non- `None` value (could be `False`) and returns it, otherwise return `None` if the timeout expires.

I'm not sure there's a use case for the latter though.

ccmlib/common.py
```diff
@@ -461,6 +461,37 @@ def scylla_extract_install_dir_and_mode(install_dir):
...
+
+    :param func: Function that will be evaluated.
+    :param timeout: Timeout in seconds
+    :param first: Time to sleep before first attempt
```
nyh, line 473: Please say "time (in seconds) to sleep", just like you said above "timeout in seconds". Seconds is probably the natural choice, but better be explicit about it.

ccmlib/common.py
nyh: It seems that in some sense, this function does return the value - the text is ambigous (and even misleading) on what "func() evaluates to True" means. It sounds like you check whether `func() == True`, but that's not actually what you do - you check `if(func())` which means that func() returned an object whose "truth value" is true (see https://docs.python.org/3/library/stdtypes.html#truth). So for example, if func() returns a string, a non-empty string is considered true, and then wait_for will return this string.

What I would have done in this case is one of two things:
1. Simplify the API. `wait_for()` will just return True or False - nothing else (not a string, not None, etc.). Does any user need more than that in practice? The API will be simpler and require less explanations in comments.
2. Keep the current return-the-value API, but fix the return value of the timeout case: In the timeout case, don't just return "None"! instead return the last "output, i.e., the last result of `func()`. This may be anything not-true - False (usually...), None, empty string or whatever. But the idea is that if the user's func() returns actual True or False, so will wait_for() instead that weird None return.

ccmlib/scylla_node.py
```diff
@@ -1204,19 +1204,18 @@ def _write_agent_log4j_properties(self, agent_dir):
...
-        def no_pending_flushes():
-            stdout, _ = self.nodetool('cfstats', timeout=wait_timeout)
+        def no_pending_flushes() -> bool:
+            stdout = self.nodetool('cfstats', timeout=wait_timeout)[0]
             pending_flushes = False
             for line in stdout.splitlines():
                 line = line.strip()
                 if line.startswith('Pending flushes'):
-                    _, pending_flushes_str = line.split(':')
+                    pending_flushes_str = line.split(':', maxsplit=1)[1]
```
nyh, line 1203: Doesn't this maxsplit=1 change the behavior of this code?
I guess it will never matter because there is just one ":" anyway, but I'm not familiar enough with this code to say.

ccmlib/scylla_node.py
nyh: Right! If you fix wait_for using one of my two recommended fixes in the previous patches, it won't be None here. But even if you don't fix it, "if not wait_for" will also work because "None" is considered false. And "if not" is clearer.

ccmlib/common.py
nyh, line 496: You could have left the check_file_exists function instead of making this a lambda :-) But I share your sentiment (in this case) of reducing the number of "convenience functions" that don't really add anything.

ccmlib/common.py
```diff
@@ -551,19 +551,15 @@ def check_socket_available(itf):
...
-            sock = socket.socket()
-            sock.connect(itf)
-            sock.close()
-            return True
-        except socket.error:
-            # Try again in another 200ms
-            time.sleep(.2)
-            continue
+    def _check_socket_listening() -> bool:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as _socket:
```
nyh, line 555: Nitpick: why add the AF_INET, SOCK_STREAM parameters, if it works without them?
What if in 10 years AF_INET6 becomes the default, and this code no longer works?

ccmlib/scylla_node.py
```diff
@@ -379,16 +379,16 @@ def stop_scylla_manager_agent(self, gently):
                 pass
 
     def _wait_java_up(self, ip_addr, jmx_port):
-        start_time = time.time()
-        while (time.time() - start_time) < 30:
+        def wait_java_up_logic():
```
nyh, line 382: I'd give this function a more descriptive name, e.g., is_java_up().

ccmlib/scylla_node.py
```diff
@@ -379,16 +379,16 @@ def stop_scylla_manager_agent(self, gently):
...
                 _socket.settimeout(1.0)
                 try:
                     _socket.connect((ip_addr, jmx_port))
                     return True
                 except (socket.timeout, ConnectionRefusedError):
-                    time.sleep(0.05)
-        return False
+                    return False
+
+        return wait_for(func=wait_java_up_logic, timeout=30)
```
nyh, line 391: Since if I remember correctly wait_for's default step is a whole second, please specify step=0.05 or you'll return the annoying oversleep issue we had here previously.

ccmlib/scylla_node.py
```diff
@@ -729,19 +729,9 @@ def do_stop(self, gently=True):
...
-            if elapsed >= wait_seconds:
-                return False
-            time.sleep(wait_time_sec)
-            if elapsed + wait_time_sec > wait_seconds:
-                wait_time_sec = wait_seconds - elapsed
-            elif wait_time_sec <= 16:
-                wait_time_sec *= 2
+        if wait_for(func=self.is_running, timeout=wait_seconds):
+            return True
+        return False
```
nyh, line 734: If you fix wait_for() like I suggested above, you wouldn't need this function at all :-(
If you don't, instead of an if I think you can simply do
```python
return bool(wait_for(...))
```

nyh (changes requested):

ccmlib/common.py
Orenef11: Done

ccmlib/common.py
Orenef11: Done

ccmlib/common.py
Orenef11: There are "helper" methods that use only once time...
thus, I remove those methods :-)

ccmlib/common.py
Orenef11: Yes, you're right, but it's a rate case.
The old logic was wrong because `socket.socket()` fails, the socket not close...

ccmlib/node.py
Orenef11: I revert my change

ccmlib/scylla_node.py
Orenef11: Done

ccmlib/scylla_node.py
```diff
@@ -1204,17 +1204,17 @@ def _write_agent_log4j_properties(self, agent_dir):
         raise NotImplementedError('ScyllaNode._write_agent_log4j_properties')
 
     def _wait_no_pending_flushes(self, wait_timeout=60):
-        def no_pending_flushes():
-            stdout, _ = self.nodetool('cfstats', timeout=wait_timeout)
+        pending_flushes_re = re.compile(r"^\s+?Pending flushes:\s?(?P<count>\d+)\s?$")
+
+        def no_pending_flushes() -> bool:
+            stdout = self.nodetool('cfstats', timeout=wait_timeout)[0]
```
Orenef11, line 1198: @nyh @bhalevy 
I checked it and it working well
image

bhalevy: nit: `\s+?` is an over-complicated equivalent of `\s*`.
I think that this regexp makes more sense: `r"^\s*Pending flushes:\s*(?P<count>\d+)\s*$"`

bhalevy: otherwise looks OK

Orenef11: You're right, I'm used to doing this but it's wrong in this case.
Fixed.

bhalevy (approved):

fruch:
took it for a ride with `next_gating` and it passed:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/114/console

ccmlib/common.py
nyh: I don't see it done, maybe you forgot to push?

ccmlib/common.py
nyh: I don't see it done, maybe you forgot to push?

ccmlib/common.py
nyh: Oh,  I see, the changes are in the next patch... Please combine the two.

ccmlib/common.py
nyh: socket.socket() will not fail, it's the connect() that can fail. And this isn't related to my quesition why pass these parameters to socket() if the defaults were good enough. But never mind, it's not important enough to fuss about.

nyh (approved):

ccmlib/common.py
Orenef11: I mean to the `connect` method that can fail and because of that the socket can be open

Orenef11:
> took it for a ride with `next_gating` and it passed: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/114/console

FYI @roydahan

ccmlib/common.py
nyh: Ok, but this is not related to my comment which was about the parameters to socket(), not to connect(). But never mind, it's fine with or without these parameters. I just thought it's better not to change it without reason.

Orenef11:
> > took it for a ride with `next_gating` and it passed: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/114/console
> 
> FYI @roydahan

@roydahan  @fruch 
3 runs are passed.

fruch (approved):
LGTM

PR #368, opened 2021-12-28
fix(ccmlib/cluster.py): Removing mutable defaults argument

Orenef11:
In Python, when passing a mutable value as a default argument in a function,
the default argument is mutated anytime that value is mutated.
Here, "mutable value" refers to anything such as a list, a dictionnary
or even a class instance.

Example:
```
def append(element, seq=[]):
    seq.append(element)
    return seq

>>> append(1)  # seq is assigned to []
[1]  # This returns a reference the *same* list as the default for `seq`
>>> append(2)  # => `seq` is now given [1] as a default!
[1, 2]  # WTFs and headaches start here…
```

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -521,7 +521,8 @@ class Cluster(object):
         self.nodelist()[0].stress(stress_options=stress_options + ['-node', ','.join(livenodes)] )
         return self
 
-    def run_cli(self, cmds=None, show_output=False, cli_options=[]):
+    def run_cli(self, cmds=None, show_output=False, cli_options=None):
+        cli_options = cli_options or []
         livenodes = [node for node in list(self.nodes.values()) if node.is_live()]
         if len(livenodes) == 0:
             raise common.ArgumentError("No live node")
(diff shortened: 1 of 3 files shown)


bhalevy:
There is no description to this PR.
Please provide the motivation

nyh (approved):
This patch is not strictly necessary, because although the array is mutable, the code doesn't actually mutate it, so there is no risk. However, I agree that there is future risk - if someone modifies this code and adds some code which does modify the array, it will cause a mess, so I agree this patch makes sense.

Orenef11:
> 

I have not carefully followed all the functions that use these values, but not all developers are aware of this problem so I chose to fix it.
Also, the fix was very easy :-)

bhalevy:
Python programmers need to know that Python passes args by reference, e.g.
```Python
>>> def append(l, val):
...  l.append(val)
... 
>>> my_list=['a']
>>> append(my_list, 'b')
>>> my_list
['a', 'b']
```

and the default value is not different.

I'm OK with this change, though, since it clarifies the functions' interface.

bhalevy (approved):

Orenef11:
@roydahan 
I separated all my changes into separate PRs because there are a lot of changes.
If there are no problems with the changes, I will create one branch and run the dtest tests via this branch.

fruch (approved):
LGTM

PR #371, opened 2022-01-18
fix(ccmlib/repository.py): Fixing SyntaxWarning warning log

Orenef11:
The following code incorrect:
```python
while attempt < 3 and ret_val != 0:
...
```
From Python version 3.8+, the following warning message is shown:
```
ccmlib/repository.py:311: SyntaxWarning: "is not" with a literal. Did you mean "!="?
```

### DIFF
--- a/ccmlib/repository.py
+++ b/ccmlib/repository.py
@@ -303,7 +303,7 @@ def compile_version(version, target_dir, verbose=False):
             # Similar patch seen with buildbot
             attempt = 0
             ret_val = 1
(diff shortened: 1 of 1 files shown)


Orenef11:
More details you can find in the following link:
[Why does Python log a SyntaxWarning for ‘is’ with literals?](https://adamj.eu/tech/2020/01/21/why-does-python-3-8-syntaxwarning-for-is-literal/)

fruch (approved):
LGTM

PR #374, opened 2022-01-25
Wait for gossip on start/start_nodes

fruch:
As part of looking at https://github.com/scylladb/scylla-dtest/pull/2455

this tries to be a way we can make sure the gossip round happens between the node.

on the other hand it might serve as a glorified sleep, and "magic" solves issue we had across test that tries to write with CL.ALL straight after nodes are up and failing at it,

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -7,6 +7,7 @@ import signal
 import yaml
 import uuid
 import datetime
+import logging
 
 from six import print_
 from distutils.version import LooseVersion
@@ -16,9 +17,11 @@ from ccmlib.cluster import Cluster
 from ccmlib.scylla_node import ScyllaNode
 from ccmlib.node import NodeError
 from ccmlib import scylla_repository
+from ccmlib.common import wait_for
 
 SNITCH = 'org.apache.cassandra.locator.GossipingPropertyFileSnitch'
 
+logger = logging.getLogger(__name__)
 
 class ScyllaCluster(Cluster):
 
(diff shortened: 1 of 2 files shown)


Orenef11:
I think it's better to create a method in Node class and use it:
[patch.zip](https://github.com/scylladb/scylla-ccm/files/7936330/patch.zip) (GitHub blocks files with a patch extension but approves files with a zip extension.

ccmlib/scylla_cluster.py
```diff
@@ -141,7 +144,12 @@ def start_nodes(self, nodes=None, no_wait=False, verbose=False, wait_for_binary_
                 for node, _, _ in started:
                     if old_node is not node:
                         old_node.watch_log_for_alive(node, from_mark=mark)
-
+                        def check_gossip():
```
Orenef11, line 147: It will better to create a separate method in Node class:
```
    def wait_for_gossip(self, timeout=30):
        if not wait_for(lambda: self.address() in self.nodetool('gossipinfo')[0], step=0.001, timeout=timeout,
                        text=f"wait to gossip for {self.address()}"):
            raise RuntimeError(f"{self.name} wasn't found in {self.name} gossip")
```

Orenef11 (changes requested):

ccmlib/scylla_cluster.py
fruch: BTW this wrong logic, we need to verify against other node, not the node vs. itself.

a node gossip would always show itself...

fruch:
this isn't needed anymore

PR #375, opened 2022-01-26
fix(ccmlib/scylla_node.py): Killing Scylla process in failure

Orenef11:
When creating a new cluster that contains several nodes, one
 of the nodes can fail.
The test failed, but the processes are still alive and
 listening to the ports.
Therefore, they must be killed so that the next test will
 create a new cluster without any problems.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -580,10 +580,20 @@ class ScyllaNode(Node):
         message = "Starting scylla: args={} wait_other_notice={} wait_for_binary_proto={}".format(args, wait_other_notice, wait_for_binary_proto)
         self.debug(message)
 
-        scylla_process = self._start_scylla(args, marks, update_pid,
-                                            wait_other_notice,
-                                            wait_for_binary_proto,
-                                            ext_env)
+        try:
+            scylla_process = self._start_scylla(args, marks, update_pid,
+                                                wait_other_notice,
+                                                wait_for_binary_proto,
+                                                ext_env)
+        except Exception:
+            if os.environ.get("KILL_SCYLLA_PROCESS_AT_FAILURE", False):
+                for node in self.cluster.nodelist():
+                    process_scylla = node._process_scylla
+                    if process_scylla.poll() is None:
+                        self.error(f"Killing scylla process for node '{node.name}'")
+                        os.kill(process_scylla.pid, signal.SIGKILL)
+            raise
+
         self._start_jmx(data)
 
         ip_addr, _ = self.network_interfaces['thrift']


fruch:
That the test to decide what to do in this case, and in dtest we have cleanup code that would do exactly this.

Orenef11:
> That the test to decide what to do in this case, and in dtest we have cleanup code that would do exactly this.

There are other repos that use ccm and failed

> That the test to decide what to do in this case, and in dtest we have cleanup code that would do exactly this.

We have some multiple repos, and some were taken from Cassandra.
Therefore, to not make any changes in the structure of the tests I thought to add this option.
In the case of python-driver, it will be hard to add this change to tests without changing lots of places in the code.

I can create a env variable for this change.

roydahan:
@Orenef11 is right, there are other repos that use ccm like driver tests.
However, I think that a flag / env variable would be the smoothest option.

Orenef11:
> @Orenef11 is right, there are other repos that use ccm like driver tests. However, I think that a flag / env variable would be the smoothest option.

Done

ccmlib/scylla_node.py
```diff
@@ -580,10 +580,20 @@ def start(self, join_ring=True, no_wait=False, verbose=False,
...
-                                            wait_other_notice,
-                                            wait_for_binary_proto,
-                                            ext_env)
+        try:
+            scylla_process = self._start_scylla(args, marks, update_pid,
+                                                wait_other_notice,
+                                                wait_for_binary_proto,
+                                                ext_env)
+        except Exception:
+            if os.environ.get("KILL_SCYLLA_PROCESS_AT_FAILURE", False):
```
Orenef11, line 589: Is the env name clear?

fruch:
@Orenef11 I'm sorry,  I think don't think this is logic CCM should be doing on it's own.
I still didn't see any proof this is an actual issue anywhere.

Orenef11:
> @Orenef11 I'm sorry, I think don't think this is logic CCM should be doing on it's own. I still didn't see any proof this is an actual issue anywhere.

It happens a lot for python-driver-matrix and I'll try to save the python-driver-matrix logs.
For now, I changing the PR's state to draft.

nyh:
My two cents:

If a CCM operation - like creating a cluster - reports a failure, it shouldn't leave behind half-done work. I don't see how any user (a command-line-user, dtest, or anything else) might ask for a 5-node cluster and be content with a 4-node cluster instead.

I think this is orthogonal to the fact that the caller (e.g., dtest) needs to ask ccm to remove the cluster when the caller ends (successfully, unsucessfully, or due to a signal). @fruch is right that if the caller **anyway** asks ccm to remove the cluster even if the creation fails, then everything will already work as it's written now - but it's weird that a caller needs to remove the cluster even if it wasn't successfully created, and I wouldn't be surprised if some callers (?) forget to do it which is I guess what Oren saw (I don't know in which caller).

Orenef11:
> My two cents:
> 
> If a CCM operation - like creating a cluster - reports a failure, it shouldn't leave behind half-done work. I don't see how any user (a command-line-user, dtest, or anything else) might ask for a 5-node cluster and be content with a 4-node cluster instead.
> 
> I think this is orthogonal to the fact that the caller (e.g., dtest) needs to ask ccm to remove the cluster when the caller ends (successfully, unsucessfully, or due to a signal). @fruch is right that if the caller **anyway** asks ccm to remove the cluster even if the creation fails, then everything will already work as it's written now - but it's weird that a caller needs to remove the cluster even if it wasn't successfully created, and I wouldn't be surprised if some callers (?) forget to do it which is I guess what Oren saw (I don't know in which caller).

I saw it in python-driver-matrix tests. The test fails because CCM failed to create the node. Therefore, I believe there are a lot of live nodes left, and as a result, there are over 200 failures.

fruch:
> Can this PR be closed?

yes

PR #378, opened 2022-03-10
fix(rollback): restore system tables

juliayakovlev:
Restore tables of system and system_schema keyspaces during rollback from snapshot that has been
taken before upgrade.

Issue: https://github.com/scylladb/scylla-enterprise/issues/1950

Restore system tables documentation:
https://docs.scylladb.com/upgrade/upgrade-enterprise/upgrade-guide-from-2018.1-to-2019.1/upgrade-guide-from-2018.1-to-2019.1-rpm/#restore-system-tables

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -11,6 +11,8 @@ import stat
 import subprocess
 import time
 import threading
+from pathlib import Path
+
 from pkg_resources import parse_version
 
 import psutil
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM

ccmlib/scylla_node.py
```diff
@@ -1272,20 +1277,74 @@ def _import_executables(self, install_dir):
             self.node.node_install_dir = self.orig_install_dir
             raise NodeUpgradeError("Failed to import executables files. %s" % exc)
 
-    def upgrade(self, upgrade_version: str):
+    def _recover_system_tables(self):
+        """
+        Part of the rollback procedure is to restore system tables.
+        For this goal a snapshot will be taken before downgrade and restore the system tables after downgrade
```
bhalevy, line 1283: "        For this goal a snapshot will be taken before downgrade and restore the system tables after downgrade"

needs to be: `For this goal a snapshot will be taken before upgrade and restore the system tables before rollback`, no?

PR #382, opened 2022-04-12
fix(get_seeds): return correct seed nodes after set_configuration_options changed.

aleksbykov:
ccmlib **cluster.populate** method create nodes with
seeds, where each node have seeds all previous seeds
and itself
ex:
node1: node1
node2: node1, node2
node3: node1, node2, node3

the **cluster.set_configuration_options()** method rewrite the
scylla.yaml per node.

If it is called after **cluster.populate** method, but before
**cluster.start** method, **cluster.set_configuration_options()**
rewrite the node's scylla.yaml with seeds configuration all seeds:
node1: node1, node2, node3
node2: node1, node2, node3
node3: node1, node2, node3

for some feature, such seed configuration could affect cluster start
and cluster will not be started.

New method parameter is presented, which allow to return
all seeds node before provided node.

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -397,8 +397,15 @@ class Cluster(object):
     def get_path(self):
         return os.path.join(self.path, self.name)
 
-    def get_seeds(self):
-        return [s.network_interfaces['storage'][0] for s in self.seeds]
+    def get_seeds(self, node=None):
+        # if first node, or there is now seeds at all
+        # or node added is not a seed, return 
+        # all cluster seed nodes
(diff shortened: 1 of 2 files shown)


aleksbykov:
Related to issues:
https://github.com/scylladb/scylla/issues/10355
https://github.com/scylladb/scylla/issues/10163

ManManson (approved):
Fixes the startup hangs for https://github.com/scylladb/scylla/issues/10287.

kostja:
@fruch could you please take a look? We need this to make sure dtest passes with raft enabled.

fruch (approved):
LGTM

PR #388, opened 2022-06-14
Fix enterprise version extraction for path/url

fruch:
In enterprise release versions and release canidates
the the return value was `enterprise` and not one of
`release|debug|dev`

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -443,6 +443,7 @@ def isOpscenter(install_dir):
     opscenter_script = os.path.join(bin_dir, 'opscenter')
     return os.path.exists(opscenter_script)
 
+
 def scylla_extract_mode(path):
     # path/url examples:
     #   build/dev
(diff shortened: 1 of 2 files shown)


ccmlib/common.py
```diff
@@ -458,13 +459,14 @@ def scylla_extract_mode(path):
     # path/url examples:
     #   /jenkins/data/relocatable/unstable/master/202001192256/scylla-package.tar.gz
     #   url=https://downloads.scylla.com/relocatable/unstable/master/202001192256/scylla-debug-package.tar.gz
-    m = re.search(r'(^|/)scylla(-(\w+))?(-(\w+))?-package([-./][\w./]+|$)', path)
+    m = re.search(r'(^|/)(scylla|scylla-enterprise)(-(\w+))?(-(\w+))?-package([-./][\w./]+|$)', path)
     if m:
         mode = m.groups()
-        return mode[2] if mode[4] or mode[2] in ('debug', 'dev') else 'release'
+        return mode[3] if mode[3] in ('debug', 'dev') else 'release'
```
bhalevy, line 465: We can use named groups in the regex expression to point at the extracted mode rather than using the group index.

bhalevy (changes requested):

ccmlib/common.py
fruch: change to named group,
but it's still need the exact same logic, since the mode is optional in the path/url

bhalevy (approved):

ccmlib/common.py
bhalevy: sure, but still it doesn't rely on the position of the group in the expression which is more robust.

PR #389, opened 2022-06-21
_grep_log_for_errors: match only if ERROR, INFO, DEBUG is a whole word

piodul:
Modifies the _grep_for_log_errors function so that it matches on ERROR,
INFO and DEBUG strings only if they are a whole word, e.g. a line which
contains "ERRORS" won't be considered by the function.

My motivation behind this change is that I would like to introduce a
feature called "TYPED_ERRORS_IN_READ_RPC" in Scylla. Feature names are
printed to logs with INFO level on startup, and the current logic in
dtests (which uses the grep_log_for_errors function) considers this to
be a failure. This commit tightens the check so that the feature can be
added without problems, and lines printed with error log level (I
suppose they are of most interest here) are still matched by the logic.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1993,12 +1993,20 @@ def _get_row_cache_entries_from_info_output(info):
 
 
 def _grep_log_for_errors(log, distinct_errors=False, search_str=None, case_sensitive=True):
+    def make_pat_for_log_level(level):
+        flags = 0 if case_sensitive else re.IGNORECASE
+        return re.compile(rf'\b{level}\b', flags=flags)
+
+    error_pat = make_pat_for_log_level("ERROR")
+    info_pat = make_pat_for_log_level("INFO")
+    debug_pat = make_pat_for_log_level("DEBUG")
+
     matchings = []
     it = iter(log.splitlines())
     for line in it:
         l = line if case_sensitive else line.lower()
-        is_error_line = ('ERROR' in l and
-                         'DEBUG' not in l.split('ERROR')[0]) if not search_str else search_str in l
+        is_error_line = (error_pat.search(l) and
+                         not debug_pat.search(error_pat.split(l)[0])) if not search_str else search_str in l
         if is_error_line:
             append_line = line if not search_str else l[l.rfind(search_str):]
             if not distinct_errors:
@@ -2007,7 +2015,7 @@ def _grep_log_for_errors(log, distinct_errors=False, search_str=None, case_sensi
             if not distinct_errors:
                 try:
                     it, peeker = itertools.tee(it)
-                    while 'INFO' not in next(peeker):
+                    while not info_pat.search(next(peeker)):
                         matchings[-1].append(next(it))
                 except StopIteration:
                     break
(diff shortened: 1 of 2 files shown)


old_tests/test_lib.py
```diff
@@ -270,3 +270,7 @@ def test_does_not_coalesce_info_lines(self):
         err = ('ERROR: You have made a terrible mistake\n'
                '  2015-05-12 14:12:12,720 INFO: why would you ever do that\n')
         self.assertGreppedLog(err, [['ERROR: You have made a terrible mistake']])
+
+    def test_ignore_error_as_a_part_of_word(self):
```
fruch, line 274: FYI, those test we are not running in CCM CI

but if it was enough for your needs...

fruch (approved):
LGTM

bhalevy (changes requested):

ccmlib/node.py
```diff
@@ -1993,12 +1993,19 @@ def _get_row_cache_entries_from_info_output(info):
 
 
 def _grep_log_for_errors(log, distinct_errors=False, search_str=None, case_sensitive=True):
+    def make_pat_for_log_level(level):
+        return re.compile(r'\b{}\b'.format(level if case_sensitive else level.lower()))
```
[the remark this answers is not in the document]
bhalevy: Then you won't need to call `level.lower()`, `rf'\b{level}\b'` could be used in both case sensitivity cases.

old_tests/test_lib.py
piodul: It looked like the most logical place for this test. It's a pity that it's not being run in CI, should I move it somewhere else? 
All tests from `TestErrorLogGrepping` seem to pass.

psarna (approved):

old_tests/test_lib.py
fruch: never mind, it's a refactoring that doesn't belong to this PR...

ccmlib/node.py
```diff
@@ -1993,12 +1993,20 @@ def _get_row_cache_entries_from_info_output(info):
 
 
 def _grep_log_for_errors(log, distinct_errors=False, search_str=None, case_sensitive=True):
+    def make_pat_for_log_level(level):
+        flags = 0 if case_sensitive else re.IGNORECASE
```
bhalevy, line 1997: This will work but I don't like open coding the default value for flags.

It would be more future-compatible to write it like:
```python
    expr = rf'\b{level}\b'
    return re.compile(expr) if case_sensitive else re.compile(expr, flags=re.IGNORECASE)
```

bhalevy (changes requested):

ccmlib/node.py
bhalevy: oops, I see it's already merged.
can be done as a follow up

fruch: This was is the formal API default value, don't thing it's going to change anytime soon

https://docs.python.org/3/library/re.html#re.compile

Also this might be a bit clearer, if you think this safety is needed

```python
kwargs = {'flags': re.I} if ignore_case else {}
return re.compile(regex, **kwargs)
```

bhalevy: > This was is the formal API default value, don't thing it's going to change anytime soon

Other than python changing its APIs quite radically between major versions, we should let it decide about the default value, not hard code it ourselves, even if it's documented (for this version).  E.g. it might change to None in a future version.

The kwargs approach is also viable...

piodul: Sent https://github.com/scylladb/scylla-ccm/pull/392 as a followup

PR #392, opened 2022-06-23
_grep_log_for_errors: avoid explicit `flags` argument in `re.compile`

piodul:
This is a followup to #389. Now, when grepping the logs in
a case-sensitive manner, instead of passing `flags=0` to `re.compile`,
the argument is omitted entirely which makes sure that the function
receives the default value for the `flags` argument. This future proofs
the logic a bit so that, in case the default value for `flags` changes
in the future, it will still work correctly.


bhalevy (approved):
LGTM, thanks

fruch (approved):
LGTM

PR #393, opened 2022-06-27
scylla_node: start: use storage network address for jmx host

bhalevy:
Don't use the thrift interface since it may be changed to localhost
by tests like `pushed_notifications_test.py::TestPushedNotifications::test_restart_node_localhost`:
```
ccmlib.node.NodeError: Error starting node node1: unable to connect to scylla-jmx port localhost:7199
```


fruch (approved):
LGTM

PR #394, opened 2022-06-27
node: pass down kwargs to node.flush() call

fruch:
so we can pass down parameters to the `nodetool()`
mostly timeout

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1230,8 +1230,8 @@ class ScyllaNode(Node):
             raise NodeError("Node %s still has pending flushes after "
                             "%s seconds" % (self.name, wait_timeout))
 
-    def flush(self, ks=None, table=None):
-        super(ScyllaNode, self).flush(ks, table)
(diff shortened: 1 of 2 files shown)


ccmlib/scylla_node.py
```diff
@@ -1230,8 +1230,8 @@ def no_pending_flushes() -> bool:
             raise NodeError("Node %s still has pending flushes after "
                             "%s seconds" % (self.name, wait_timeout))
 
-    def flush(self, ks=None, table=None):
-        super(ScyllaNode, self).flush(ks, table)
+    def flush(self, ks=None, table=None, **kwargs):
+        super(ScyllaNode, self).flush(ks, table, **kwargs)
```
bhalevy, line 1234: the timeout parameter can be passed down to `_wait_no_pending_flushes` too,
but can be done as a follow up.

bhalevy (approved):

ccmlib/scylla_node.py
fruch: Let's added when some code actually needs it

PR #395, opened 2022-07-14
fix error log detection

gusev-p:
false positives on the lines like:
INFO  2022-07-14 07:21:30,117 [shard 0] gossip - Feature check passed. Local node 127.0.3.1 features = {CDC, ..., TYPED_ERRORS_IN_READ_RPC}


gusev-p:
already fixed in the upstream

PR #398, opened 2022-08-17
Generate key in PKCS#8 format instead of PKCS#1

Bouncheck:
Due to language constraints java driver desperately needs the key to be in `PKCS#8` format.
I'm not fully aware if this change would break any other tool.


fruch:
Only too tests in the python-driver are using this code, I'll check them tomorrow, and if it works I'll merge.

If not, we'll see how we parameterize it, or create multiple keys.

fruch:
@Bouncheck tested, and it's works o.k. for the scylla-driver test.

fruch (approved):
LGTM

PR #402, opened 2022-09-05
ccmlib - node.py - ignore additional messages from stress output

fee-mendes:
This commit ignore a new string ("too many windows") part of cassandra-stress output prior to simply returning an error back to the caller. The decision to ignore that specific string is because it is tied to a specific pull request, rather than simply ignoring a more generic message such as "Warning" which may mistakenly ignore messages which shouldn't.

The first time we saw several tests relying on C-S outputs failing was in https://github.com/scylladb/scylladb/commit/4d7f55a29f20d2ca413aa04eb79666738d8e2467.
However - in https://github.com/scylladb/scylla-dtest/pull/2331 the "fix" for it was simply deprecate and override all mentions of DateTieredCompactionStrategy to TimeWindowCompactionStrategy.

As of https://github.com/scylladb/scylladb/pull/11445 we are now introducing a TimeWindowCompactionStrategy guard rails and - since cassandra-stress doesn't
provide a simple way to create a table with a specific default_time_to_live value (other than manually going through an user provided profile or hacking it around by creating the table early in the process) - this broke some tests which didn't expect the new introduced warning.

An example of such a failure may be found under https://jenkins.scylladb.com/job/releng/job/Scylla-CI/2066/testReport/junit/resharding_test/TestReshardingSingleNodeGating/Sanity_Tests___test_resharding_by_murmur3_gating_1_TimeWindowCompactionStrategy_15_/ .

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1281,7 +1281,10 @@ class Node(object):
 
     def stress_object(self, stress_options=[], ignore_errors = False, **kwargs):
         out, err = self.stress(stress_options, True, **kwargs)
-        if not ignore_errors and err != "" and not err.startswith("Failed to connect over JMX; not collecting these stats") and not err.startswith("Picked up JAVA_TOOL_OPTIONS"):
+        if not ignore_errors and err != "" \
+           and not err.startswith("Failed to connect over JMX; not collecting these stats") \
+           and not err.startswith("Picked up JAVA_TOOL_OPTIONS") \
+           and "too many windows" not in err:
             return err
         p = re.compile(r'^\s*([^:]+)\s*:\s*(\S.*)\s*$')
         res = {}


fruch (commented):
Maybe we change this and delegate the selection of expected warning/errors to the test in dtest ?

fee-mendes:
@fruch yes, certainly that would be the ideal -- as then people would have the ability to make their own changes according to their specific scenarios. As guard-rails more and more come in the future -- this essentially means that any other "unexpected" messages thrown at C-S have a probability for failure and headache. :-)

fee-mendes:
Perhaps rather than relying on a per-case string matching (which is ugly indeed I must admit) we can simply ignore anything containing `generated server side warning(s):` ? Not that it makes it more beautiful than the former, but at least it will effectively avoid referencing server-side warnings as errors.

Ideally - one would prefer to move the logic to each actual dtest - but this will already require ccm changes anyway, and there are at least 5 dtests which rely on the very same function to work today. Refactoring all dtests that ever exist seems like too much just to get unblocked.

fruch:
> I want to comment that what this patch (or the code it modifies) calls an "error" isn't actually an error - as you can see in https://jenkins.scylladb.com/job/releng/job/Scylla-CI/2066/testReport/junit/resharding_test/TestReshardingSingleNodeGating/Sanity_Tests___test_resharding_by_murmur3_gating_1_TimeWindowCompactionStrategy_15_/, 
> 
> the "error" is:
> ```
> E           Exception: Error running cassandra-stress: WARN  06:47:25,481 Query '[0 bound values] CREATE TABLE IF NOT EXISTS standard1 (key blob PRIMARY KEY 
> E           , "C0" blob
> E           , "C1" blob
> E           , "C2" blob
> E           , "C3" blob
> E           , "C4" blob) WITH compression = {} AND compaction = { 'class' : 'TimeWindowCompactionStrategy'};' 
> generated server side warning(s): TimeWindowCompactionStrategy tables without a default_time_to_live may potentially introduce too many windows. Ensure that insert statements specify a TTL (via USING TTL), when inserting data to this table
> ```
> 
> This is not really an error, it's a server-side **warning**, that our code which runs cassandra-stress for some reason decided to elevate to the status of an error. But it isn't an error, and our code should probably realize it and not convert it into an exception, without needing to resort to string matching. Or perhaps we decided to deliberately catch all warnings and consider them as errors?
> 
> In my opinion, if a CREATE TABLE succeeded, then any warning of any kind (a message saying "generated server side warning") should be fine. If the error was serious, the table creation shouldn't have succeeded.

The core issue here is the cassandra-stress is printing those warning to the stderr, and the CCM code basic assumption the that anything written into stderr means there was an error.

It should be checking return codes if anything, and passing back stdout, stderr and the return codes to the caller test that should decide if it's failure or not.

But changing the node.stress() api, need to be done with care.

nyh:
I have a silly question (which is probably outside the scope of this pull request): Why is code for running **cassandra-stress** even part of CCM? Was it always (in Cassandra's original implementation) part of CCM or did we add it? Because it sounds to me that running cassandra-stress should be part of dtest, not of CCM (the "Cassandra Cluster Manager"). Running a load is not part of managing a cluster...

fruch:
> I have a silly question (which is probably outside the scope of this pull request): Why is code for running **cassandra-stress** even part of CCM? Was it always (in Cassandra's original implementation) part of CCM or did we add it? Because it sounds to me that running cassandra-stress should be part of dtest, not of CCM (the "Cassandra Cluster Manager"). Running a load is not part of managing a cluster...

yes it's is part of cassandra CCM, and a whole bunch of other scripts as well (keep in mind all of them are originating from the same cassandra repo, so it might be logical for them back then), we followed that path in few places, for example our CCM also know how to install/run scylla-manager (and now sni_proxy), so yes in some cases ccm is doing a bit more the title would suggest.

one more thing to remember cassandra-stress is part of the scylla-java-tools, which CCM uses and installs.

fruch:
the upstream is using code was changed and passing back now a namedtuple with rc,stdout,stderr: https://github.com/riptano/ccm/blob/4609ac46beccd3ab86463a345ed59e998d664d04/ccmlib/node.py#L2297

and the `stress_object()` call, is home made here in scylla (used in 14 places)
`node.stress()` is used in 106+ places in dtest, but non of them even hold the return value. 

so chaining it should affect dtest.

PR #412, opened 2022-11-08
utils/sni_proxy.py: align implementation to delivered cloud bundle

fruch:
* align the DNS names and node_domain usage
* pass the ip address in `server` (all drivers should be fix/align to support using ip in the `server` for testing sake)


fruch:
@avelanarius is java-driver is using this?, can you get someone to test with this change ?

zimnx (approved):
LGTM

avelanarius:
@fruch I can confirm that Java Driver integration tests (one test class that uses SNI proxy from ccm) correctly pass after this was merged.

fruch:
> @fruch I can confirm that Java Driver integration tests (one test class that uses SNI proxy from ccm) correctly pass after this was merged.

great, thanks.

Issue #413, opened 2022-11-09
"ccm create" fails with "-v release:2021.1.16"


avelanarius:
When executing `ccm create ccm_1 -i 127.0.1.  -n 1  --scylla  -v release:2021.1.16`, ccm fails with:

```
S3 download: scylla-enterprise-x86_64-unified-package-2021.1.16.0.20221022.3224379e7.tar.gz100%|██████████| 1.12G/1.12G [00:35<00:00, 31.3MB/s]
Extracting /tmp/ccm-3_kulzdn.tar.gz (https://s3.amazonaws.com/downloads.scylladb.com/downloads/scylla-enterprise/relocatable/scylladb-2021.1/scylla-enterprise-x86_64-unified-package-2021.1.16.0.20221022.3224379e7.tar.gz, /tmp/tmpaf6m1de_) as version release/2021.1.16 ...
Relocatable package format version 2 detected.
Usage: install.sh [options]

Options:
  --root /path/to/root     alternative install root (default /)
  --prefix /prefix         directory prefix (default /usr)
  --python3 /opt/python3   path of the python3 interpreter relative to install root (default /opt/scylladb/python3/bin/python3)
  --housekeeping           enable housekeeping service
  --nonroot                install Scylla without required root priviledge
  --sysconfdir /etc/sysconfig   specify sysconfig directory name
  --help                   this helpful message
Traceback (most recent call last):
  File "/home/piotrgrabowski/.local/bin/ccm", line 75, in <module>
    cmd.run()
  File "/home/piotrgrabowski/.local/lib/python3.11/site-packages/ccmlib/cmds/cluster_cmds.py", line 223, in run
    cluster = ScyllaCluster(self.path, self.name, install_dir=self.options.install_dir,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/piotrgrabowski/.local/lib/python3.11/site-packages/ccmlib/scylla_cluster.py", line 50, in __init__
    super(ScyllaCluster, self).__init__(path, name, partitioner,
  File "/home/piotrgrabowski/.local/lib/python3.11/site-packages/ccmlib/cluster.py", line 71, in __init__
    dir, v = self.load_from_repository(version, verbose)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #414, opened 2022-11-09
ccmlib/scylla_repository.py: unified install don't unavailable arguments

fruch:
code was using `--supervisor` to disable systemd checks, but this flag was introduced only in scylla 4.6
and it breaks installing older releases, for example `2021.1.16`

for other flags we check the installer version, but this specifc flag didn't come with a version change, hence we are looking for it in `install.sh`

Fixes: #413


fruch:
@syuu1228 seems like we need to pay attention and update the version, on each flag that is being added or changed.

PR #415, opened 2022-11-14
fix(unified-installer): handle inner direcoty

fruch:
in recent fixes we added a check of the content
of `install.sh` it seems be conflicting with the change made to support `package_version==3.0` which means there a inner directory inside the tarball that
need to be leveled to align to the structure the
ccm code expects


fruch:
testing in:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/197/

fruch:
Running in:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/200/

proved to be working correctly

PR #417, opened 2022-11-15
utils/sni_proxy.py: fix hostname regexes in tables

avelanarius:
Before this change, all requests to `{host_id}.cql.cluster-id.scylla.com` would get erroneously routed to the same node - the first node. The  reason for such behavior is explained by `sniproxy.conf` documentation:

"Each request's hostname is matched against entries in the table in  order, until a match is found and that server is used."

`cql.cluster-id.scylla.com` would match to `{host_id}.cql.cluster-id.scylla.com` and it was the first entry in the table, meaning all requests would get routed to `cql.cluster-id.scylla.com`.

Fix this issue by adding `^` and `$` to the regexes, to make sure that the entire request hostname matches the hostname in the entry.

### DIFF
--- a/ccmlib/utils/sni_proxy.py
+++ b/ccmlib/utils/sni_proxy.py
@@ -96,12 +96,12 @@ def configure_sni_proxy(conf_dir, nodes_info, listen_port=443):
     tables = ""
     mapping = {}
     address, port, host_id = list(nodes_info)[0]
-    tables += f"  cql.cluster-id.scylla.com {address}:{port}\n"
+    tables += f"  ^cql.cluster-id.scylla.com$ {address}:{port}\n"
     mapping['FIRST_ADDRESS'] = address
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM, test with the python driver and it's was working

fruch:
@avelanarius 

do you go a test with this working for you in github actions ?

fruch:
> @avelanarius
> 
> do you go a test with this working for you in github actions ?

I've also taken this change for a ride in the python-driver CI, and it was working

avelanarius:
> @avelanarius
> 
> do you go a test with this working for you in github actions ?

Not currently, I've still fixing https://github.com/scylladb/java-driver/issues/167. I observed ccm wrong behavior in test that started 3 node "serverless" cluster with ccm.

PR #418, opened 2022-11-17
scylla_repository.py: fix the retrival of release canidates version

fruch:
since moving to using unifed package by default the retrival of release canidates were broken

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -69,22 +69,30 @@ def release_packages(s3_url, version, arch='x86_64', scylla_product='scylla'):
     :param version:
     :return:
     """
-    major_version = version if version.count('.') == 1 else '.'.join(n for n in version.split('.')[:2])
-    s3_url = s3_url.format(major_version)
 
+    major_version = extract_major_version(version)
+    s3_url = s3_url.format(major_version)
+    
     files_in_bucket = aws_bucket_ls(s3_url)
 
     if not files_in_bucket:
         raise RuntimeError(
             f"Failed to get release packages list for version {major_version}. URL: {s3_url}.")
-
-    all_unified_packages = [name for name in files_in_bucket if f'{scylla_product}-{arch}-unified-package' in name]
-    candidates = [package for package in all_unified_packages if ".rc" not in package and '.rc' not in version]
-
-    release_unified_packages = [package for package in candidates if version in package]
+    # examples of unified packages file names:
+    # 'scylla-x86_64-unified-package-5.1.2-0.20221225.4c0f7ea09893-5.1.2.0.20221225.4c0f7ea09893.tar.gz'
+    # 'scylla-x86_64-unified-package-5.1.3-0.20230112.addc4666d502-5.1.3.0.20230112.addc4666d502.tar.gz'
+    all_unified_packages = [name for name in files_in_bucket if f'{scylla_product}-{arch}-unified-package' in name or f'{scylla_product}-unified-package' in name]
+    release_unified_packages = [package for package in all_unified_packages if version in package]
 
     def extract_version(filename):
-        version_regex = re.compile(r'-(\d+!)?\d+([.-]\d+)?([.-]\d+)?([a-z]+\d*)?([.-](post|dev|rc)\d*)*')
+        """
+        extract the version out of a unified package filename
+        from 'scylla-x86_64-unified-package-5.1.0~rc0-0.20220810.86a6c1fb2b79-5.1.0-rc0.0.20220810.86a6c1fb2b79.tar.gz'
+        would extract '5.1.0~rc'
+        from 'scylla-x86_64-unified-package-5.1.3-0.20230112.addc4666d502-5.1.3.0.20230112.addc4666d502.tar.gz'
+        would extract '5.1.3'
+        """
+        version_regex = re.compile(r'-(\d+!)?\d+([.-]\d+)?([.-]\d+)?([a-z]+\d*)?([.~-](post|dev|rc)\d*)*')
         match = version_regex.search(filename)
         if match:
             return filename[match.span()[0]+1:match.span()[1]]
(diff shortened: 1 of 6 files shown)


ccmlib/scylla_repository.py
```diff
@@ -69,22 +69,23 @@ def release_packages(s3_url, version, arch='x86_64', scylla_product='scylla'):
...
+    
     files_in_bucket = aws_bucket_ls(s3_url)
 
     if not files_in_bucket:
         raise RuntimeError(
             f"Failed to get release packages list for version {major_version}. URL: {s3_url}.")
 
     all_unified_packages = [name for name in files_in_bucket if f'{scylla_product}-{arch}-unified-package' in name]
-    candidates = [package for package in all_unified_packages if ".rc" not in package and '.rc' not in version]
+    candidates = [package for package in all_unified_packages if any([f'{c}rc' not in package and f'{c}rc' not in version for c in ['.', '-', '~']])]
```
bhalevy, line 83: you want all to mismatch, no?  not any...
anyhow, why not use a regular expression to match `[.~-]rc`?

ccmlib/scylla_repository.py
```diff
@@ -69,22 +69,23 @@ def release_packages(s3_url, version, arch='x86_64', scylla_product='scylla'):
...
 
     release_unified_packages = [package for package in candidates if version in package]
 
     def extract_version(filename):
-        version_regex = re.compile(r'-(\d+!)?\d+([.-]\d+)?([.-]\d+)?([a-z]+\d*)?([.-](post|dev|rc)\d*)*')
+        version_regex = re.compile(r'-(\d+!)?\d+([.-]\d+)?([.-]\d+)?([a-z]+\d*)?([.-~](post|dev|rc)\d*)*')
```
bhalevy, line 88: `[.-~]` is the range of all characters between `'.'` and `'~'` (think `[a-z]`)
`[.~-]` will work better (the '-' can be either first or last to be matched as a regular character.

ccmlib/scylla_repository.py
```diff
@@ -181,6 +184,21 @@ def read_build_manifest(url):
...
+    and formolize them in to a semver base version 2020.2.0~rc5
+    """
+    version = version.replace('-', '~').replace('.rc', '~rc')
+    version = version if version.count('0~rc') else version.replace('~rc', '.0~rc')
+
+    return version
+
+def extract_major_version(version):
+    major_version = version if version.count('.') == 1 else '.'.join(n for n in version.split('.')[:2])
+    major_version = major_version.split('~')[0]
```
bhalevy, line 199: what about `'-'`?

bhalevy (changes requested):

ccmlib/scylla_repository.py
fruch: it by now normalized to be only ~

bhalevy: what if we backport this patch to older branches?

fruch: It should work as well on older branches.

keep in mind we on only use rcs manually.

In CI/Jenkins we only run from relocate getting picked by scylla-pkg

we use the release option only on upgrade tests and only with formal releases

fruch: Anyhow I would refactor this code a bit more

So it can have integration tests (without downloading the whole relocatable file)

fruch: actually had to treat old versions a bit differently, but not here

ccmlib/scylla_repository.py
fruch: fixed

ccmlib/scylla_repository.py
fruch: actually found out, the whole logic isn't needed, since we later scan this list for the matching version.
I think it's intend not to fetch release candidate when asking for something like `release:5.1` which has only RCs.

I think it's point less, if someone asks for a release that in process it's o.k. to receive the release candidate

fruch:
V2:

- added a matrix of tests, to verify we can find all the release and release candidate (with out pulling the artifacts)
- refactor the code a bit, and removed some unneeded logic in the process

fruch:
@juliayakovlev @bhalevy, can one of you give a final look

ccmlib/scylla_repository.py
```diff
@@ -69,22 +69,21 @@ def release_packages(s3_url, version, arch='x86_64', scylla_product='scylla'):
...
 
     if not files_in_bucket:
         raise RuntimeError(
             f"Failed to get release packages list for version {major_version}. URL: {s3_url}.")
 
-    all_unified_packages = [name for name in files_in_bucket if f'{scylla_product}-{arch}-unified-package' in name]
-    candidates = [package for package in all_unified_packages if ".rc" not in package and '.rc' not in version]
-
-    release_unified_packages = [package for package in candidates if version in package]
+    all_unified_packages = [name for name in files_in_bucket if f'{scylla_product}-{arch}-unified-package' in name or f'{scylla_product}-unified-package' in name]
```
juliayakovlev, line 82: Maybe we need names example here

ccmlib/scylla_repository.py
```diff
@@ -69,22 +69,21 @@ def release_packages(s3_url, version, arch='x86_64', scylla_product='scylla'):
...
+    release_unified_packages = [package for package in all_unified_packages if version in package]
 
     def extract_version(filename):
-        version_regex = re.compile(r'-(\d+!)?\d+([.-]\d+)?([.-]\d+)?([a-z]+\d*)?([.-](post|dev|rc)\d*)*')
+        version_regex = re.compile(r'-(\d+!)?\d+([.-]\d+)?([.-]\d+)?([a-z]+\d*)?([.~-](post|dev|rc)\d*)*')
```
juliayakovlev, line 86: And here will be good to see names example, to know which templates are supported

juliayakovlev (changes requested):
LGTM. small comment

ccmlib/scylla_repository.py
fruch: added examples

ccmlib/scylla_repository.py
fruch: added examples

PR #419, opened 2022-11-25
Fixed yaml model (insecureSkipTlsVerify)

wprzytula:
Field insecureSkipTlsVerify was incorrectly put in AuthInfo instead of Datacenter. Therefore, generated .yaml config bundles were invalid.


wprzytula:
@avelanarius

fruch (approved):

fruch:
I'll need to take a look at the python driver, to see if we read it from the correct place. (we don't do validation...

fruch (approved):
LGTM

PR #420, opened 2022-11-30
node.py: change print to logging in `run_cqlsh`

fruch:
in dtest we should be printing anything to stdout, everything should go into the logs.

when ccm is use as standalone (via cli), those would be printed to stdout.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -894,10 +894,11 @@ class Node(object):
             output = p.communicate()
 
             for err in output[1].split('\n'):
-                print_("(EE) ", err, end='')
+                if err.strip():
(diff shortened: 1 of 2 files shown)


ccmlib/node.py
```diff
@@ -866,6 +866,7 @@ def run_cli(self, cmds=None, show_output=False, cli_options=[]):
                     i = i + 1
 
     def run_cqlsh(self, cmds=None, show_output=False, cqlsh_options=None, return_output=False):
+
```
aleksbykov, line 869: Do we need this empty line?

aleksbykov (approved):
LGTM

ccmlib/node.py
fruch: probably not

PR #421, opened 2022-12-14
scylla_node: start: add replace_node parameter

bhalevy:
To be used for replacing a node given its host_id
rather than its ip address.

Makes use of the newly added --replace-node-first-boot scylla option.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -309,10 +309,10 @@ class ScyllaNode(Node):
         if wait_for_binary_proto:
             try:
                 t = timeout * 4 if timeout is not None else 420 if self.cluster.scylla_mode != 'debug' else 900
-                self.wait_for_binary_interface(from_mark=self.mark, process=self._process_scylla, timeout=t)
+                self.wait_for_binary_interface(from_mark=self.mark, process=self._process_scylla, timeout=60)
             except TimeoutError as e:
-                if not self.wait_for_starting(from_mark=self.mark):
-                    raise e
+                if not self.wait_for_starting(from_mark=self.mark, timeout=t):
+                    raise NodeError(f"{e}")
                 pass
 
         return self._process_scylla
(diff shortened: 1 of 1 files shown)


bhalevy:
Ref https://github.com/scylladb/scylladb/pull/12316

This change does not directly depend on the option being added since no code currently
uses the new `replace_node` parameter, while the change is backward compatible with `replace_address`.
but we should wait with merging it until the scylla PR is finalized and merged.

ccmlib/scylla_node.py
```diff
@@ -563,7 +569,10 @@ def start(self, join_ring=True, no_wait=False, verbose=False,
             args += ['--overprovisioned']
         if '--prometheus-address' not in args:
             args += ['--prometheus-address', data['api_address']]
-        if replace_address:
+        if replace_node:
+            assert replace_address is None, "replace_node and replace_address cannot be specified together"
+            args += ['--replace-node-first-boot', replace_node]
```
fruch, line 574: what doesn't first-boot mean, would anyone thing to call it in a different context ? i.e. passing it not on first boot ? 

and it's a bit weird that this commit doesn't say host_id anywhere... i.e. `replace_node` is actually `replace_node_host_id`, i.e. nothing here indicate what's need to be passing down.

bhalevy:
https://github.com/scylladb/scylladb/pull/12316 is merged

ccmlib/scylla_node.py
bhalevy: It basically follows the convention of --replace-endpoint-first-boot.
The details are in the respective documentation that was updated in https://github.com/scylladb/scylladb/pull/12316.

fruch:
> [scylladb/scylladb#12316](https://github.com/scylladb/scylladb/pull/12316) is merged

@bhalevy this have conflict with other change you did with the boot sequence logs

bhalevy:
> > [scylladb/scylladb#12316](https://github.com/scylladb/scylladb/pull/12316) is merged
> 
> @bhalevy this have conflict with other change you did with the boot sequence logs

rebasing...

bhalevy:
@fruch branch was rebased

fruch (changes requested):
LGTM

fruch (approved):
LGTM

PR #424, opened 2023-01-17
node.py: disable commitlog O_DSYNC, preallocation

avikivity:
Commitlog O_DSYNC is intended to make Raft and schema writes durable in the face of power loss. To make O_DSYNC performant, we preallocate the commitlog segments, so that the commitlog writes only change file data and not file metadata (which would require the filesystem to commit its own log).

However, in tests, this causes each ScyllaDB instance to write 384MB of commitlog segments. This overloads the disks and slows everything down.

Fix this by disabling O_DSYNC (and therefore preallocation) during the tests. They can't survive power loss, and run with --unsafe-bypass-fsync anyway.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -582,6 +582,12 @@ class ScyllaNode(Node):
         kernel_page_cache_supported |= current_node_is_enterprise and parse_version(current_node_version) >= parse_version('2022.1.dev')
         if kernel_page_cache_supported and '--kernel-page-cache' not in args:
             args += ['--kernel-page-cache', '1']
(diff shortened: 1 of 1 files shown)


fruch:
@avikivity is there are reference to the change in scylla that started doing this preallocation ? (i'm guess it getting blocked by CI failing in some way)

mykaul:
I think someone complained on something similiar - https://groups.google.com/g/scylladb-dev/c/HJSc6RIIQw0/m/yOADFTuKAwAJ?utm_medium=email&utm_source=footer ?

fruch (approved):
LGTM

PR #425, opened 2023-01-20
scylla_node: expose --memory in bytes

tchaikov:
* add a tiny helper to convert memory size like "1G" to 1024^3. seastar/src/util/conversions.cc supports IEC style postfixes, but since we don't use them, let's keep the implementation simple.
* expose ScyllaNode.memory() so dtest can use it when verifying the system_table. currently, in dtest's system_tables_test.py:TestRuntimeInfoTable, the total size of memory is passed in via `jvm_args` passed to self.prepare_cluster(). to improve the readability, and to be more future-proof, it'd be simpler to expose this setting from node. so dtest can verify the total size of memory exposed by system table by comparing it with the value exposed by node instead of a magic number of keep "1G" as a member variable of the test.

Signed-off-by: Kefu Chai <kefu.chai@scylladb.com>

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -61,6 +61,7 @@ class ScyllaNode(Node):
         self._smp_set_during_test = False
         self._mem_mb_per_cpu = 512
         self._mem_set_during_test = False
+        self._memory = None
         self.__conf_updated = False
         self.scylla_manager = scylla_manager
         self.jmx_pid = None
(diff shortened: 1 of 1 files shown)


fruch (approved):

bhalevy:
Please backport to 5.2
See https://jenkins.scylladb.com/job/scylla-5.2/job/dtest-daily-release/2/testReport/system_tables_test/TestRuntimeInfoTable/Run_Dtest_Parallel_Cloud_Machines___FullDtest___full_split000___test_default_content/
```
>                   assert int(row.value) == node.memory(), f"Unexpected memory value: {row.value}"
E                   AttributeError: 'ScyllaNode' object has no attribute 'memory'
```

fruch:
@bhalevy done

PR #427, opened 2023-01-25
scylla_node: Allow set JAVA_HOME to override scylla-jmx link

### LINKED ISSUE #426: FR: Allow overriding 'scylla-jmx' link source in node creation
Scylla-jmx required jdk8 java to run. Scylla-node links 'scylla-jmx' symlink to hardcoded '/usr/bin/java'. 
On dev machine, running dtest, one has to switch 'alternatives' all the time between 'real' java stuff (CDC drivers etc) and this. 

Would be neat if this could be at least somewhat controlled via environment. A set JAVA_HOME springs to mind.

elcallio:
Fixes #426

To smoothly use different scylla-jmx java executable w/o changing link in /usr/bin.


fruch (approved):
LGTM

fruch:
@elcallio can you add example of usage of it in dtest readme ? for others that might need it ?

PR #430, opened 2023-02-10
node: keep hostid result in self.node_hostid

bhalevy:
The hostid nver changes so there no reason to
use nodetool every time to get it.

Tests that want to exercise nodetool for that can do that by explicitly running nodetool, or passing force_refresh=True if they insist :)

Otherwise, cache the result in self.node_hostid
to be reused next time hostid() is called.

This is particularly needed for node operations
scenarios when an operation like replacenode
runs after removenode that happened to stop the
to-be-replaced node.

Since scylladb/scylla-dtest@97a341adc4adeb1a27e00db1bb4614b0d1f53618 the replace node operation operates on the node hostid and that cannot be retrieved if the node is already dead without this change.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1362,16 +1362,18 @@ class Node(object):
         self.status = Status.DECOMMISSIONED
         self._update_config()
 
-    def hostid(self, timeout=60):
-        info = self.nodetool('info', capture_output=True, timeout=timeout)[0]
-        id_lines = [s for s in info.split('\n')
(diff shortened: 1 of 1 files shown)


fruch (approved):

bhalevy:
@fruch please backport to 5.2.
See https://jenkins.scylladb.com/job/scylla-5.2/job/dtest-daily-release/3/testReport/repair_based_node_operations_test/TestRepairBasedNodeOperations/

PR #431, opened 2023-02-15
scylla_repository: Fix handling of full version release candidates

fruch:
when ccm was pass `2023.1.0~rc0` it would fail like this:

```
  File ".../site-packages/ccmlib/scylla_repository.py", line 136, in release_packages
    raise ValueError(f"Not expected version number: {version}. S3_URL: {s3_url}")
ValueError: Not expected version number: 2023.1.0.0~rc0. S3_URL: https://s3.amazonaws.com/downloads.scylladb.com/downloads/scylla-enterprise/relocatable/scylladb-2023.1
```

it was a faulty regex that wasn't behaving correctly also fix the filtering of unified packages, since the nameing changed a bit, once again

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -81,7 +81,9 @@ def release_packages(s3_url, version, arch='x86_64', scylla_product='scylla'):
     # examples of unified packages file names:
     # 'scylla-x86_64-unified-package-5.1.2-0.20221225.4c0f7ea09893-5.1.2.0.20221225.4c0f7ea09893.tar.gz'
     # 'scylla-x86_64-unified-package-5.1.3-0.20230112.addc4666d502-5.1.3.0.20230112.addc4666d502.tar.gz'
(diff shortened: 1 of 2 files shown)


Bouncheck (approved):
LGTM
Creating cluster with described version also worked.

ccmlib/scylla_repository.py
```diff
@@ -204,7 +206,7 @@ def normalize_scylla_version(version):
         return version
 
     version = version.replace('-', '~').replace('.rc', '~rc')
-    version = version if re.match(r'\d*.\d*.0~rc', version) else version.replace('~rc', '.0~rc')
```
juliayakovlev, line 207: Do we expectect `.0~` always in the version. Can it be `.1~` for example?

fruch: no, release candidate are always on the first release, 2023.1.0~rc0. 5.2.0~rc0

avelanarius:
ping @fruch - this is required to fix GitHub Actions CI in Java Driver repo.

PR #433, opened 2023-03-01
scylla_node: expose ScyllaNode.smp()

tchaikov:
so the dtest tests are able to acess the exact number of shard used
for testing. for instance, some of them might need to calculate the
exact total memory based on this option.


fruch (approved):
LGTM

PR #436, opened 2023-03-14
scylla_node: set default smp count to 2

bhalevy:
This is how we run dtests by default
so it gets really confusing for users who don't provide --smp in SCYLLA_EXT_OPTS

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -56,7 +56,7 @@ class ScyllaNode(Node):
         self._process_scylla_waiter = None
         self._process_agent = None
         self._process_agent_waiter = None
-        self._smp = 1
+        self._smp = 2
         self._smp_set_during_test = False
         self._mem_mb_per_cpu = 512
         self._mem_set_during_test = False


bhalevy:
If they care about smp, they should set it in the jvm_args

bhalevy:
With this change, I'd like to remove the default setting of smp and memory in
https://github.com/scylladb/scylla-dtest/blob/7f666e8bc85efa37a6820eb25db8692405ef7794/scripts/run_test.sh#L98
and
https://github.com/scylladb/scylla-pkg/blob/66e6297ad43f6e707b967f3718d714aaa59a7708/scripts/jenkins-pipelines/shell_scripts/dtest.sh#L430

fruch:
> If they care about smp, they should set it in the jvm_args or SCYLLA_EXT_OPTS

yeah but if they don't, we are changing the default under "their" feet, might break some tests.

But I'm o.k. with letting it break, the fixes in that case should be easy, to pin it back in the test code, and not CCM

fruch (approved):
LGTM

PR #440, opened 2023-03-15
scylla_repository: re-raise exception on `run()`

fruch:
69d304 introduced change to `run()` so we can log
better the output of commands, but it failed
to raise the error after that.

that made the output a bit confusing and failed the whole idea.


fgelcer (approved):
LGTM

PR #441, opened 2023-03-17
Replace eu-west-1 with system.local data_center value

Gor027:
The `create_cloud_config` function generates files config_data.yaml and config_path.yaml. The files were created with the hardcoded datacenter name eu-west-1 which is now replaced by fetching the real datacenter name from the node's `system.local` table.


fruch (approved):
LGTM

fruch:
@Gor027 you tested it with the test in both java and python drivers ?

Gor027:
> @Gor027 you tested it with the test in both java and python drivers ?

I have tested it with the Rust driver: https://github.com/scylladb/scylla-rust-driver/blob/main/examples/cloud.rs

PR #444, opened 2023-03-23
Add support for multi-dc cloud setup

Gor027:
Currently, only one SNI proxy container is created for the whole cluster for serverless support. So, the generated configuration files `config_data.yaml` and `config_path.yaml` contain only local datacenter information.
 This PR adds support for multi-dc cloud setup, so when starting a multi-DC cluster with SNI proxy, an SNI proxy container will be created per DC, so that each one can only connect to nodes in their datacenter. This will allow for better testing of our drivers after adding serverless support.


ccmlib/utils/sni_proxy.py
```diff
@@ -96,26 +104,32 @@ def configure_sni_proxy(conf_dir, nodes_info, listen_port=443):
...
     with open(sniproxy_conf_path, 'w') as fp:
         fp.write(tmpl.substitute(TABLES=tables, **mapping))
 
     return sniproxy_conf_path
 
 
-def start_sni_proxy(conf_dir, nodes_info, listen_port=443):
-    address, _, _, _ = list(nodes_info)[0]
-    sniproxy_conf_path = configure_sni_proxy(conf_dir, nodes_info, listen_port=listen_port)
+def start_sni_proxy(conf_dir, nodes_info_per_dc, listen_port=443):
```
fruch, line 130: please don't change the APIs, this is breaking the python-driver code tests.
this are calling this API directly

fruch: the python-driver are not using ccm from  commandline like driver tests

ccmlib/utils/sni_proxy.py
```diff
@@ -33,38 +33,46 @@ def file_or_memory(path=None, data=None):
...
 def create_cloud_config(ssl_dir, port, address, nodes_info, username='cassandra', password='cassandra'):
-
     def encode_base64(filename):
         return base64.b64encode(open(os.path.join(ssl_dir, filename), 'rb').read()).decode()
 
     cadata = encode_base64('ccm_node.cer')
     certificate_data = encode_base64('ccm_node.cer')
     key_data = encode_base64('ccm_node.key')
-    _, _, _, data_center = list(nodes_info)[0]  # TODO: Add ability to create cloud config for multi-DC setup
+    default_dc = "dc1"  # TODO: make default datacenter configurable
```
fruch, line 42: at least take it from the information supplied in `node_info`, and not hard-code ?

Gor027: Fixed it.

ccmlib/utils/sni_proxy.py
Gor027: I have reverted API changing modifications.

fruch:
@Gor027 Can you take for a ride with the python-driver, to verify it's not breaking it ?

I won't be able to get to it until Monday...

Gor027:
> @Gor027 Can you take for a ride with the python-driver, to verify it's not breaking it ?
> 
> I won't be able to get to it until Monday...

I tested these changes with python-driver by triggering CI on my fork with `scylla-ccm` installed from my branch:
https://github.com/Gor027/python-driver/pull/1

Gor027:
@fruch ping

fruch (approved):
LGTM

But still want to check why CI is failing

fruch:
seem like @avelanarius found the issue with CI: https://github.com/scylladb/scylla-ccm/issues/448

PR #446, opened 2023-03-26
tox.ini: remove tox.ini

tchaikov:
* remove tox.ini
* drop .tox from .gitignore.

tox.ini was added back to 732b8f897e706243b4270b611dd4ac1b6ae8c44c, but we don't use it in our CI or daily practice. also, the nose test framework is not actively maintained anymore. see https://pypi.org/project/nose/ so let's drop it.


fruch (approved):
LGTM

Issue #448, opened 2023-04-13
CCM CI is failing to even start


avelanarius:
Our CCM GitHub Actions CI is using a Ubuntu 18.04 image. However, this image is deprecated and even scheduled to be removed: https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

GitHub even purpusefully breaks the jobs running on it "to raise awareness of the upcoming removal" (not kidding - see the blogpost above). It looks like they removed it for good right now.

We should update the CI to use the newer Ubuntu image.

PR #449, opened 2023-04-13
github actions: update to image ubuntu-latest

fruch:
- since the image we are usign it deprecated, switch to latest
- also update version to some of the actions

Fixes: #448


.github/workflows/integration-tests.yml
```diff
@@ -12,14 +12,14 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
mykaul, line 15: What happens when 23 is out? Don't we wish to have more granularity and control over the version?

fruch: this small CI isn't that complex, I'm o.k. with using latest.
rather then get it break cause of deprecations

PR #450, opened 2023-04-18
make_cassandra_env: hardcode java_home to jvm8

fruch:
since we are switching to use jvm11 for scylla, but still have test that are using cassandra 3.11 that works only with jvm8, we are hardwiring the code to always try to select the jvm8 for cassandra

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -14,8 +14,9 @@ import sys
 import time
 import tempfile
 import logging
+import pathlib
 from itertools import zip_longest
-from typing import Callable
+from typing import Callable, Optional
 
 import yaml
 from boto3.session import Session
@@ -192,7 +193,7 @@ def rmdirs(path):
         shutil.rmtree(path, ignore_errors=True)
 
 
-def make_cassandra_env(install_dir, node_path, update_conf=True):
+def make_cassandra_env(install_dir, node_path, update_conf=True, hardcode_java_version: Optional[str] = None):
     if is_win() and get_version_from_build(node_path=node_path) >= '2.1':
         sh_file = os.path.join(CASSANDRA_CONF_DIR, CASSANDRA_WIN_ENV)
     else:
@@ -241,6 +242,22 @@ def make_cassandra_env(install_dir, node_path, update_conf=True):
     env['cassandra.config'] = "file://" + os.path.join(node_path, 'conf', 'cassandra.yaml')
     env['CASSANDRA_HOME'] = install_dir
     env['CASSANDRA_CONF'] = os.path.join(node_path, 'conf')
+
+    if hardcode_java_version:
+        known_jvm_names = {
+            '8': ('8-openjdk', '8-jdk', '1.8.0-openjdk'),
+            '11': ('11-openjdk', '11-jdk', '1.11.0-openjdk'),
+        }
+        assert hardcode_java_version in known_jvm_names.keys(), \
+            f"hardcode_java_version={hardcode_java_version} not supported in:\n{known_jvm_names}"
+
+        for java_path in pathlib.Path('/usr/lib/jvm/').iterdir():
+            if java_path.is_dir and any(name in java_path.name for name in known_jvm_names[hardcode_java_version]):
+                env['JAVA_HOME'] = str(java_path)
+                break
+        else:
+            raise ArgumentError("java-8 wasn't found in /usr/lib/jvm/")
+
     return env
 
 
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -185,7 +185,7 @@ class Node(object):
         update_conf = not self.__conf_updated
         if update_conf:
             self.__conf_updated = True
-        return common.make_cassandra_env(self.get_install_dir(), self.get_path(), update_conf)
+        return common.make_cassandra_env(self.get_install_dir(), self.get_path(), update_conf, hardcode_java_version='8')
 
     def get_install_cassandra_root(self):
         return self.get_install_dir()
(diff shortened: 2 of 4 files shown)


mykaul:
> since we are switching to use jvm11 for scylla, but still have test that are using cassandra 3.11 that works only with jvm8, we are hardwiring the code to always try to select the jvm8 for cassandra

Shouldn't we switch to testing with C* 4.x? I know some of the drivers CI did.

ccmlib/common.py
```diff
@@ -241,6 +242,18 @@ def make_cassandra_env(install_dir, node_path, update_conf=True):
     env['cassandra.config'] = "file://" + os.path.join(node_path, 'conf', 'cassandra.yaml')
     env['CASSANDRA_HOME'] = install_dir
     env['CASSANDRA_CONF'] = os.path.join(node_path, 'conf')
+
+    if hardcode_java_home_8:
+        output = subprocess.check_output(["update-alternatives", "--display", "java"], text=True).splitlines()
+        output = [l.split()[0] for l in output if l.startswith('/')]
+        for java_path in output:
+            known_jvm8_names = ['8-openjdk', '8-jdk']
+            if any([name in java_path for name in known_jvm8_names]):
```
tchaikov, line 251: i'd suggest change the parameter to `java_version`, which defaults to `1.8.0`, and put something like
```py
if java_version is not None:
    java_home = f"/usr/lib/jvm/java-{java_version}"
    java = shutil.which("java', path=os.path.join(java_home, 'bin'))
    if java is None:
        raise ArgumentError(f"java-8 wasn't found in alternatives: {output}")
    env['JAVA_HOME'] = java_home
```

fruch: each env we run this had totally different names, there are multiple `java_version` we should be looking.

fruch:
> > since we are switching to use jvm11 for scylla, but still have test that are using cassandra 3.11 that works only with jvm8, we are hardwiring the code to always try to select the jvm8 for cassandra
> 
> Shouldn't we switch to testing with C* 4.x? I know some of the drivers CI did.

there are 5 tests that uses cassandra 3.11, we were waiting for a few years for 4.0 to come out of beta :)
more then that there is test data from 3.X, and we probably should generate test data for 4.0 

Anyhow that work shouldn't hold us from switching JVM for scylla own parts.

ccmlib/common.py
tchaikov: i see. the suggestion was posted to explain the possibility to trade a hardwired boolean to a more flexible variable, if we want to test multiple, we can use a list instead of a single value here. and even your implementation only test for a single java version, right?

fruch: supporting multiple versions, would need much more work then this.

the whole thing I'm trying to avoid using JAVA_HOME, since it would affect also other tools.

I'll open up a an issue for ccm, so we can support both 3.11 and 4.0 and up

fruch: we want to support also ubuntu, and running github actions (which we use something else)

using `subprocess.check_output` vs. `shutil.which` isn't big of a difference from my POV

fruch:
> > since we are switching to use jvm11 for scylla, but still have test that are using cassandra 3.11 that works only with jvm8, we are hardwiring the code to always try to select the jvm8 for cassandra
> 
> Shouldn't we switch to testing with C* 4.x? I know some of the drivers CI did.

raised it in dtest:
https://github.com/scylladb/scylla-dtest/issues/3125

ccmlib/common.py
fruch: > > supporting multiple versions, would need much more work then this.
> > the whole thing I'm trying to avoid using JAVA_HOME, since it would affect also other tools.
> 
> i am confused. you are actually using it. and it only affect the process which consumes the return value of `make_cassandra_env()`.

`env` is passed down directly to the process of cassandra, setting JAVA_HOME by the user/test would affect all process after, like scylla jmx processes.

tchaikov: > we want to support also ubuntu, and running github actions (which we use something else)
> 
> using `subprocess.check_output` vs. `shutil.which` isn't big of a difference from my POV

i see. guess that's why you prefer using `update-alternatives --display java`.

fruch: BTW I can confirm Cassandra 4.1 was working o.k. with jvm8, at least for our dtest needs (only running a few tests)

fruch: > BTW I can confirm Cassandra 4.1 was working o.k. with jvm8, at least for our dtest needs (only running a few tests)

Should add cassandra 4.1 to the integration tests CI of CCM

ccmlib/common.py
```diff
@@ -241,6 +242,18 @@ def make_cassandra_env(install_dir, node_path, update_conf=True):
     env['cassandra.config'] = "file://" + os.path.join(node_path, 'conf', 'cassandra.yaml')
     env['CASSANDRA_HOME'] = install_dir
     env['CASSANDRA_CONF'] = os.path.join(node_path, 'conf')
+
+    if hardcode_java_home_8:
+        output = subprocess.check_output(["update-alternatives", "--display", "java"], text=True).splitlines()
```
tchaikov, line 247: i still don't like the way how we are using `update-alternatives`, this command is a user facing tool in the sense that its output format is supposed to be consumed by human. so the way to parse its output is fragile. i'd suggest use something like https://github.com/scylladb/scylla-jmx/blob/master/scripts/select-java .

ccmlib/common.py
```diff
@@ -241,6 +242,18 @@ def make_cassandra_env(install_dir, node_path, update_conf=True):
...
+        output = [l.split()[0] for l in output if l.startswith('/')]
+        for java_path in output:
+            known_jvm8_names = ['8-openjdk', '8-jdk', '1.8.0-openjdk']
+            if any([name in java_path for name in known_jvm8_names]):
```
tchaikov, line 251:
```suggestion
            if any(name in java_path for name in known_jvm8_names):
```

tchaikov (approved):

Issue #456, opened 2023-05-21
JMX doesn't start


nyh:
I'm trying to run dtest on my machine, and the cluster fails to start reporting that it can't connect to JMX:

```
>       self.cluster.start(wait_other_notice=True, wait_for_binary_proto=True)
...
        if not self._wait_java_up(ip_addr, jmx_port):
            e_msg = "Error starting node {}: unable to connect to scylla-jmx port {}:{}".format(
                     [self.name](http://self.name/), ip_addr, jmx_port)
>           raise NodeError(e_msg, scylla_process)
E           ccmlib.node.NodeError: Error starting node node1: unable to connect to scylla-jmx port [127.0.71.1:7199](http://127.0.71.1:7199/)
```

In logs/system.log.jmx I see only this:
```
Starting scylla-jmx: args=['/home/nyh/.dtest/dtest-j7k_32_1/test/node1/bin/symlinks/scylla-jmx', '-Dapiaddress=127.0.93.1', '-Djavax.management.builder.initial=com.scylladb.jmx.utils.APIBuilder', '-Djava.rmi.server.hostname=127.0.93.1', '-Dcom.sun.management.jmxremote', '-Dcom.sun.management.jmxremote.host=127.0.93.1', '-Dcom.sun.management.jmxremote.port=7199', '-Dcom.sun.management.jmxremote.rmi.port=7199', '-Dcom.sun.management.jmxremote.local.only=false', '-Xmx256m', '-XX:+UseSerialGC', '-Dcom.sun.management.jmxremote.authenticate=false', '-Dcom.sun.management.jmxremote.ssl=false', '-jar', '/home/nyh/.dtest/dtest-j7k_32_1/test/node1/bin/scylla-jmx-1.0.jar']
```

I don't see any other output in this log, which I also want to report as a bug, but of course what bothers me more is why it doesn't start.

I see in the above command line that it runs `node1/bin/symlinks/scylla-jmx` as Java. But this file is (I checked now) a link to `/usr/bin/java`.  But my /usr/bin/java is Java 17, and if I understand correctly, this isn't supported by JMX :-( 

I see in the CCM code there is some attempt to find among the Javas I have installed if I have 8 or 11 installed (I have Fedora 37, and both Java 8 and 11 installed!), but perhaps for some reason this doesn't work?

CC @fruch

mykaul:
Might be related to some changes @tchaikov has been doing in this area?

nyh:
[...]

PR #457, opened 2023-05-21
use jmx `select-java` script to execute jmx

fruch:
since scylladb/scylla-jmx@96adf0ef51ad49b6abb98867f73cf69705f5f3e3 introduced this new script, the ccm code should be looking for it, and using it insted of using the default java or hoping for the user to set JAVA_HOME correcrtly to JMX needs

Ref: https://github.com/scylladb/scylla-jmx/commit/96adf0ef51ad49b6abb98867f73cf69705f5f3e3

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1041,19 +1041,29 @@ class ScyllaNode(Node):
                                    os.path.join(self.get_bin_dir(), 'scylla-jmx-1.0.jar'), replace=replace)
             self.hard_link_or_copy(os.path.join(self.get_jmx_dir(), 'scylla-jmx'),
                                    os.path.join(self.get_bin_dir(), 'scylla-jmx'), replace=replace)
+            select_java = Path(self.get_jmx_dir()) / 'select-java'
         else:
             self.hard_link_or_copy(os.path.join(self.get_jmx_dir(), 'target', 'scylla-jmx-1.0.jar'),
                                    os.path.join(self.get_bin_dir(), 'scylla-jmx-1.0.jar'), replace=replace)
             self.hard_link_or_copy(os.path.join(self.get_jmx_dir(), 'scripts', 'scylla-jmx'),
                                    os.path.join(self.get_bin_dir(), 'scylla-jmx'), replace=replace)
+            select_java = Path(self.get_jmx_dir()) / 'scripts' / 'select-java'
 
         os.makedirs(os.path.join(self.get_bin_dir(), 'symlinks'), exist_ok=exist_ok)
         scylla_jmx_file = os.path.join(self.get_bin_dir(), 'symlinks', 'scylla-jmx')
         if os.path.exists(scylla_jmx_file) and replace:
             os.remove(scylla_jmx_file)
-        java_home = os.environ.get('JAVA_HOME', '/usr')
-        java_exe = os.path.join(java_home, 'bin', 'java')
-        os.symlink(java_exe, scylla_jmx_file)
+        if java_home := os.environ.get('JAVA_HOME'):
+            # user selecting specific Java
+            java_exe = Path(java_home) / 'bin' / 'java'
+            os.symlink(java_exe, scylla_jmx_file)
+        elif select_java.exists():
+            # JMX lookup logic
+            os.symlink(select_java, scylla_jmx_file)
+        else:
+            # older scylla versions, just use default java
+            java_exe = Path('/usr') / 'bin' / 'java'
+            os.symlink(java_exe, scylla_jmx_file)
 
         parent_dir = os.path.dirname(os.path.realpath(__file__))
         resources_bin_dir = os.path.join(parent_dir, 'resources', BIN_DIR)


nyh (approved):
Excellent, it works! (please add "Fixes #456" to the commit message). 

HOWEVER, what is sad is that the test directory still has symlinks/scylla-jmx being created pointing to /usr/bin/java (I still can't figure out who does it! maybe dtest?), but at least now it's no longer being used and dtest works.

fruch:
> Excellent, it works! (please add "Fixes #456" to the commit message).
> 
> HOWEVER, what is sad is that the test directory still has symlinks/scylla-jmx being created pointing to /usr/bin/java (I still can't figure out who does it! maybe dtest?), but at least now it's no longer being used and dtest works.

i'm keeping the `symlinks/scylla-jmx` since I still won't to be able to use older scylla versions that doesn't have the `java-select` script.

I agree it might be a bit confusing, but the situation is a bit more under control. since the two usage in CCM of java, jmx and Cassandra now can select the correct java on their own, regardless of the user selection of jvm.

ccmlib/scylla_node.py
```diff
@@ -191,6 +191,9 @@ def _start_jmx(self, data):
         jmx_jar_dir = os.path.join(self.get_path(), BIN_DIR)
         jmx_java_bin = os.path.join(jmx_jar_dir, 'symlinks', 'scylla-jmx')
         jmx_jar = os.path.join(jmx_jar_dir, 'scylla-jmx-1.0.jar')
+        jmx_select_java = Path(jmx_jar_dir) / 'select-java'
+        if jmx_select_java.exists():
+            jmx_java_bin = jmx_select_java
```
nyh, line 196: Nitpick: I would put the first setting of jmx_java_bin above as a "else" here. It will make it less confusing when somebody comes back to this code a year from now and doesn't understand which of the code is relevant.

Another note (slightly more serious):  I don't know why somebody called this symlink "scylla-jmx" but I suspect it was so in "top" or "killall" you'll see the string "scylla-jmx". But with your change, it will now be called "select-java", not "scylla-jmx". Perhaps if you want to do something more along the line of the original goal of this code, instead of modifying this code you should modify the code which creates this link in the first place - and link select-java to symlinks/scylla-jmx?

ccmlib/scylla_node.py
```diff
@@ -1041,11 +1044,19 @@ def run_patchelf(patchelf_cmd):
                                    os.path.join(self.get_bin_dir(), 'scylla-jmx-1.0.jar'), replace=replace)
             self.hard_link_or_copy(os.path.join(self.get_jmx_dir(), 'scylla-jmx'),
                                    os.path.join(self.get_bin_dir(), 'scylla-jmx'), replace=replace)
+            select_java = Path(self.get_jmx_dir()) / 'select-java'
+            if select_java.exists():
+                self.hard_link_or_copy(select_java,
+                                       Path(self.get_bin_dir()) / 'select-java', replace=replace)
```
nyh, line 1050: I don't understand what this code does. You are making a copy of select-java itself? Why? Also, does the _start_jmx() above actually use this copy, or the original one?

nyh (approved):

ccmlib/scylla_node.py
fruch: It uses the copy,

It's common across ccm that we copy the artifacts we use into the cluster code

We could just symlink it, in this case should be the same

ccmlib/scylla_node.py
fruch: I am quite sure we aren't looking at the processes by name like that, since we are saving the pid of the process and kill it with the pid.

fruch: I've took you advice and now we symlink `select-java` into `scylla-jmx`
it's simplify the change into one place.

ccmlib/scylla_node.py
```diff
@@ -1041,19 +1041,24 @@ def run_patchelf(patchelf_cmd):
...
+            select_java = Path(self.get_jmx_dir()) / 'scripts' / 'select-java'
 
         os.makedirs(os.path.join(self.get_bin_dir(), 'symlinks'), exist_ok=exist_ok)
         scylla_jmx_file = os.path.join(self.get_bin_dir(), 'symlinks', 'scylla-jmx')
         if os.path.exists(scylla_jmx_file) and replace:
             os.remove(scylla_jmx_file)
-        java_home = os.environ.get('JAVA_HOME', '/usr')
-        java_exe = os.path.join(java_home, 'bin', 'java')
-        os.symlink(java_exe, scylla_jmx_file)
+        if select_java and not os.environ.get('JAVA_HOME'):
```
fruch, line 1056: I've decided that if the use want to point to a specific JVM we'll still gonna honor JAVA_HOME

tchaikov: no need to check `select_java` here, as it is always evaluated to True here. as  you assigns to this variable with a `Path` in both branches above. i don't see another branch leading to this line.

also, can we favor positive conditions over negative ones? i'd suggest use something like:
```py
java_home = os.environ.get('JAVA_HOME')
if java_home:
    java_exe = os.path.join(java_home, 'bin', 'java')
    os.symlink(java_exe, scylla_jmx_file)
else:
    # use select-java as a fallback, if $JAVA_HOME is not defined.
    os.symlink(select_java, scylla_jmx_file)
```

fruch: > no need to check `select_java` here, as it is always evaluated to True here. as  you assign to this variable with a `Path` in both branches above. i don't see another branch leading to this line.
> 
> also, can we favor positive conditions over negative ones? i'd suggest use something like:
> ```py
> java_home = os.environ.get('JAVA_HOME')
> if java_home:
>     java_exe = os.path.join(java_home, 'bin', 'java')
>     os.symlink(java_exe, scylla_jmx_file)
> else:
>     # use select-java as a fallback, if $JAVA_HOME is not defined.
>     os.symlink(select_java, scylla_jmx_file)
> ```

I can reverse the logic, but we should still support case select-java doesn't exist, i.e. old releases

nyh (changes requested):

ccmlib/scylla_node.py
nyh: Anyway I can confirm the current code works if you do have the select-java script (new Scylla). Thanks.

fruch: fix the logic to be a bit more straight forward, with no negatives

tchaikov (approved):

nyh (approved):

PR #462, opened 2023-05-28
scylla_cluster: fix handling of wait_other_notice

### LINKED ISSUE #461: wait_other_notice doesn't wait for the right thing
ccmlib's "wait_other_notice" option wants to wait until all the nodes of the cluster are fully aware of each other. This is necessary, for example, if you want to send a CL=ALL request to this cluster - and don't want the coordinator to think one of the replicas is not available and fail the request.

In today's CCM implementation, this wait is implemented by the `watch_log_for_alive()` function in ccmlib/node.py. This function says that node 127.0.0.1 is fully aware of node 127.0.0.2 when it sees in 127.0.0.1's log a message like

```
gossip - InetAddress 127.0.60.2 is now UP, status = UNKNOWN
```

The problem is that Scylla's code doesn't really print this message as the last thing of becoming "fully aware" that the new node is up. Not only does it print it in the middle of some function, even worse - the node becoming "up" isn't the end of the story. After a node comes up it is still listed as "joining" (getting ranges, copying data, etc.) and while it is listed as "joining", the other node still won't use it!

So instead of relying on an untrustworthy log message, CCM should use the ScyllaDB REST API - wait until the other node is in `http://{self.address()}:10000/gossiper/endpoint/live` but **not** in  `http://{self.address()}:10000/storage_service/nodes/joining`. 

I tested this in the dtest that used to fail because of this bug (`consistency_test.py::TestConsistency::test_short_read_delete`) and it seems now to work flawlessly.

Note that my proposed solution will only work on ScyllaDB, not Cassandra. I think this is acceptable for now (CCM has different source files for Cassandra anyway).

FYI @fruch @asias @kostja

nyh:
After starting a multi-node cluster, it's important to wait until all nodes are aware that all other nodes are available, otherwise if the user sends a CL=ALL request to one node, it might not be aware that the other replicas are usable, and fail the request.

For this reason a cluster's start() has a wait_other_notice=True option, and dtests correctly use it. However, the implementation doesn't wait for the right thing... To check if node A thinks that B is available, it checks that A's log contains the message "InetAddress B is now UP". But this message is unreliable - when it is printed, A still doesn't think that B is fully available - it can still think that B is in a "joining" state for a while longer. If wait_other_notice returns at this point, and the user sends a CL=ALL request to node A, it will fail.

The solution I propose in this patch uses the REST API, instead of the log, to wait until node A thinks node B is both live and finished joining.

This patch is needed if Scylla is modified to boot up faster. We start seeing dtests which use RF=ALL in the beginning of a test failing, because the node we contact doesn't know that the other nodes are usable.

Fixes #461

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -143,10 +143,10 @@ class ScyllaCluster(Cluster):
                                    verbose=verbose, from_mark=mark)
 
         if wait_other_notice:
-            for old_node, mark in marks:
+            for old_node, _ in marks:
                 for node, _, _ in started:
                     if old_node is not node:
-                        old_node.watch_log_for_alive(node, from_mark=mark)
+                        old_node.watch_rest_for_alive(node)
 
         return started
 
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -311,6 +311,7 @@ class ScyllaNode(Node):
             for node, _ in marks:
                 t = timeout if timeout is not None else 120 if self.cluster.scylla_mode != 'debug' else 360
                 node.watch_rest_for_alive(self, timeout=t)
+                self.watch_rest_for_alive(node, timeout=t)
 
         if wait_for_binary_proto:
             t = timeout * 4 if timeout is not None else 420 if self.cluster.scylla_mode != 'debug' else 900
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -18,6 +18,7 @@ import psutil
 import yaml
 import glob
 import re
+import requests
 
 from ccmlib.common import CASSANDRA_SH, BIN_DIR, wait_for, copy_directory
 
@@ -307,9 +308,9 @@ class ScyllaNode(Node):
                                 self._process_scylla)
 
         if wait_other_notice:
-            for node, mark in marks:
+            for node, _ in marks:
                 t = timeout if timeout is not None else 120 if self.cluster.scylla_mode != 'debug' else 360
-                node.watch_log_for_alive(self, from_mark=mark, timeout=t)
+                node.watch_rest_for_alive(self, timeout=t)
 
         if wait_for_binary_proto:
             t = timeout * 4 if timeout is not None else 420 if self.cluster.scylla_mode != 'debug' else 900
@@ -1336,6 +1337,34 @@ class ScyllaNode(Node):
 
     def rollback(self, upgrade_to_version):
         self.upgrader.upgrade(upgrade_version=upgrade_to_version, recover_system_tables=True)
+
+    def watch_rest_for_alive(self, nodes, timeout=120):
+        """
+        Use the REST API to wait until this node detects that the nodes listed
+        in "nodes" become fully operational (live and no longer "joining").
+        This is similar to watch_log_for_alive but uses ScyllaDB's REST API
+        instead of the log file and waits for the node to be really useable,
+        not just "UP" (see issue #461)
+        """
+        tofind = nodes if isinstance(nodes, list) else [nodes]
+        tofind = set([node.address() for node in tofind])
+        url_live = f"http://{self.address()}:10000/gossiper/endpoint/live"
+        url_joining = f"http://{self.address()}:10000/storage_service/nodes/joining"
+        endtime = time.time() + timeout
+        while time.time() < endtime:
+            live = set()
+            response = requests.get(url=url_live)
+            if response.text:
+                live = set(response.json())
+            response = requests.get(url=url_joining)
+            if response.text:
+                live = live - set(response.json())
+            if tofind.issubset(live):
+                # This node thinks that all given nodes are alive and not
+                # "joining", we're done.
+                return
+            time.sleep(0.1)
+        raise TimeoutError(f"watch_rest_for_alive() timeout after {timeout} seconds")
 
     @property
     def gnutls_config_file(self):
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1341,7 +1341,7 @@ class ScyllaNode(Node):
     def watch_rest_for_alive(self, nodes, timeout=120):
         """
         Use the REST API to wait until this node detects that the nodes listed
-        in "nodes" become fully operational (live and no longer "joining").
+        in "nodes" become fully operational and knows of its tokens.
         This is similar to watch_log_for_alive but uses ScyllaDB's REST API
         instead of the log file and waits for the node to be really useable,
         not just "UP" (see issue #461)
@@ -1350,6 +1350,7 @@ class ScyllaNode(Node):
         tofind = set([node.address() for node in tofind])
         url_live = f"http://{self.address()}:10000/gossiper/endpoint/live"
         url_joining = f"http://{self.address()}:10000/storage_service/nodes/joining"
+        url_tokens = f"http://{self.address()}:10000/storage_service/tokens/"
         endtime = time.time() + timeout
         while time.time() < endtime:
             live = set()
@@ -1359,9 +1360,19 @@ class ScyllaNode(Node):
             response = requests.get(url=url_joining)
             if response.text:
                 live = live - set(response.json())
+            # Verify that node knows not only about the existance of the
+            # other node, but also its tokens:
             if tofind.issubset(live):
                 # This node thinks that all given nodes are alive and not
-                # "joining", we're done.
+                # "joining", we're almost done, but still need to verify
+                # that the node knows the others' tokens.
+                check = tofind
+                tofind = set()
+                for n in check:
+                    response = requests.get(url=url_tokens+n)
+                    if response.text == '[]':
+                        tofind.add(n)
+            if not tofind:
                 return
             time.sleep(0.1)
         raise TimeoutError(f"watch_rest_for_alive() timeout after {timeout} seconds")


ccmlib/scylla_cluster.py
```diff
@@ -146,7 +146,7 @@ def start_nodes(self, nodes=None, no_wait=False, verbose=False, wait_for_binary_
             for old_node, mark in marks:
```
fruch, line 146: just for readability, to show this isn't used 
```suggestion
            for old_node, _ in marks:
```

ccmlib/scylla_node.py
```diff
@@ -309,7 +310,7 @@ def _start_scylla(self, args, marks, update_pid, wait_other_notice,
         if wait_other_notice:
             for node, mark in marks:
```
fruch, line 311:
```suggestion
            for node, _ in marks:
```

ccmlib/scylla_node.py
```diff
@@ -1335,6 +1336,34 @@ def upgrade(self, upgrade_to_version):
...
+            if response.text:
+                live = set(response.json())
+            response = requests.get(url=url_joining)
+            if response.text:
+                live = live - set(response.json())
+            if tofind.issubset(live):
+                # This node thinks that all given nodes are alive and not
+                # "joining", we're done.
+                return
+            time.sleep(0.1)
```
fruch, line 1364: I think this sleep can be shorter

fruch:
@nyh seems like there are some places in dtest using `watch_log_for_alive`, we probably should replace them after this one gets merged

ccmlib/scylla_node.py
nyh: I'm happy that from tests that take 5 minutes, we went to "0.1 second is too long" :-)
What kind of sleep did you have in mind? Say 0.05? 0.01?

In any case, note that this sleep time is not cumulative: If we slept 0.1 seconds for the benefit of one node, by the time we go check the other node, the check is likely to pass immediately, without any sleep at all. So at worst, this sleep can add 0.1 seconds to the total cluster-creation time (well, not exactly... we also do this check after adding each node, so maybe 0.1 second per node).

nyh:
> @nyh seems like there are some places in dtest using `watch_log_for_alive`, we probably should replace them after this one gets merged

Do we run dtests also on Cassandra? If we do, my new function won't work there :-(

fruch:
> > @nyh seems like there are some places in dtest using `watch_log_for_alive`, we probably should replace them after this one gets merged
> 
> Do we run dtests also on Cassandra? If we do, my new function won't work there :-(

we could run with cassnadra, but we don't do that regularly

we could overload `watch_log_for_alive` to always go to `watch_rest_for_alive` if it's scylla.
and then test would still work with no change needed

nyh:
Pushed new version with that "_" thing.
Didn't change the sleep from 0.1 seconds - I can change it to anything, but I don't know what to pick...

nyh:
Pushed a new version, the REST API now also waits for nodes to know other nodes' tokens.
With this patch, when test is modified to disabled "wait for gossip to settle", we're down from 170 failures when I started to "just" 38.

nyh:
Ping.

fruch (approved):
LGTM

fruch:
@bhalevy, as whom refactored it a few time, care to also give it a look ?

ccmlib/scylla_node.py
```diff
@@ -1337,6 +1338,34 @@ def upgrade(self, upgrade_to_version):
...
+        instead of the log file and waits for the node to be really useable,
+        not just "UP" (see issue #461)
+        """
+        tofind = nodes if isinstance(nodes, list) else [nodes]
+        tofind = set([node.address() for node in tofind])
+        url_live = f"http://{self.address()}:10000/gossiper/endpoint/live"
+        url_joining = f"http://{self.address()}:10000/storage_service/nodes/joining"
+        endtime = time.time() + timeout
+        while time.time() < endtime:
+            live = {}
```
bhalevy, line 1357: live should be initialized as `live = set()`
otherwise there might be a minuscule chance that `live - set(response.json())` below will return the following error: `TypeError: unsupported operand type(s) for -: 'dict' and 'set'`

ccmlib/scylla_node.py
```diff
@@ -1359,9 +1360,19 @@ def watch_rest_for_alive(self, nodes, timeout=120):
...
                 # This node thinks that all given nodes are alive and not
-                # "joining", we're done.
+                # "joining", we're almost done, but still need to verify
+                # that the node knows the others' tokens.
+                check = tofind
+                tofind = set()
+                for n in check:
+                    response = requests.get(url=url_tokens+n)
+                    if response.text == '[]':
+                        tofind.add(n)
```
bhalevy, line 1375: I wonder how this would work with topology-related tests that currently track the log to inject some event during e.g. bootstrap or replace.
Although they can pass `wait_other_notice=False` and implement their own state tracking but they might break with this change.

ccmlib/scylla_node.py
```diff
@@ -311,6 +311,7 @@ def _start_scylla(self, args, marks, update_pid, wait_other_notice,
             for node, _ in marks:
                 t = timeout if timeout is not None else 120 if self.cluster.scylla_mode != 'debug' else 360
                 node.watch_rest_for_alive(self, timeout=t)
+                self.watch_rest_for_alive(node, timeout=t)
```
bhalevy, line 314: We could also call this once with the list of nodes after finishing this loop.

bhalevy: but it's not a must

fruch:
@nyh let's do a run of this change without the wait for gossip change, to see that we aren't breaking any tests with it.

And then we could merge it

mykaul:
@eliransin - do you need to review this to get this moving forward?

fruch:
> @nyh let's do a run of this change without the wait for gossip change, to see that we aren't breaking any tests with it.
> 
> And then we could merge it

running this with a full suite of dtest, to verify it's not causing regressions (or not too many of them) 
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/314/

fruch:
> @eliransin - do you need to review this to get this moving forward?

it's a change that affects all of the tests, I would want it tested first (@nyh has tested it so far only with the `'skip_wait_for_gossip_to_settle': 0`)

ccmlib/scylla_node.py
dkropachev: Maybe we could make use of ThreadPoolExecutor to wait all the nodes in parallel?

fruch: I think It's a bit overkill, and it won't make it happen any quicker, i.e. we are waiting for scylla here.

ccmlib/scylla_node.py
nyh: Oops, I forgot that {} wasn't an empty set :-(
Will fix this.

ccmlib/scylla_node.py
nyh: I don't know what kind of tests you mean, but we can see if any tests starts failing with this patch.
I think that if a certain test wants to capture some exact moment during the middle of the boot process, it can't ask ccm to wait until the boot is ove.

nyh:
pushed new version - rebased, and changed {} to set() as @bhalevy noticed.

nyh:
> > @eliransin - do you need to review this to get this moving forward?
> 
> it's a change that affects all of the tests, I would want it tested first (@nyh has tested it so far only with the `'skip_wait_for_gossip_to_settle': 0`)

I think you did this above, but I don't know how to read the results (if I understood correctly, we can't expect zero failures?).

fruch:
> > @nyh let's do a run of this change without the wait for gossip change, to see that we aren't breaking any tests with it.
> > And then we could merge it
> 
> running this with a full suite of dtest, to verify it's not causing regressions (or not too many of them) https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/314/

@nyh only a handful of failures

some are timing out the 2min timeout of currently introduced wait function.
I'm running now only the gating with it:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/315/

fruch:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/314/testReport/junit/bootstrap_test/TestBootstrap/Run_Dtest_Parallel_Cloud_Machines___FullDtest___full_split018___test_seeds_on_duty/

this test expect boot to failed, test should enable wait notice others

https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/314/testReport/replace_address_test/TestReplaceAddress/Run_Dtest_Parallel_Cloud_Machines___HeavyDtest___heavy_split001___test_replace_node_diff_ip_take_write_use_host_id_rbo_disabled_/

fails a bit like failures in debug runs... 

https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/314/testReport/compaction_additional_test/TestTimeWindowDataSegregation/Run_Dtest_Parallel_Cloud_Machines___FullDtest___full_split006___test_twcs_multiple_sstables_during_bootstrap/

missing data in node3, doesn't fail like that on master runs


failing as in master:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/314/testReport/repair_based_node_operations_test/TestRepairBasedNodeOperations/Run_Dtest_Parallel_Cloud_Machines___FullDtest___full_split019___test_ignore_dead_nodes_for_replace_option_2/

https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/314/testReport/compaction_additional_test/TestTimeWindowDataSegregation/Run_Dtest_Parallel_Cloud_Machines___FullDtest___full_split021___test_enable_disable_optimized_query_for_twcs/

https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/314/testReport/update_cluster_layout_tests/TestUpdateClusterLayout/Run_Dtest_Parallel_Cloud_Machines___FullDtest___full_split010___test_decommission_after_changing_node_ip_2/

https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/314/testReport/cqlsh_tests.cqlsh_copy_tests/TestCqlshCopy/Run_Dtest_Parallel_Cloud_Machines___FullDtest___full_split000___test_copy_to_with_child_process_crashing/

bhalevy:
I don't see how it would be ok to introduce regressions

nyh:
> I don't see how it would be ok to introduce regressions

It's not ok, but to be honest, I don't have the peace of mind to fix random buggy dtests while I keep getting requests to do other things at top priority, so I'm not going to fix this any time soon.

roydahan:
I started a new run with this: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/

The 3 test failures @fruch mentioned are directly related but looks quite easy to fix, so we can assign someone to fix.

However, this PR by itself won't improve runtime and possibly even degrade runtime.

nyh:
> I started a new run with this: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/
> 
> The 3 test failures @fruch mentioned are directly related but looks quite easy to fix, so we can assign someone to fix.

Thanks.
 
> However, this PR by itself won't improve runtime and possibly even degrade runtime.

The idea is to have a dtest PR which runs Scylla without huge sleeps. The branch is https://github.com/nyh/scylla-dtest/commits/quick-boot-2. There is a one-line patch https://github.com/nyh/scylla-dtest/commit/0b23742b4622c9b6e18f4613722ce433640cf41c plus one fix for one test's bug - https://github.com/nyh/scylla-dtest/commit/0b23742b4622c9b6e18f4613722ce433640cf41c. There will need to be more patches because a few dozen tests still fail (not as many as before the ccm patches, and but still, a non-zero list).

The next step can be to either make the dtest patch an optional parameter (so developers can use it), but @fruch and @eliransin suggested that it can be the default, and just add a tag for tests that can't work in this "sleepless" mode and only they will be run in the slow mode.

fruch:
run with the fixes of https://github.com/scylladb/scylla-dtest/pull/3313 in:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/317

fruch:
one more test that need to be addressed:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/317/testReport/throughput_limits_test/TestStreamingLimitThroughput/Run_Dtest_Parallel_Cloud_Machines___FullDtest___full_split005___test_limit_rbno_bootstrap_throughput/

kbr-scylla:
@nyh This is incorrect. It may fix some problems but introduces others

Consider the following:
- 2 nodes A, B
- kill node A
- start node A with `wait_other_notice=True`

The `start` call will often return quickly, but not because node B noticed that A restarted
but because node B didn't even notice that A was killed yet!

That's why the old check used the log markers --- the log markers ensure that we don't depend on obsolete state for `wait_other_notice`. That is, other nodes really have to notice this new node restart!

We're actually now observing this scenario trying to unflake one dtest with @temichus (https://github.com/scylladb/scylla-dtest/pull/3991)...

temichus:
> @nyh This is incorrect. It may fix some problems but introduces others
> 
> Consider the following:
> 
> * 2 nodes A, B
> * kill node A
> * start node A with `wait_other_notice=True`
> 
> The `start` call will often return quickly, but not because node B noticed that A restarted but because node B didn't even notice that A was killed yet!
> 
> That's why the old check used the log markers --- the log markers ensure that we don't depend on obsolete state for `wait_other_notice`. That is, other nodes really have to notice this new node restart!
> 
> We're actually now observing this scenario trying to unflake one dtest with @temichus ([scylladb/scylla-dtest#3991](https://github.com/scylladb/scylla-dtest/pull/3991))...

@kbr-scylla 

I've created PR where the issue can be reproduced https://github.com/scylladb/scylla-dtest/pull/4042/files
approx with 500 runs, we should get at least 1 error.

bhalevy:
> @nyh This is incorrect. It may fix some problems but introduces others
> 
> Consider the following:
> 
> * 2 nodes A, B
> * kill node A
> * start node A with `wait_other_notice=True`
> 
> The `start` call will often return quickly, but not because node B noticed that A restarted but because node B didn't even notice that A was killed yet!

You can stop A with wait_other_notice=True to ensure node B notices that A went down, before restarting it.

> 
> That's why the old check used the log markers --- the log markers ensure that we don't depend on obsolete state for `wait_other_notice`. That is, other nodes really have to notice this new node restart!
> 
> We're actually now observing this scenario trying to unflake one dtest with @temichus ([scylladb/scylla-dtest#3991](https://github.com/scylladb/scylla-dtest/pull/3991))...

kbr-scylla:
> You can stop A with wait_other_notice=True to ensure node B notices that A went down, before restarting it.

Which is only a workaround for the problem, the implementation of `wait_other_notice` is still incorrect

bhalevy:
> > You can stop A with wait_other_notice=True to ensure node B notices that A went down, before restarting it.
> 
> Which is only a workaround for the problem, the implementation of `wait_other_notice` is still incorrect

Then it needs to be fixed.  Issue nr?

kbr-scylla:
I opened https://github.com/scylladb/scylla-ccm/issues/563

Issue #463, opened 2023-05-30
SCYLLA_MANAGER_PACKAGE can only be URL


fruch:
* we should be able to use also local file if needed
* we should be able to download from s3 when possible same as we do for scylla unified packages

PR #464, opened 2023-06-01
scylla-manager: make `--scylla-manager-package` option more robust

fruch:
* now it would first try to download from s3 via boto api (faster)
* local tar.gz file of manager can be used (no need to have it on s3/http)
* hash to identify the specific version now based on ETag and not on the url path itself.

Fixes: #463

### Testing
- [ ] - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/manager-dtest/8/console

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -1,5 +1,3 @@
-
-
 import logging
 import random
 import time
@@ -16,17 +14,14 @@ import sys
 import glob
 import urllib
 
-import hashlib
 import requests
 import yaml
-
-
 import packaging.version
 
 from ccmlib.common import (
     ArgumentError, CCMError, get_default_path, rmdirs, validate_install_dir, get_scylla_version, aws_bucket_ls,
     DOWNLOAD_IN_PROGRESS_FILE, print_if_standalone, LockFile)
-from ccmlib.utils.download import download_file, download_version_from_s3
+from ccmlib.utils.download import download_file, download_version_from_s3, get_url_hash
 from ccmlib.utils.version import parse_version
 
 GIT_REPO = "http://github.com/scylladb/scylla.git"
@@ -407,7 +402,7 @@ def download_packages(version_dir, packages, s3_url, scylla_product, version, ve
     return package_version, packages
 
 
-def setup_scylla_manager(scylla_manager_package=None):
+def setup_scylla_manager(scylla_manager_package=None, verbose=False):
 
     """
     download and cache scylla-manager RPMs,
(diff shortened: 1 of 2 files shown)


ccmlib/scylla_repository.py
```diff
@@ -414,25 +409,30 @@ def download_packages(version_dir, packages, s3_url, scylla_product, version, ve
...
             os.makedirs(manager_install_dir)
-            tar_data = requests.get(scylla_manager_package, stream=True)
             destination_file = os.path.join(manager_install_dir, "manager.tar.gz")
-            with open(destination_file, mode="wb") as f:
-                f.write(tar_data.raw.read())
+
+            if os.path.exists(scylla_manager_package) and scylla_manager_package.endswith('.tar.gz'):
+                destination_file = scylla_manager_package
+            elif is_valid(scylla_manager_package):
+                _, target = tempfile.mkstemp(suffix=".tar.gz", prefix="ccm-")
```
tchaikov, line 431: what is `target` for? presumably, you wanted to use it for the path for the temporary file for the compressed scylla manager package. but it turns out you actually download the file to `destination_file`.

fruch: you are correct it's a copy paste, and this line should replace 
```
destination_file = os.path.join(manager_install_dir, "manager.tar.gz")
```

name should be temporary

ccmlib/utils/download.py
```diff
@@ -158,3 +159,24 @@ def download_version_from_s3(url: str, target_path: str, verbose=False):
...
+
+    # first try is on s3
+    parts = urllib.parse.urlparse(url)
+    _, bucket_name, download_path = parts.path.split('/', maxsplit=2)
+    s3_client = Session().client('s3', config=botocore.client.Config(signature_version=botocore.UNSIGNED))
+
+    try:
+        metadata = s3_client.head_object(Bucket=bucket_name, Key=download_path)
+        return metadata.get('ETag')[1:-1]
+    except botocore.client.ClientError as ex:
```
tchaikov, line 184: nit, remove `as ex`, as we don't care about the value of the exception.

ccmlib/utils/download.py
```diff
@@ -148,6 +148,10 @@ def download_version_from_s3(url: str, target_path: str, verbose=False):
             logging.warning(f"url: '{url}' wasn't found on S3")
             logging.warning(f"download might be very slow")
             return None
+        elif 'Forbidden' in str(ex):
```
tchaikov, line 151: instead of inspecting the stringified representation of the exception, i'd suggest follow [AWS's document](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html#parsing-error-responses-and-catching-exceptions-from-aws-services), and compare the error code with the ones defined by the [document](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList) like:
```py
except botocore.client.ClientError as ex:
  error_code = error.response['Error']['Code']
  if error_code in ('NoSuchBucket', 'NoSuchKey'):
    logging.warning(f"url: '{url}' wasn't found on S3")
    logging.warning(f"download might be very slow")
    return None
  elif error_code == "AccessDenied"
    logging.warning(f"url: '{url}' Forbidden (403) on S3")
    logging.warning(f"download might be very slow")
    return None
  else:
    raise
```

ccmlib/utils/download.py
fruch: fixed

ccmlib/scylla_repository.py
```diff
@@ -414,25 +409,30 @@ def download_packages(version_dir, packages, s3_url, scylla_product, version, ve
...
+                res = download_version_from_s3(url=scylla_manager_package, target_path=destination_file, verbose=verbose)
+                if not res:
+                    download_file(url=scylla_manager_package, target_path=destination_file, verbose=verbose)
```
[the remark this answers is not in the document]
fruch: No, it would fail inside the last option, and would raise exception.

ccmlib/utils/download.py
fruch: replaced it

ccmlib/scylla_repository.py
fruch: move the temp file into the proper place

ccmlib/utils/download.py
```diff
@@ -143,10 +144,15 @@ def download_version_from_s3(url: str, target_path: str, verbose=False):
     try:
         metadata = s3_client.head_object(Bucket=bucket_name, Key=download_path)
     except botocore.client.ClientError as ex:
-        if 'Not Found' in str(ex):
+        error_message = ex.response['Error']['Message']
```
tchaikov, line 147: could you please revisit my previous comment, and review the document. for a sample of how the error is examined, see https://github.com/aws/aws-cli/blob/855cd0ddb3c7fd47cab02a390281f29cc9bc744c/awscli/customizations/s3uploader.py#L123-L127

i think the "Message" field is supposed to be consumed by human being, not to be checked by an application.

ccmlib/scylla_repository.py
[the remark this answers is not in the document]
tchaikov: what if `is_valid(scylla_manager_package)` returns `False`?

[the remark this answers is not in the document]
fruch: then it won't get to this code part

Issue #465, opened 2023-06-05
"waiting for scylla process to be running" log spam


kbr-scylla:
Huge spam in the logs when I run dtest locally:
```
15:50:51,771 ccm             DEBUG cluster.py     :709  | node3: Starting scylla-jmx: args=['/home/kbraun/.dtest/dtest-8v67tmp7/test/node3/bin/symlinks/scylla-jmx', '-Dapiaddress=127.0.41.3', '-Djavax.management.builder.initial=com.scylladb.jmx.utils.API
Builder', '-Djava.rmi.server.hostname=127.0.41.3', '-Dcom.sun.management.jmxremote', '-Dcom.sun.management.jmxremote.host=127.0.41.3', '-Dcom.sun.management.jmxremote.port=7199', '-Dcom.sun.management.jmxremote.rmi.port=7199', '-Dcom.sun.management.jmxre
mote.local.only=false', '-Xmx256m', '-XX:+UseSerialGC', '-Dcom.sun.management.jmxremote.authenticate=false', '-Dcom.sun.management.jmxremote.ssl=false', '-jar', '/home/kbraun/.dtest/dtest-8v67tmp7/test/node3/bin/scylla-jmx-1.0.jar']
Waiting for scylla process to be running (0.000007 secs)                                                                                                                                                                                                      
Waiting for scylla process to be running (0.010086 secs)                                                                                                                                                                                                      
Waiting for scylla process to be running (0.020234 secs)                                                                                                                                                                                                      
Waiting for scylla process to be running (0.030353 secs)                                                                       
Waiting for scylla process to be running (0.040477 secs)                                                                       
Waiting for scylla process to be running (0.050609 secs)                                                                                                                                                                                                      
Waiting for scylla process to be running (0.060743 secs)                                                                                                                                                                                                      
Waiting for scylla process to be running (0.070882 secs)                                                                                                                                                                                                      
Waiting for scylla process to be running (0.081018 secs)                                                                       
Waiting for scylla process to be running (0.091158 secs)                                                                                                                                                                                                      
Waiting for scylla process to be running (0.101272 secs)                                                                                                                                                                                                      
Waiting for scylla process to be running (0.111388 secs)                                                                                                                                                                                                      
Waiting for scylla process to be running (0.121505 secs)       
Waiting for scylla process to be running (0.131624 secs)       
Waiting for scylla process to be running (0.141749 secs)       
...
Waiting for scylla process to be running (9.469546 secs)
[...]

PR #466, opened 2023-06-05
ccmlib/common: remove prints from `wait_for`

fruch:
remove text parameter from `wait_for` it was spamming the outout for now real reason. we should be print to stdout during tests. for debug propuses we
should be using `logging.debug`, which in
this case isn't needed.

Fixes: #465

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -512,7 +512,7 @@ def scylla_extract_install_dir_and_mode(install_dir):
     return install_dir, scylla_mode
 
 
-def wait_for(func: Callable, timeout: int, first: float = 0.0, step: float = 1.0, text: str = ""):
+def wait_for(func: Callable, timeout: int, first: float = 0.0, step: float = 1.0):
     """
     Wait until func() evaluates to True.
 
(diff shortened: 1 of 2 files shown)


fruch:
> This will break a bunch of dtests tho:
> 
> ```
> [kbraun@fedora scylla-dtest]$ git grep "wait_for.*text="
> alternator_tests.py:        wait_for(self.is_table_schema_synced, timeout=30, text='Waiting until table schema is updated',
> alternator_tests.py:        wait_for(self.is_table_schema_synced, timeout=30, text='Waiting until table schema is updated',
> alternator_tests.py:        wait_for(self.is_table_schema_synced, timeout=30, text='Waiting until table schema is updated',
> alternator_tests.py:        wait_for(self.is_table_schema_synced, timeout=30, text='Waiting until table schema is updated',
> alternator_tests.py:        wait_for(self.is_table_schema_synced, timeout=30, step=3, text='Waiting until table schema is updated',
> cdc_test.py:        wait_for(check_metadata, timeout=60, text='Waiting until metadata is updated')
> dtest_class.py:def forever_wait_for(func, step=1, text=None, **kwargs):
> dtest_class.py:def wait_for(func, step=1, text=None, timeout=None, throw_exc=True, **kwargs):
> dtest_scylla_manager.py:            is_status_reached = wait_for(func=self.is_status_in_list, step=step, text=text, list_status=list_status,
> ics_compaction_test.py:        wait_for(func=is_compaction_executed, text=str(is_compaction_executed),
> internode_ssl_test.py:                wait_for(is_port_used, text=f'Waiting port {port} is used', step=0.5, timeout=2,
> manager_backup_tests.py:        is_status_reached = wait_for(func=self._is_node_at_status, step=step, text=text, timeout=timeout, throw_exc=True,
> manager_general_tests.py:        wait_for(func=has_removed_node_reached_dn, step=step, text="Node status has yet to reach DN", timeout=timeout,
> native_transport_ssl_test.py:                wait_for(is_port_used, text=f'Waiting port {port} is used', step=0.5, timeout=2,
> tools/cdc_utils.py:        wait_for(check_metadata, timeout=60, text='Waiting until metadata is updated')
> ```

it's a function with the same name, but dtest has its own implementation (even that is quite similar)

PR #470, opened 2023-06-17
Use advisory locks for lockfiles

Lorak-mmk:
Up until now the locking mechanism used to prevent concurrent downloads was just the existence of a lock file. This implementation had several problems:
 - Locking procedure was unsafe - it was possible for more than one process to enter critical section at the same time. This was "fixed" by putting random sleep in front of it, but that is not a proper synchronisation mechanism.
 - Download procedure was clearing whole directory at one point. This caused lockfile to be removed too, so it was possible for another process to enter the critical section at this point.
 - If the process was killed unexpectedly, the lockfile remained in place. If you try to start the download again it will wait for an hour for the lockfile to be deleted. This is frustrating when running e.g. tests for some driver, as they just hang and you have no idea why.

This PR replaces that with advisory locks and fixes all those problems. Now the download procedure will just restart if it was aborted previously. It's no longer possible for 2 processes to enter critical section, so the (0-5) seconds sleep is removed. Download procedure no longer removes the lockfile.

There is one thing I don't quite understand: why do we use lockfile for Scylla downloads but not for Cassandra downloads? Is there something that makes Cassandra downloads safe in this regard, or is that just an oversight?

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -3,6 +3,7 @@
 #
 
 import fnmatch
+import fcntl
 import os
 import platform
 import re
@@ -16,7 +17,7 @@ import tempfile
 import logging
 import pathlib
 from itertools import zip_longest
-from typing import Callable, Optional
+from typing import Callable, Optional, TextIO, Union
 
 import yaml
 from boto3.session import Session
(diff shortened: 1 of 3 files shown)


fruch:
Regarding Cassandra, I don't think it's safe either, but we are not using it that much.

I dtest we have used it in 5 tests, and the Cassandra version we use comes pre cached in the docker image we use.

ccmlib/common.py
```diff
@@ -65,6 +66,118 @@ class UnavailableSocketError(CCMError):
...
+        return self.read_contents()[1]
+
+    def __enter__(self):
+            (success, blocking_pid_opt) = self.try_lock()
+            if success:
+                return self
+            print(f"Another download running into '{os.path.dirname(self.filename)}', "
+                  f"by process '{blocking_pid_opt}'. Waiting for parallel downloading finished")
+
+            if not wait_for(func=lambda: self.try_lock()[0], timeout=3600):
```
fruch, line 173: Maybe we can pick a smaller timeout

And print some explanations about what a user might need to do, to get rid of the lock, if needed ?

fruch (approved):
LGTM

ccmlib/common.py
```diff
@@ -65,6 +66,118 @@ class UnavailableSocketError(CCMError):
...
+        # 'a+' is the only mode that satisfies all of this.
+        self.filename = filename
+        self.file = open(filename, 'a+')
+        self.locked = False
+
+    def try_lock(self, new_status: str = None) -> (bool, Optional[int]):
+        """Tries to take a lock. If it fails, it return PID of the process
+        that currently owns this lock.
+        """
+        assert not self.locked
```
tchaikov, line 117: i am confused. if the lock is already owned by some process, `self.locked` is True. so in this case, `(False, blocking_pid)` should be returned. but in your implementation, `AssertionError` is raised instead.

ccmlib/common.py
```diff
@@ -65,6 +66,118 @@ class UnavailableSocketError(CCMError):
...
+    filename: str | bytes | os.PathLike
+        Path to a lockfile - used for logging
+    file: TextIO
+        File handle which will be used to take a lock.
+    locked: bool
+        Current status of the lock.
+    """
+    filename: str | bytes | os.PathLike
+    file: TextIO
+    locked: bool
```
tchaikov, line 101: nit, better off adding `_` prefix to these member variables. as it seems they are not part of the public interface of `LockFile`. if they are, then we might want to provide read-only accessors or functions for accessing them.

ccmlib/common.py
Lorak-mmk: `self.locked` is True if it is locked by this process.

Lorak-mmk: This assert will throw if a user tries to take already owned lock, so e.g. a code like this:
```
l = LockFile('lockfile')
(success, pid) = l.try_lock()
assert success # Let's assume there are no other instances
(success, pid) = l.try_lock() # This would throw
```
I can add a comment that the lock is not reentrant.

ccmlib/common.py
Lorak-mmk: What value would you propose? Current one is the same as it was before, I just copied it.

ccmlib/common.py
tchaikov: > I can add a comment that the lock is not reentrant.

that'd be helpful. i see. thank you for the detailed explanation. this makes more sense now.

ccmlib/common.py
```diff
@@ -65,6 +66,118 @@ class UnavailableSocketError(CCMError):
...
+            (blocking_pid, old_status) = file_data.split(';', 1)
+            blocking_pid = int(blocking_pid)
+            return (blocking_pid, old_status)
+        except:
+            return (None, None)
+    
+    def read_status(self) -> Optional[str]:
+        return self.read_contents()[1]
+
+    def __enter__(self):
```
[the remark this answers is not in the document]
fruch: We still need support up to python 3.8

Some of the usages can't jump to python 3.11

Regardless we can return self, without stating it in type annotations

fruch:
Anyhow I would like to see it running dtest upgrade test that are the cause of introducing this lock, since they are downloading the same versions in multiple test in the same run.

@juliayakovlev can you help @Lorak-mmk getting those run in Jenkins with this ccm ?

ccmlib/common.py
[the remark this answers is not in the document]
Lorak-mmk: Python driver supports down to Python 3.6 and it uses CCM for testing.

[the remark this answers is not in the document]
Lorak-mmk: Added explicit return type - but I used `'LockFile'` instead of `Self` in order to support older Python versions.

ccmlib/common.py
Lorak-mmk: Added the note with an example. Also changed description of the `_locked` member.

Lorak-mmk:
Addressed review comments - apart from shortening the timeout value as I'm not sure what value would be good here.

tchaikov (approved):
lgtm

fruch:
Running it on:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/312/

fruch:
> Running it on: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/312/

run pass with success

PR #471, opened 2023-06-19
scylla_node: add run_scylla_sstable()

denesb:
Convenience method allowing the invocation of the scylla-sstable tool, built into the scylla sstable. The method uses the same mechanism for selecting which sstable to work on, as Node.run_sstable2json(). But other than that, there are notable differences, run_scylla_sstable():
* Allows for accessing all scylla-sstable operations, not just dump-data (the scylla-sstable equivalent to sstable2json).
* Allows direct pass-through of command-line arguments to the underlying scylla-sstable invokation. This ensures maximum forward-compatibility, without the need to patch this method each time the underlying code's capabilities change.
* Has a batch mode, which allows for passing all sstables at once to scylla-sstable (which accepts 1 or more sstable args).
* It looses the `enumerate_keys` parameter.
* Returns the (stdout, stderr) for each underlying invocation, instead of writing the output to a file.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1348,6 +1348,45 @@ class ScyllaNode(Node):
                 return str(candidate)
         raise ValueError(f"gnutls.config wasn't found in any path: {candidates}")
 
+    def run_scylla_sstable(self, command, additional_args=None, keyspace=None, datafiles=None, column_families=None, batch=False):
+        """Invoke scylla-sstable, with the specified command (operation) and additional_args.
+
+        For more information about scylla-sstable, see https://docs.scylladb.com/stable/operating-scylla/admin-tools/scylla-sstable.html.
+
+        Params:
+        * command - The scylla-sstable command (operation) to run.
+        * additional_args - Additional command-line arguments to pass to scylla-sstable, this should be a list of strings.
+        * keyspace - Restrict the operation to sstables of this keyspace.
+        * datafiles - Restrict the operation to the specified sstables (Data components).
+        * column_families - Restrict the operation to sstables of these column_families. Must contain exactly one column family when datafiles is used.
+        * batch - If True, all sstables will be passed in a single batch. If False, sstables will be passed one at a time.
+            Batch-mode can be only used if column_families contains a single item.
+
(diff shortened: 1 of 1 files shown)


ccmlib/scylla_node.py
```diff
@@ -1348,6 +1348,45 @@ def gnutls_config_file(self):
...
+        if additional_args is None:
+            additional_args = []
+        scylla_path = common.join_bin(self.get_path(), BIN_DIR, 'scylla')
+        env = self.get_env()
+        sstables = super()._Node__gather_sstables(datafiles, keyspace, column_families)
+        ret = {}
+
+        def do_invoke(sstables):
+            common_args = [scylla_path, "sstable", command] + additional_args
+            res = subprocess.run(common_args + sstables, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
```
fruch, line 1378: Add

text=True

And you can remove the decode calls from below

ccmlib/scylla_node.py
```diff
@@ -1348,6 +1348,45 @@ def gnutls_config_file(self):
...
+
+        Returns: map: {sstable: (stdout, stderr)} of all invokations. When batch == True, a single entry will be present, with empty key.
+
+        Raises: subprocess.CalledProcessError if scylla-sstable returns a non-zero exit code.
+        """
+        if additional_args is None:
+            additional_args = []
+        scylla_path = common.join_bin(self.get_path(), BIN_DIR, 'scylla')
+        env = self.get_env()
+        sstables = super()._Node__gather_sstables(datafiles, keyspace, column_families)
```
fruch, line 1373: Why use super and not self ?

ccmlib/scylla_node.py
denesb: Neat, fixed in v2.

ccmlib/scylla_node.py
denesb: It took me a while to figure out how to call the method with `__` prefix. I guess `super()` is a leftover of things I tried. Removed in v2.

denesb:
V2:
* Add `text=True` to `subprocess.run()` args.
* Use `self` instead of `super()`.

PR #472, opened 2023-07-03
Add experimental Nix Flake support

Lorak-mmk:
Nix is, citing their GH repo,  'a powerful package manager for Linux and other Unix systems that makes package management reliable and reproducible.'
See here for a more detailed description: https://serokell.io/blog/what-is-nix

We want to use Nix in drivers (starting with Java driver) to create reproducible, easy to use development environments, and later to extend this to CI and release process.
In order to do that, we need to create Nix package for CCM, as it's used by many drivers in integration tests. This PR accomplishes experimental version of it.

PR introduces Nix flake that can be used to add CCM as a dependency somewhere or to quickly enter a development environment where you can run and test CCM (with caveats due to Scylla using absolute paths when searching for Java - see commit messages / README / flake.nix).
There is also a CI pipeline running tests in Nix environment - marked with `continue-on-error` in order to not affect CCM's development process if something breaks in Nix. When we are sure everything is correct and stable, this parameter can be removed.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,58 @@
+name: CCM Integration tests (using Nix)
+
+on:
+  push:
+    branches:
+     - master
+     - next*
+
+  pull_request:
+    branches:
+     - next*
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+
+    steps:
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v20
+      with:
+        github_access_token: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Cache binary versions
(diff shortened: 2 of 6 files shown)


fruch (approved):
LGTM
as long as driver team is taking ownership on it (and it works)

avelanarius:
> LGTM as long as driver team is taking ownership on it (and it works)

Yes, we are taking ownership on it. We'll experiment with it as a base of more approachable dev environments of all drivers (as each driver is written in a different language and requires different toolchain).

avelanarius (approved):
LGTM. Tested it on my computer, starting Scylla 5.1.0 and 2022.1.7 (different relocatable versions) and it all now works correctly.

@fruch I think you can merge it now.

This will make it easier to use CCM without having to worry about having correct Java/Python/etc. on the system: `nix shell github:scylladb/scylla-ccm` and everything will work after that single command. The source itself is very small (57 lines of `flake.nix`) so maintaining it won't be a huge burden for me and @Lorak-mmk.

PR #474, opened 2023-07-06
ccmlib/scylla_repository: introduce get_manger_* functions

fruch:
those helper function could be user to retrive scylla manager relocatable packages, that we need in dtest for testing manager, those comes in two flavors:
* one to retrive lastet version of of specific branches
* one to retrive specific released versions

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -1,9 +1,3 @@
-import logging
-import random
-import time
-from pathlib import Path
-from typing import NamedTuple
-
 import os
 import tarfile
 import tempfile
@@ -13,6 +7,11 @@ import re
 import sys
 import glob
 import urllib
+import logging
+import random
+import time
+from pathlib import Path
+from typing import NamedTuple, Literal
 
 import requests
 import yaml
@@ -566,3 +565,40 @@ def run_scylla_unified_install_script(install_dir, target_dir, package_version):
     run('''{0}/install.sh --prefix {1} --nonroot{2}'''.format(
         install_dir, target_dir, install_opt), cwd=install_dir)
     run(f'''ln -s {install_dir}/scylla/conf conf''', cwd=target_dir)
+
+
+Architecture = Literal['x86_64', 'aarch64']
+BASE_DOWNLOADS_URL = 'https://s3.amazonaws.com/downloads.scylladb.com'
+
+
+def get_manager_latest_reloc_url(branch: str = "master", architecture: Architecture = None) -> str:
+    """
+    get the latest manager relocatable version of a specific branch
+    """
+    architecture = architecture or os.environ.get('SCYLLA_ARCH', 'x86_64')
+
+    url = f"{BASE_DOWNLOADS_URL}/manager/relocatable/unstable/{branch}/"
+    # filter only specific architecture
+    all_packages = reversed(aws_bucket_ls(url))
+    latest_package = next(filter(lambda tar: architecture in tar, all_packages))
+
+    # return latest
+    return f'{BASE_DOWNLOADS_URL}/{latest_package}'
+
+
+def get_manager_release_url(version: str = '', architecture: Architecture = None) -> str:
+    """
+    get latest official relocatable of manager releases of specific versions i.e. '3.1' or '3.1.1'
+    only works from release 3.1 and up
+    when version is empty string, won't return latest release (by date, so can be from older branch)
+    """
+    architecture = architecture or os.environ.get('SCYLLA_ARCH', 'x86_64')
+
+    url = f"{BASE_DOWNLOADS_URL}/downloads/scylla-manager/relocatable"
+
+    version_regex = re.compile('scylla-manager_(.*)-0')
+    # filter only specific architecture and version
+    all_packages = reversed(aws_bucket_ls(url))
+    latest_package = next(filter(lambda tar: architecture in tar and version in version_regex.search(tar)[0], all_packages))
+
+    return f'{BASE_DOWNLOADS_URL}/downloads/scylla-manager/relocatable/{latest_package}'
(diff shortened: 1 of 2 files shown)


dkropachev (approved):

ccmlib/scylla_repository.py
```diff
@@ -567,3 +562,37 @@ def run_scylla_unified_install_script(install_dir, target_dir, package_version):
     run('''{0}/install.sh --prefix {1} --nonroot{2}'''.format(
         install_dir, target_dir, install_opt), cwd=install_dir)
     run(f'''ln -s {install_dir}/scylla/conf conf''', cwd=target_dir)
+
+
+Architecture = Literal['x86_64', 'aarch64']
+
+
+def get_manager_latest_reloc_url(branch: str = "master", architecture: Architecture = None) -> str:
```
tchaikov, line 570: if we want to enforce type annotations, we might need to, well, enforce it. like

```py
def get_manager_latest_reloc_url(branch: str = "master", architecture: Optional[Architecture] = None) -> str:
```

ccmlib/scylla_repository.py
```diff
@@ -567,3 +562,37 @@ def run_scylla_unified_install_script(install_dir, target_dir, package_version):
...
+    get the latest manager relocatable version of a specific branch
+    """
+    architecture = architecture or os.environ.get('SCYLLA_ARCH', 'x86_64')
+
+    url = f"https://s3.amazonaws.com/downloads.scylladb.com/manager/relocatable/unstable/{branch}/"
+    # filter only specific architecture
+    all_packages = [tar for tar in aws_bucket_ls(url) if architecture in tar]
+
+    # return latest
+    return f'https://s3.amazonaws.com/downloads.scylladb.com/{all_packages[-1]}'
```
tchaikov, line 581: nit, no need to materialize the filtered package list for picking the last one. could put something like
```py
all_packages = reversed(aws_bucket_ls(url))
return next(filter(lambda tar: architecture in tar, all_packages))
```

ccmlib/scylla_repository.py
```diff
@@ -567,3 +562,37 @@ def run_scylla_unified_install_script(install_dir, target_dir, package_version):
...
+    """
+    architecture = architecture or os.environ.get('SCYLLA_ARCH', 'x86_64')
+
+    url = "https://s3.amazonaws.com/downloads.scylladb.com/downloads/scylla-manager/relocatable"
+
+    version_regex = re.compile('scylla-manager_(.*)-0')
+    # filter only specific architecture and version
+    all_packages = [tar for tar in aws_bucket_ls(url) if architecture in tar and version in version_regex.search(tar)[0]]
+
+    return f'https://s3.amazonaws.com/downloads.scylladb.com/downloads/scylla-manager/relocatable/{all_packages[-1]}'
```
tchaikov, line 598: ditto. i'd avoid repeating the bucket name and the base url if possible.

ccmlib/scylla_repository.py
```diff
@@ -15,12 +7,15 @@
 import sys
 import glob
 import urllib
-
 import hashlib
+import logging
+import random
+import time
```
tchaikov, line 13: i assume that you want to reorder the `import` statements. but in what order? in alphabetical order ? if yes, then there are still some other `import`s are not ordered.

fruch: The grouping of them is more important to me than the alphabetic order or any other order.

I.e. builtin first, third party, in package import

ccmlib/scylla_repository.py
fruch: It's mostly for documentation now, not going to enforce it with tools.

ccmlib/scylla_repository.py
```diff
@@ -567,3 +562,37 @@ def run_scylla_unified_install_script(install_dir, target_dir, package_version):
...
+
+    url = f"https://s3.amazonaws.com/downloads.scylladb.com/manager/relocatable/unstable/{branch}/"
+    # filter only specific architecture
+    all_packages = [tar for tar in aws_bucket_ls(url) if architecture in tar]
+
+    # return latest
+    return f'https://s3.amazonaws.com/downloads.scylladb.com/{all_packages[-1]}'
+
+
+def get_manager_release_url(version: str = '', architecture: Architecture = None) -> str:
```
tchaikov, line 584: we can also consolidate these two functions into a single one, like
```py
def get_manager_url(version: Optional[str] = None,
                    architecture: Architecture = None) -> str:
    base_url = 'https://s3.amazonaws.com/downloads.scylladb.com'
url = f'{base_url}/manager/relocatable/unstable/{branch}/'
    all_packages = reversed(aws_bucket_ls(url))
    if architecture is None:
        architecture = os.environ.get('SCYLLA_ARCH', 'x86_64')
    matched_packages = filter(lambda tar: architecture in tar, all_packages)

    if version is not None:
        version = version.replace('.', '\.')
        version_regex = re.compile(f'scylla-manager_{version}.*-0')
        matched_packages = (tar for tar in matched_packages if version_regex.search(tar))

    latest_package = next(matched_packages)
    return f'{base_url}/{latest_package}'
```

ccmlib/scylla_repository.py
tchaikov: yes, if we want to use for document, i think we need to get it right from the very beginning.

ccmlib/scylla_repository.py
tchaikov: okay, i see. makes more sense now.

ccmlib/scylla_repository.py
fruch: the url isn't the same in both, the release are in one place, and the development version are if a different location.
I think it's o.k. to have two different function

ccmlib/scylla_repository.py
fruch: `aws_bucket_ls` already materialize the whole list
filtering the first few might be quicker

PR #475, opened 2023-07-13
Fix Python 3.8 compatibility

Lorak-mmk:
Recent change introduced typing that is incompatible with Python < 3.9. There are other Scylla projects relying on CCM (e.g. most drivers) and for some of them this change caused CI failures (as of now - for Rust driver and Python driver).

This PR fixes the problem by making this typing compatible with older Python versions.
It also adds CI that runs on older Python. I wanted it to be 3.6, because that's the oldest version supported by Python driver, but it seems this version can't be installed by `setup-python@v4` action, so I bumped it to 3.8

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -14,12 +14,16 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.11"]
+
     steps:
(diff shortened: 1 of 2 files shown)


fruch:
@Lorak-mmk we are not running python-driver integration tests with all versions of python (only the unittests)

I'm o.k. with doing 3.8 and up.

but we should upgrade CI where it's relevent to to 3.9 at least

fruch (approved):

Issue #477, opened 2023-07-24
self.watch_rest_for_alive(other_node) fails en-mass during bootstrap in debug mode


bhalevy:
See for example https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-daily-debug/245/testReport/rebuild_test/TestRebuild/Run_Dtest_Parallel_Cloud_Machines___FullDtest___full_split094___test_rebuild_many_keyspaces/
```
self = <rebuild_test.TestRebuild object at 0x7f1b7ed0bf50>

    def test_rebuild_many_keyspaces(self):
        """
        Test rebuilding many keyspaces in same dc works as expected.
        """
    
        num_keys = 100
        num_keyspaces = 50
        num_tables = 2
    
        cluster = self.cluster
        cluster.set_configuration_options(values={'endpoint_snitch': 'GossipingPropertyFileSnitch'})
        node1 = self.add_node(1)
    
        # start node in dc1
        node1.start(wait_for_binary_proto=True)
    
        # populate data in dc1
        session = self.patient_exclusive_cql_connection(node1)
        dc = 'dc1'
        logger.debug(f"Creating {num_keyspaces} keyspaces")
        keyspaces = ['ks_{:04d}'.format(i) for i in range(0, num_keyspaces)]
        for ks in keyspaces:
            create_ks(session, ks, {dc: 1})
    
        logger.debug(f"Creating {num_tables} table(s) in each ks")
        tables = ['cf_{:04d}'.format(i) for i in range(0, num_tables)]
        for ks in keyspaces:
            for cf in tables:
                create_c1c2_table(session, cf=f'{ks}.{cf}', debug_query=False)
                insert_c1c2(session, n=num_keys, ks=ks, cf=cf, consistency=ConsistencyLevel.ALL)
    
        keys = [i for i in range(0, num_keys)]
        self._check_data(session, keyspaces, tables, keys)
        session.shutdown()
    
        logger.debug("Bootstrapping node2 with {auto_bootstrap: false}")
        node2 = self.add_node(2)
>       node2.start(wait_other_notice=True, wait_for_binary_proto=True)

rebuild_test.py:315: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../scylla/.local/lib/python3.11/site-packages/ccmlib/scylla_node.py:669: in start
[...]

PR #479, opened 2023-07-24
scylla_node: disable logging for `watch_rest_for_alive`

fruch:
disable `urllib3.connectionpool` logging for this function, it doing lots of retries and we don't need to see each request that is being sent out in the log

Ref: #477

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -13,6 +13,7 @@ import time
 import threading
 from pathlib import Path
 from collections import OrderedDict
+import logging
 
 import psutil
 import yaml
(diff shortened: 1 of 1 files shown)


ccmlib/scylla_node.py
```diff
@@ -1347,36 +1348,40 @@ def watch_rest_for_alive(self, nodes, timeout=120):
...
-                tofind = set()
-                for n in check:
-                    response = requests.get(url=url_tokens+n)
-                    if response.text == '[]':
-                        tofind.add(n)
-            if not tofind:
-                return
-            time.sleep(0.1)
-        raise TimeoutError(f"watch_rest_for_alive() timeout after {timeout} seconds")
+        logging.getLogger('urllib3.connectionpool').disabled = True
```
nyh, line 1351: Looks good, but may I ask what was the intention of enabling urrlib3 logging in the first place?
Doesn't it always do EXACTLY what happened here - that if you do a lot of REST API requests you just log all of them?  Isn't it always a bad idea, not just in this function?

Who actually enabled this connection logging, was it CCM or dtest?

fruch: We log everything up to debug level into dtest log file.

We disable some things by level in dtest.

But those specific logs are very very useful when doing calls to http APIs, and we have quite a few of those in dtest, I don't want to disable them all.

PR #480, opened 2023-07-24
scylla_cluster,scylla_node: increase debug timeout for `watch_rest_for_alive`

fruch:
seems like the 360s default for debug for `watch_rest_for_alive` isn't enough as it was for `watch_log_for_alive`,
making it 600s (10min)

Fix: #477

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -146,7 +146,8 @@ class ScyllaCluster(Cluster):
             for old_node, _ in marks:
                 for node, _, _ in started:
                     if old_node is not node:
-                        old_node.watch_rest_for_alive(node)
+                        t = 120 if self.scylla_mode != 'debug' else 600
+                        old_node.watch_rest_for_alive(node, timeout=t)
 
         return started
 
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -309,7 +309,7 @@ class ScyllaNode(Node):
 
         if wait_other_notice:
             for node, _ in marks:
-                t = timeout if timeout is not None else 120 if self.cluster.scylla_mode != 'debug' else 360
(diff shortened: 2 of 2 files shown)


bhalevy:
@fruch can you please test this change with update_cluster_layout_tests.py::TestLargeScaleCluster::test_add_many_nodes_under_load?

fruch:
> @fruch can you please test this change with update_cluster_layout_tests.py::TestLargeScaleCluster::test_add_many_nodes_under_load in debug mode?

I've sent it on:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/327/

we'll need to check it out in few hours

fruch:
@bhalevy 

It passed:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/327/artifact/logs-long.debug.000/dtest-gw0.log/*view*/

even that I was a bit surprised it took only 11m (and start only 5 nodes)

bhalevy:
> @bhalevy
> 
> It passed: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/327/artifact/logs-long.debug.000/dtest-gw0.log/*view*/
> 
> even that I was a bit surprised it took only 11m (and start only 5 nodes)

It was changed in https://github.com/scylladb/scylla-dtest/pull/3268 to stop adding nodes once stress is complete
since the point in running it in debug mode is to see that's everything is stable and doesn't trigger any sanitizer errors.

bhalevy (approved):

nyh (approved):
Ok, but.. 10 minutes to boot a node? Wow, that's preposterous even for debug mode. Oh, well. 600 is just as arbitrary as 360, so I'm fine with that.

PR #481, opened 2023-07-25
scylla_repository: use LockFile for manager downloads

fruch:
now that we want to start using manager in parallel tests we need to make sure we don't try to download into the same direcotry at the same time.

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -412,23 +412,33 @@ def setup_scylla_manager(scylla_manager_package=None, verbose=False):
         dir_hash = get_url_hash(scylla_manager_package)
 
         # select a dir to change this version of scylla-manager based on the md5 of the path or the etag of the s3 object
-        manager_install_dir = directory_name(os.path.join('manager', dir_hash))
-        if not os.path.exists(manager_install_dir):
-            os.makedirs(manager_install_dir)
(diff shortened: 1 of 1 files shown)


ccmlib/scylla_repository.py
```diff
@@ -412,23 +412,33 @@ def setup_scylla_manager(scylla_manager_package=None, verbose=False):
...
+        with LockFile(manager_install_dir / DOWNLOAD_IN_PROGRESS_FILE) as f:
+            if f.read_status() != 'done':
+                # First ensure that we are working on a clean directory
+                # This prevents lockfile deletion by download_packages, as it doesn't have to clean the directory.
+                for p in Path(manager_install_dir).iterdir():
+                    if p.name != DOWNLOAD_IN_PROGRESS_FILE:
+                        if p.is_dir():
+                            shutil.rmtree(path=p)
+                        else:
+                            p.unlink(missing_ok=True)
```
Lorak-mmk, line 426: Now I'm wondering why my code from #470 works - probably there were only directories, but I think it should be changed to your version anyway to be safe.

Lorak-mmk (approved):

ccmlib/scylla_repository.py
fruch: yes, I was surprise I was falling on it.

but yes in scylla, there are only folders when you stop midway

PR #483, opened 2023-07-27
scylla_cluster: introduce `cluster.default_wait_*` properties

fruch:
* moving the setting those default timeout into the cluster `__init__`
  and a user of the cluster (i.e. dtest) can set those higher
  if needed.

* removed `timeout` argument from `_start_scylla` since no
  code in ccm or dtest were using it.

* reversed the order of `wait_for_binary_proto` with `wait_other_notice`
  since it was agree that `wait_for_starting` should happen first
  and the chances that after `wait_for_binary_proto` would end,
  `wait_other_notice` could finish much faster (current assumption
  is that no test is calling wait_other_notice=True while
  wait_for_binary_proto=False)

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -42,6 +42,8 @@ class ScyllaCluster(Cluster):
 
         self.started = False
         self.force_wait_for_cluster_start = (force_wait_for_cluster_start != False)
+        self.default_wait_other_notice_timeout = 120 if self.scylla_mode != 'debug' else 600
+        self.default_wait_for_binary_proto = 420 if self.scylla_mode != 'debug' else 900
         self._scylla_manager = None
         self.skip_manager_server = skip_manager_server
         self.scylla_version = cassandra_version
@@ -146,8 +148,7 @@ class ScyllaCluster(Cluster):
             for old_node, _ in marks:
                 for node, _, _ in started:
                     if old_node is not node:
-                        t = 120 if self.scylla_mode != 'debug' else 600
-                        old_node.watch_rest_for_alive(node, timeout=t)
+                        old_node.watch_rest_for_alive(node, timeout=self.default_wait_other_notice_timeout)
 
         return started
 
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -272,7 +272,7 @@ class ScyllaNode(Node):
         return bool(self.grep_log(f"{bootstrap_message}|{resharding_message}", from_mark=from_mark))
 
     def _start_scylla(self, args, marks, update_pid, wait_other_notice,
-                      wait_for_binary_proto, ext_env, timeout=None):
+                      wait_for_binary_proto, ext_env):
         log_file = os.path.join(self.get_path(), 'logs', 'system.log')
         # In case we are restarting a node
         # we risk reading the old cassandra.pid file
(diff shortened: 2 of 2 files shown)


fruch:
@bhalevy don't know if the best of interface, but it would allow for some test to set it higher before starting cluster/nodes

ccmlib/scylla_node.py
```diff
@@ -310,15 +310,15 @@ def _start_scylla(self, args, marks, update_pid, wait_other_notice,
...
+                t = self.cluster.default_wait_other_notice_timeout
                 node.watch_rest_for_alive(self, timeout=t)
                 self.watch_rest_for_alive(node, timeout=t)
 
         if wait_for_binary_proto:
-            t = timeout * 4 if timeout is not None else 420 if self.cluster.scylla_mode != 'debug' else 900
+            t = self.cluster.default_wait_for_binary_proto
             from_mark = self.mark
             try:
-                self.wait_for_binary_interface(from_mark=from_mark, process=self._process_scylla, timeout=timeout or 180)
```
fruch, line 321: @bhalevy I don't know if it's on purpose that 180 is hard coded here, and the `t` was only used in the exception

ccmlib/scylla_cluster.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, path, name, partitioner=None, install_dir=None,
 
         self.started = False
         self.force_wait_for_cluster_start = (force_wait_for_cluster_start != False)
+        self.default_wait_other_notice_timeout = 120 if self.scylla_mode != 'debug' else 600
```
bhalevy, line 45: 120 isn't enough apparently, I've seen it timeout locally in dev mode and also on jenkins in release mode

ccmlib/scylla_node.py
bhalevy: It's probably a bug.
I don't think we should maintain it

ccmlib/scylla_cluster.py
fruch: we seen it only on 50 nodes case ? i.e. this change is for that.

I don't want to double the default time, just for 1 test case

bhalevy: https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/314/testReport/update_cluster_layout_tests/TestLargeScaleCluster/Run_Dtest_Parallel_Cloud_Machines___LongDtest___long_split000___test_add_many_nodes_under_load/
```
ccmlib.node.TimeoutError: watch_rest_for_alive() timeout after 120 seconds
```

Why does it matter to increase the default value while timeouts are rare?

bhalevy: This happened on node32 node: https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-release/314/artifact/logs-long.release.000/1690338660733_update_cluster_layout_tests.py%3A%3ATestLargeScaleCluster%3A%3Atest_add_many_nodes_under_load/node32.log

ccmlib/scylla_node.py
fruch: I've put the same timeout for all of the calls under `wait_for_binary_proto`

ccmlib/scylla_cluster.py
fruch: it mean that in cases of failure, we'll wait more time until we get to the failure.
if for release it works for 99% of tests, I'm o.k. with 1% of the test specifying 
that they need extra timeout for them to boot.
 
I can easily put a default 1h, and it would probably would be enough for all of the cases,
devs are complaining again again, on how how slow/hard is to debug a dtest, not me.

ccmlib/scylla_node.py
bhalevy: ok, but we need to explain the reversal of order here in the commit message and/or code comment

fruch: @bhalevy updated the commit message with more details as for the order change

bhalevy (approved):

PR #484, opened 2023-07-28
scylla_node: add ScyllaNode.dump_sstables()

tchaikov:
`ScyllaNode.dump_sstables()` is a wrapper around
`ScyllaNode.run_scylla_sstable()`. it provides a more user friendly interface than the latter. it is introduced so that tests can use it with less pain when migrating from
`node.run_sstable2json()` to `node.run_scylla_sstable()`.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -3,6 +3,7 @@
 
 from datetime import datetime
 import errno
+import json
 import os
 import signal
 import shutil
@@ -13,6 +14,8 @@ import time
 import threading
 from pathlib import Path
 from collections import OrderedDict
+from typing import Any, Dict, List
+
 import logging
 
 import psutil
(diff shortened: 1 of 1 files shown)


tchaikov:
with the proposed change, we can have something like
```diff
diff --git a/compaction_test.py b/compaction_test.py
index 4b4059dd..53100bca 100644
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -180,19 +180,10 @@ class TestCompaction(Tester):
         node.flush()
         node.compact()
 
-        sstable_dumps = node.run_scylla_sstable('dump-data', ['--merge'],
-                                                keyspace=self.KEYSPACE_NAME)
-        numfound = 0
-        jsoninfo = []
-        for stdout, _ in sstable_dumps.values():
-            partitions = json.loads(stdout)['sstables']['anonymous']
-            jsoninfo += partitions
-            for partition in partitions:
-                if 'tombstone' in partition:
-                    numfound += 1
-
+        partitions = node.dump_sstables(keyspace=self.KEYSPACE_NAME)
+        numfound = sum('tombstone' in partition for partition in partitions)
         logger.debug(f'Number of tombstones found on node {node.name}: {numfound}')
-        return numfound, jsoninfo
+        return numfound, partitions
 
     def _test_compaction_delete_tombstone_gc(self, tombstone_gc_mode='repair', node_num: int = 2,
                                              r_factor: int = None, delete_keys: bool = True, partition_num: int = 100):
```

after https://github.com/scylladb/scylla-dtest/pull/3351 lands

ccmlib/scylla_node.py
```diff
@@ -1433,6 +1436,57 @@ def do_invoke(sstables):
...
+        if column_family is not None:
+            column_families = [column_family]
+            batch = True
+        sstable_dumps = self.run_scylla_sstable('dump-data', ['--merge'],
+                                                keyspace, datafiles, column_families,
+                                                batch)
+        all_partitions: List[Dict[str, Any]] = []
+        for stdout, _ in sstable_dumps.values():
+            partitions = json.loads(stdout)['sstables']['anonymous']
+            all_partitions += partitions
```
denesb, line 1487: Note that this will possibly mash together partitions from different tables. Which might be confusing. I think we should enforce that the user always passes a keyspace and exactly one table parameter. I think this is how most users will use this anyway. The rest (if any) can call `run_scylla_sstable()` instead.

tchaikov: changed accordingly.

ccmlib/scylla_node.py
```diff
@@ -1433,6 +1436,49 @@ def do_invoke(sstables):
                 ret[sst] = do_invoke([sst])
         return ret
 
+    def dump_sstables(self,
+                      keyspace: str,
+                      column_family: str) -> list[dict[str, Any]]:
+        """dump the partitions in the specified sstables using `scylla sstable dump-data`
```
denesb, line 1442: this comment is now out-of-date, we dump all sstables from the specified keyspace and table.

denesb (approved):
The tagline of comment of the method is outdated, LGTM otherwise.

tchaikov:
v2

- ks and cf are not optional anymore.
- do not accept `sstables`, as no users are using this parameter so far.

ccmlib/scylla_node.py
tchaikov: sure. revised.

tchaikov:
v3:

- revised the comment to reflect the latest API interface.

denesb (approved):

fruch (approved):
LGTM

PR #485, opened 2023-07-28
node.py: compact(): add keyspace and tables argument

denesb:
To allow specifying the keyspace/table to compact, instead of compacting all of them.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1336,8 +1336,12 @@ class Node(object):
             cmd += f" {table}"
         self.nodetool(cmd, **kwargs)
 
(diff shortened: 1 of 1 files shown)


denesb:
@fruch the tests failed but as far as I could determine, they have nothing to do with the changes I did. Unless I mis-understood something.

fruch:
> @fruch the tests failed but as far as I could determine, they have nothing to do with the changes I did. Unless I mis-understood something.

seems like it was a networking issue on github actions, and rerun of them passed

fruch (approved):
LGTM

PR #486, opened 2023-08-02
scylla_node: add datafiles param to ScyllaNode.dump_sstables()

tchaikov:
it turns out we do have a use case in dtest where the caller of `node.run_sstable2json()` passes a single sstable to this method. so let's cater its needs.

see compaction_additional_test.py

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -14,7 +14,7 @@ import time
 import threading
 from pathlib import Path
 from collections import OrderedDict
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 import logging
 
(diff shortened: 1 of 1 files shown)


fruch (approved):

ccmlib/scylla_node.py
```diff
@@ -1438,11 +1438,13 @@ def do_invoke(sstables):
 
     def dump_sstables(self,
                       keyspace: str,
-                      column_family: str) -> List[Dict[str, Any]]:
+                      column_family: str,
+                      datafile: Optional[str] = None) -> List[Dict[str, Any]]:
```
denesb, line 1442: Lets make this a list instead, it is more forward compatible, and you already use a list internally anyway.

tchaikov: sure. i don't see the needs so far. but yeah, it's more future-proof.

tchaikov:
@denesb could you take another look?

tchaikov:
v2:

- accept a list of sstables.

denesb (approved):

PR #487, opened 2023-08-02
scylla_node: do not error out when no sstables to dump

tchaikov:
`scylla sstable dump-data` errors out when there are no sstables specified. but we expect a simple API exposed by `dump_sstables()`, so that its caller does not have to handle the error in this case. so, in this change, we just change `run_scylla_sstable()` so that it fills an empty list in the json returned in the dict.

this should enable `dump_sstable()` to continue work. and its caller should be happy now.

also, this change removes the unused `env` local variable.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1418,11 +1418,18 @@ class ScyllaNode(Node):
         if additional_args is None:
             additional_args = []
         scylla_path = common.join_bin(self.get_path(), BIN_DIR, 'scylla')
-        env = self.get_env()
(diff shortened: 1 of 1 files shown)


denesb (approved):
The change makes sense.

tchaikov:
v2:

```diff
- empty_dump = {'': {'sstables': {'anonymous': []}}}
+ empty_dump = {'sstables': {'anonymous': []}}
```

bhalevy (approved):

bhalevy:

PR #488, opened 2023-08-02
ccmlib/node: run_sstable2json to handle non-exist files

fruch:
seems like introducing of raising failure on any failure of sstabledump has started to fail some dtests
since it was used in some places right after compation and some of the sstable were deleted between listing them to using them with sstable dump, so we need to ignore failures from non existing files.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -985,7 +985,10 @@ class Node(object):
             args = sstable2json + [sstablefile]
             if enumerate_keys:
                 args = args + ["-e"]
-            res = subprocess.run(args, env=env, stdout=out_file, text=True, check=True)
+            res = subprocess.run(args, env=env, stdout=out_file, stderr=subprocess.PIPE, text=True)
+            if not res.returncode == 0 and "Cannot find file" not in res.stderr:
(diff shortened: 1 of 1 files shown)


denesb (approved):

ccmlib/node.py
```diff
@@ -985,7 +985,10 @@ def run_sstable2json(self, out_file=None, keyspace=None, datafiles=None, column_
             args = sstable2json + [sstablefile]
             if enumerate_keys:
                 args = args + ["-e"]
-            res = subprocess.run(args, env=env, stdout=out_file, text=True, check=True)
+            res = subprocess.run(args, env=env, stdout=out_file, stderr=subprocess.PIPE, text=True)
+            if not res.returncode == 0 and "Cannot find file" not in res.stderr:
```
tchaikov, line 989: i'd suggest avoid `not` if possible:

```py
try:
    res = subprocess.run(args, env=env, stdout=out_file, text=True, check=True)
except CalledProcessError as e:
    if 'Cannot find file' in res.stderr:
        pass
    else:
        raise ToolError(e.cmd, e.returncode, e.stdout, e.stderr)
```

tchaikov: proposed at #489

PR #490, opened 2023-08-06
ccmlib/node: fix `run_sstable2json` on exception case

fruch:
a741f782bb4c9fef33bd3c42ac3bacb9a0f1cc42 has an issue it was looking into the wrong variable to check the stderr, and when there a failure would cause the
wrong failure

fix: scylladb/scylla-dtest#3454

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -987,12 +987,12 @@ class Node(object):
                 args = args + ["-e"]
             try:
                 res = subprocess.run(args, env=env, stdout=out_file, text=True, check=True)
+                common.print_if_standalone(res.stdout if res.stdout else '', debug_callback=self.info, end='')
             except subprocess.CalledProcessError as e:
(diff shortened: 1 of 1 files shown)


ccmlib/node.py
```diff
@@ -987,12 +987,12 @@ def run_sstable2json(self, out_file=None, keyspace=None, datafiles=None, column_
                 args = args + ["-e"]
             try:
                 res = subprocess.run(args, env=env, stdout=out_file, text=True, check=True)
+                common.print_if_standalone(res.stdout if res.stdout else '', debug_callback=self.info, end='')
```
tchaikov, line 990: okay. so we print out the stdout even only if the return code is not zero.

tchaikov (approved):
lgtm. just to be clear, in addition to the one explained in the commit message, this fix includes another change.

tchaikov:
thank you Israel for fixing this regression!

PR #493, opened 2023-08-08
node: let sstable_regexp match with sstable with uuid generation

tchaikov:
also update the comment to reflect the latest changes.

this should enable the dtest to test with uuid-based sstable identifiers enabled.


fruch (approved):
LGTM

PR #494, opened 2023-08-08
ccmlib/node: fix `run_sstable2json` on exception case - yet again

fruch:
seems like 730e556bce9a74e2bee9f5e00ced3968e272b8db wasn't enough and it didn't handled cases command would fail and stderr is None

also seem like we are not pipeing stderr, as we should for this exception handling to work correctly.


tchaikov (approved):
ahh, right! thanks!

PR #496, opened 2023-08-11
scylla_node: improve the sample return val of node.dump_sstable()

tchaikov:
so we can have better understanding of the format when using this function.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1455,6 +1455,10 @@ class ScyllaNode(Node):
         :return: return all the partitions collected in the specified sstables
         :raises: subprocess.CalledProcessError if scylla-sstable returns a non-zero exit code.
 
+        the $SSTABLE is returned from this method, please see
+        https://opensource.docs.scylladb.com/stable/operating-scylla/admin-tools/scylla-sstable.html#dump-data
+        for the JSON schema of it.
+
         a typical return value might look like:
         ```
         [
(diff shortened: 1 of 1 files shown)


tchaikov:
inspired by https://github.com/scylladb/scylla-dtest/pull/3487 , where we inspect the ttl of a cell.

denesb (approved):
Maybe add a link to https://opensource.docs.scylladb.com/stable/operating-scylla/admin-tools/scylla-sstable.html#dump-data, where the format is documented.

tchaikov:
v2:

- reference the link to the document of scylla-sstable

tchaikov:
@denesb thanks! revised to add the link.

ccmlib/scylla_node.py
```diff
@@ -1455,6 +1455,10 @@ def dump_sstables(self,
         :return: return all the partitions collected in the specified sstables
         :raises: subprocess.CalledProcessError if scylla-sstable returns a non-zero exit code.
 
+        only the $PARTITION is returned from this method, please see
```
denesb, line 1458: Isn't it the `$SSTABLE` symbol in fact that we return here?
```
$SSTABLE := [$PARTITION, ...]
```
We return multiple partitions.

tchaikov: ahh, indeed!

tchaikov:
v3:

- s/PARTITION/SSTABLE/ in the comment

tchaikov:
@scylladb/qa-maintainers hello maintainers, could you help merge this change?

PR #498, opened 2023-08-17
node: run_cqlsh: dump tool's stdout and stderr on timeout

bhalevy:
To help understand spurious timeout error we encounter.

Refs https://github.com/scylladb/scylla-cqlsh/issues/37

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -889,7 +889,18 @@ class Node(object):
                     p.stdin.write(cmd + ';\n')
             p.stdin.write("quit;\n")
 
-            output = p.communicate(timeout=timeout)
+            try:
+                output = p.communicate(timeout=timeout)
+            except subprocess.TimeoutExpired as e:
+                self.error(f"(EE) {e}")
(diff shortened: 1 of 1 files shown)


ccmlib/node.py
```diff
@@ -889,7 +889,18 @@ def run_cqlsh(self, cmds=None, show_output=False, cqlsh_options=None, return_out
                     p.stdin.write(cmd + ';\n')
             p.stdin.write("quit;\n")
 
-            output = p.communicate(timeout=timeout)
+            try:
+                output = p.communicate(timeout=timeout)
+            except subprocess.TimeoutExpired as e:
+                self.error(f"(EE) {e}")
+                pat = re.compile(r'\r*\n')
```
fruch, line 896: not sure why we need this regex for, wouldn't we print the whole outout/stdout ?

fruch (approved):
LGTM, but it won't fix the actual issue, I think it's a test we might consider moving to cqlsh itself.
i.e. it's check internal mechanism of cqlsh and not scylla.

PR #499, opened 2023-08-21
readme: Add info on creating a build-id cluster 

cvybhu:
Add instructions that describe how to create a cluster of Scylla nodes with a specific build-id.

It's not obvious how to achieve this - you have to download the right package from the right place, and then use the right combination of flags for `ccm`.

Let's add the instructions to make it easier for future users of `ccm`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ https://github.com/scylladb/scylla/wiki/Using-CCM
 
 Scylla usage examples:
 ---------------------------------------------------------------------
-### To create a 3-node Scylla cluster do:
+### Creating a 3-node Scylla cluster:
 ```bash
 $ ccm create my_cluster --scylla --vnodes -n 3 -v release:2022.2
 $ ccm start
@@ -29,7 +29,7 @@ node3: UP
 ```
 The nodes will be available at 127.0.0.1, 127.0.0.2 and 127.0.0.3.
 
-### To create a multi-datacenter cluster that has 3 nodes in dc1, 4 nodes in dc2 and 5 nodes in dc3 do:
+### Creating a multi-datacenter cluster that has 3 nodes in dc1, 4 nodes in dc2 and 5 nodes in dc3:
 ```bash
 $ ccm create my_multi_dc_cluster --scylla --vnodes -n 3:4:5 -v release:2022.2
 $ ccm start
--- a/README.md
+++ b/README.md
@@ -36,6 +36,22 @@ $ ccm start
 # Wait a lot...
 ```
 
+### Creating a cluster of nodes with a specific build-id:
+
+Let's say you want to create a cluster of Scylla with build id `f6e718548e76ccf3564ed2387b6582ba8d37793c` (it's `Scylla 2023.1.0~rc8-20230731.b6f7c5a6910c`).
+
+1. Go to https://backtrace.scylladb.com and find your desired Scylla version
+2. Click the arrow down symbol (\\/) to show all available download links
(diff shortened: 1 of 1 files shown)


ptrsmrn (approved):
Thank you very much for the step-by-step instruction!

README.md
```diff
@@ -15,7 +15,7 @@ https://github.com/scylladb/scylla/wiki/Using-CCM
 
 Scylla usage examples:
 ---------------------------------------------------------------------
-### To create a 3-node Scylla cluster do:
+### Create a 3-node Scylla cluster:
```
nyh, line 18: Nitpick: I would use the gerund ("creating") instead of the imperative ("create").
But it's just a nitpick, the "create" is also readable - it just sounds like a command, something you must do for some reason.

README.md
```diff
@@ -36,6 +36,22 @@ $ ccm start
 # Wait a lot...
 ```
 
+### Create a cluster of nodes with a specific build-id
+
+Let's say you want to create a cluster of `Scylla 2023.1.0~rc8-20230731.b6f7c5a6910c`, with build id `f6e718548e76ccf3564ed2387b6582ba8d37793c`.
```
nyh, line 41: It sounds like you are expected to know exactly which build id you are looking for, but in fact the first step below tells you how to find that build id. Maybe you should start saying you want "Scylla 2023.1.0" and then below you'll be told to search for that in backtrace.scylladb.com.

README.md
```diff
@@ -36,6 +36,22 @@ $ ccm start
...
+
+1. Go to https://backtrace.scylladb.com and find your desired Scylla version
+2. Click the arrow down symbol (\\/) to show all available download links
+3. Download the unified Scylla package (`unified-pack-url-x86_64`)
+4. Create a 3 node cluster:
+```bash
+# The unified package will be extracted to ~/.ccm/scylla-repository/my_custom_scylla
+# Make sure that the version name (my_custom_scylla) is different for each unified package you use, otherwise ccm will use the previously extracted version. 
+ccm create my_cluster -n 3 --scylla --vnodes \
+    --version my_custom_scylla \
```
nyh, line 51: Wow. If you don't invent this name "my_custom_scylla", can't it just use the full name of the downloaded version? :-(

nyh (approved):
Looks good, I have some minor suggestions for improvement, but as-is it is already - obviously - better than what we have today.

README.md
cvybhu: Without the `version` argument it fails with:
```bash
ccm: error: ./ is not a valid cassandra directory. You must define a cassandra dir or version.
```

I don't know if there's a better way to do it. I found this one in the `dtest` docs I think:
https://github.com/scylladb/scylla-dtest/blob/3fe7961e06ff02c72000b8c162a3838185456a39/README.md?plain=1#L204

README.md
cvybhu: Okay changed to `Creating`.

README.md
cvybhu: I have the `build-id` and I want to find a download for the package with this build-id, so I use https://backtrace.scylladb.com to get the download link based on the `build-id` alone.

I modified this line to focus on the `build-id` more.

cvybhu:
v2:
* `Create` -> `Creating`
* Improve the line to focus on the fact that we have only `build-id`

fruch:
@cvybhu 
you might refer also to https://github.com/scylladb/scylla-ccm/blob/next/scripts/download_master.sh, isn't doing something quite similar, but looking for the latest build on the master branch, and not a specific build-id.

PR #501, opened 2023-08-29
scylla_node: populate env variable when running scylla-sstable 

tchaikov:
if scylla is built in a dtest environment, and the shared libraries
which it is linked against cannot be found in the testbed's default ld.so.conf
paths, the tests which run scylla-sstable would fail.

so, in this change. let's apply the `self._launch_env` when
running scylla-sstable as well. previously, these env variables
are only applied when launching scylla as a daemon.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1414,7 +1414,11 @@ class ScyllaNode(Node):
                 stdout, stderr = json.dumps(empty_dump), ''
                 return (stdout, stderr)
             common_args = [scylla_path, "sstable", command] + additional_args
-            res = subprocess.run(common_args + sstables, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, check=True)
+            try:
+                env = self._launch_env
+            except AttributeError:
+                env = os.environ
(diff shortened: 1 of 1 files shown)


tchaikov:
tested using
```console
$ scripts/dbuild_collect_so.sh ~/dev/scylladb/build/cmake/scylla /tmp/dynamic_libs_for_dtest
$ SCYLLA_DBUILD_SO_DIR=/tmp/dynamic_libs_for_dtest JAVA_HOME=/usr/lib/jvm/jre-11 pytest --cassandra-dir $HOME/dev/scylladb/build/cmake scylla_sstable_test.py::TestScyllaSstableDumpData::test_scylla_sstable_basic
```

not really a dbuild build. but i also verified by checking the env variables manually.

bhalevy (approved):

tchaikov:
https://github.com/scylladb/scylla-ccm/actions/runs/6011490212/job/16304966887?pr=501#step:6:627

i am not sure why s-c failed to connect to scylla ..

tchaikov:
@bhalevy thank you for your review. as this change fails the ccm CI test. see above. i am closing this one in favor of a minimal test: #502 .  could you please review #502 instead?

will try to bring the cleanup change back later.

PR #502, opened 2023-08-29
scylla_node: populate env variable when running scylla-sstable

tchaikov:
if scylla is built in a dtest environment, and the shared libraries which it is linked against cannot be found in the testbed's default ld.so.conf paths, the tests which run scylla-sstable would fail.

so, in this change. let's apply the `self._launch_env` when running scylla-sstable as well. previously, these env variables are only applied when launching scylla as a daemon.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1414,7 +1414,11 @@ class ScyllaNode(Node):
                 stdout, stderr = json.dumps(empty_dump), ''
                 return (stdout, stderr)
             common_args = [scylla_path, "sstable", command] + additional_args
-            res = subprocess.run(common_args + sstables, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, check=True)
+            try:
+                env = self._launch_env
+            except AttributeError:
+                env = os.environ
+            res = subprocess.run(common_args + sstables, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, check=True, env=env)
             return (res.stdout, res.stderr)
 
         if batch:


tchaikov:
only for testing the CCM CI test using Nix.

tchaikov:
tested using
```console
$ scripts/dbuild_collect_so.sh ~/dev/scylladb/build/cmake/scylla /tmp/dynamic_libs_for_dtest
$ SCYLLA_DBUILD_SO_DIR=/tmp/dynamic_libs_for_dtest JAVA_HOME=/usr/lib/jvm/jre-11 pytest --cassandra-dir $HOME/dev/scylladb/build/cmake scylla_sstable_test.py::TestScyllaSstableDumpData::test_scylla_sstable_basic
```

not really a dbuild build. but i also verified by checking the env variables manually.

fruch (approved):
LGTM

fruch:
I'll for @bhalevy to confirm, as the one that was hit by this issue, to validate it's solving it

denesb (approved):

tchaikov:
> I'll for @bhalevy to confirm, as the one that was hit by this issue, to validate it's solving it

@fruch Israel, Benny approved a superset of this change. see #501. but it introduced a regression identifier by the ccm's own unit tests. so i abandoned it.

bhalevy:
> if scylla is built in a dtest environment,

nit: s/dtest/dbuild/

> and the shared libraries which it is linked against cannot be found in the testbed's default ld.so.conf paths, the tests which run scylla-sstable would fail.
> 
> so, in this change. let's apply the `self._launch_env` when running scylla-sstable as well. previously, these env variables are only applied when launching scylla as a daemon.

ccmlib/scylla_node.py
```diff
@@ -1414,7 +1414,11 @@ def do_invoke(sstables):
                 stdout, stderr = json.dumps(empty_dump), ''
                 return (stdout, stderr)
             common_args = [scylla_path, "sstable", command] + additional_args
-            res = subprocess.run(common_args + sstables, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, check=True)
+            try:
+                env = self._launch_env
+            except AttributeError:
+                env = os.environ
+            res = subprocess.run(common_args + sstables, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, check=True, env=env)
```
bhalevy, line 1421: LGTM.  This is almost identical to the change I made in my tree.

bhalevy (approved):

PR #503, opened 2023-08-30
scylla_node: add a helper for accessing environ variables

tchaikov:
scylla_node: add a helper for accessing environ variables

before this change, we always try to access `self._launch_env`
and fall back to `os.environ`, and optionally update the
return env dict, then use it. the pattern repeats itself
couple times in ScyllaNode class.

in this change, to consolidate it, a helper is added. it
does not do anything clever just preserve existing behavior.
despite that existing behavior could be potentially
confusing: we actually _mutate_ the returned dict if
it references the existing `self._launch_env`.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -148,6 +148,16 @@ class ScyllaNode(Node):
         return common.make_cassandra_env(self.get_install_cassandra_root(),
                                          self.get_node_cassandra_root(), update_conf=update_conf)
 
+    def _get_environ(self, extra_env = None, /, **kwargs):
+        try:
+            env = self._launch_env
+        except AttributeError:
+            env = dict(os.environ)
+        if extra_env is not None:
+            env.update(extra_env)
+        env.update(kwargs)
+        return env
+
     def get_cassandra_version(self):
         # TODO: Handle versioning
         return '3.0'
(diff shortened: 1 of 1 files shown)


ccmlib/scylla_node.py
```diff
@@ -47,6 +47,7 @@ def __init__(self, name, cluster, auto_bootstrap, thrift_interface,
         self._node_install_dir = None
         self._node_scylla_version = None
         self._relative_repos_root = None
+        self._launch_env = dict(os.environ)
```
fruch, line 50: I'm not sure I like the idea of copying the env at class __init__,
it would block the test from doing things like change env and restart the node, as an example.

I would suggest turn it into a property function that would do the copy as needed, and add need parameters like `LD_LIBRARY_PATH` as needed

tchaikov: @fruch hi Israel, thank you for your review. just wanted to understand the use case better: so it's something like

```py
node = ScyllaNode()
os.environ['FOO'] = 'bar'
node.start() # we expect $FOO to take effect here
```

right?

tchaikov: i failed to foresee a use case like this but anyway. will change to a property as you suggest. personally, i am not a big fan of too many knobs.

tchaikov:
v2:

- instead of initializing `self._launch_env` in constructor, introduce a helper for accessing / updating it.

@fruch hi Israel, could you help re-review this change?

fruch (approved):
LGTM

PR #506, opened 2023-09-03
ccmlib/node: add `env` argument to `Node.run_cqlsh()`

fruch:
for enabling more of dtest need, added `env` argument to 
pass in environment variables needed for the test.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -863,12 +863,16 @@ class Node(object):
                         print(log, end='')
                     i = i + 1
 
-    def run_cqlsh(self, cmds=None, show_output=False, cqlsh_options=None, return_output=False, timeout=600):
+    def run_cqlsh(self, cmds=None, show_output=False, cqlsh_options=None, return_output=False, timeout=600, extra_env=None):
         cqlsh_options = cqlsh_options or []
         cqlsh = self.get_tool('cqlsh')
         if not isinstance(cqlsh, list):
             cqlsh = [cqlsh]
+
         env = self.get_env()
+        if extra_env:
+            env.update(extra_env)
+
         host = self.network_interfaces['thrift'][0]
         if self.get_base_cassandra_version() >= 2.1:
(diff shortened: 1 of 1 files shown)


ccmlib/node.py
```diff
@@ -863,12 +863,15 @@ def run_cli(self, cmds=None, show_output=False, cli_options=[]):
                         print(log, end='')
                     i = i + 1
 
-    def run_cqlsh(self, cmds=None, show_output=False, cqlsh_options=None, return_output=False, timeout=600):
+    def run_cqlsh(self, cmds=None, show_output=False, cqlsh_options=None, return_output=False, timeout=600, env=None):
```
tchaikov, line 866:
```suggestion
    def run_cqlsh(self, cmds=None, show_output=False, cqlsh_options=None, return_output=False, timeout=600, extra_env=None):
```

ccmlib/node.py
```diff
@@ -863,12 +863,15 @@ def run_cli(self, cmds=None, show_output=False, cli_options=[]):
...
         cqlsh_options = cqlsh_options or []
         cqlsh = self.get_tool('cqlsh')
         if not isinstance(cqlsh, list):
             cqlsh = [cqlsh]
-        env = self.get_env()
+        if env:
+            env = self.get_env().update(env)
+        else:
+            env = self.get_env()
```
tchaikov, line 874:
```suggestion
        env = self.get_env()
        if extra_env is not None:
            env.update(extra_env)
        return env
```
        else:
            env = self.get_env()
```

tchaikov (approved):
left a couple nit. just for better readability.  we are adding/overriding existing `env`. so `extra_env` might be able to pass this message to programmer better.

tchaikov (approved):
lgtm

bhalevy (approved):

ccmlib/node.py
fruch: done

PR #508, opened 2023-09-05
scylla_repository: remove the need of `scylla_product`

fruch:
change the code to be able to scan across both OSS and enterprise location to find a release or unstable relocatable package

`SCYLLA_PRODUCT` envirment variable is still in the code but used only when fetching really old release that doesn't have unified relocatable package (ex. 2020.1), leave this code so we can still fetch and use those old versions


juliayakovlev (approved):

bhalevy:
What's the status of this PR?

fruch:
> What's the status of this PR?

It's actually ready and working just fine (verified it again, with release version, and with branch versions)

roydahan:
Which branches we need to backport to?
(Status is "Waiting backport")

Issue #509, opened 2023-09-15
Add tablets support to node stress command


soyacz:
When testing tablets, keyspaces must be created with additional parameter `initial_tablets`. Also `NetworkTopologyStrategy` is required.

Request is to make `node.stress` command automatically adding proper stress_options when tablets are enabled.

PR #510, opened 2023-09-15
enable tablets in kespaces created by node.stress

soyacz:
We want to test tablets functionality without need for test adaptation (creating keyspace with tablets enabled  before node.stress is run).

Modify `stress_options` by add `initial_tablets` option and change replication strategy to `NetworkTopologyStrategy` to support tablets tests.

fixes: https://github.com/scylladb/scylla-ccm/issues/509

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -720,3 +720,7 @@ class Cluster(object):
     @staticmethod
     def is_docker():
         return False
+
+    @property
+    def tablets_enabled(self) -> bool:
+        return 'tablets' in self._config_options.get('experimental_features', [])
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -4,6 +4,7 @@
 import errno
 import glob
 import itertools
+import math
 import os
 import re
 import shutil
@@ -16,6 +17,7 @@ import warnings
 from datetime import datetime
 import locale
 from collections import namedtuple
+from typing import List, Optional
 
 import yaml
 
@@ -1277,7 +1279,14 @@ class Node(object):
                 # specify used jmx port if not already set
                 if not [opt for opt in stress_options if opt.startswith('jmx=')]:
                     stress_options.extend(['-port', 'jmx=' + self.jmx_port])
-
+        if self.cluster.tablets_enabled:
+            smp = sum([node._smp or 1 for node in self.cluster.nodes.values()])
+            # default initial tablets: 32 * `sum of SMPs in cluster` coerced to power of 2 (suitable for small number of tables)
+            initial_tablets = int(math.pow(2, math.ceil(math.log(smp*32, 2))))
+            self.debug("tablets enabled, adjusting stress options by replication strategy change.")
+            stress_options = self._set_keyspace_initial_tablets(
+                stress_options, datacenter=self.data_center, initial_tablets=initial_tablets
+            )
         args = stress + stress_options
         stdout_handle = kwargs.pop("stdout", subprocess.PIPE)
         stderr_handle = kwargs.pop("stderr", subprocess.PIPE)
@@ -1294,6 +1303,41 @@ class Node(object):
             return handle_external_tool_process(p, ['stress'] + stress_options)
         except KeyboardInterrupt:
             pass
+
+    @staticmethod
+    def _set_keyspace_initial_tablets(stress_options: List[str], initial_tablets:int = 8,
+                                      datacenter: Optional[str] = None, rf: int = 1) -> List[str]:
+        normalized_stress_options = []
+        for stress_option in stress_options:
+            # sometimes users pass schema params in one str, split it
+            if stress_option.startswith("-schema"):
+                if " " in stress_option:
+                    normalized_stress_options += stress_option.split()
+                    continue
+            normalized_stress_options.append(stress_option)
+        stress_options = normalized_stress_options
+        datacenter = datacenter or 'datacenter1'
+        default_replication = f"replication(strategy=NetworkTopologyStrategy,{datacenter}={rf},initial_tablets={initial_tablets})"
+        try:
+            idx = stress_options.index('-schema')
+            replication = stress_options[idx + 1]
+            if 'replication' in replication:
+                if 'NetworkTopologyStrategy' in replication:
+                    # if NTP already set, just prepend initial_tablets
+                    replication = re.sub(r'replication\(.+?\)',
+                                         lambda m: m.group(0).replace(')', f',initial_tablets={initial_tablets})'), replication)
+                else:
+                    replication = re.sub(r'replication\(.*factor=(\d+)\)',
+                                         f'replication(strategy=NetworkTopologyStrategy,{datacenter}=' + r'\1' + f',initial_tablets={initial_tablets})',
+                                         replication)
+                stress_options[idx + 1] = replication
+            else:
+                # no replication provided in -schema, use default
+                stress_options.insert(idx + 1, default_replication)
+        except ValueError:
+            # no -schema provided
+            stress_options += ['-schema', default_replication]
+        return stress_options
 
     @staticmethod
     def _set_stress_val(key, val, res):
--- a/tests/test_set_keyspace_initial_tablets.py
+++ b/tests/test_set_keyspace_initial_tablets.py
@@ -1,0 +1,30 @@
+import pytest
+from typing import List, Optional
+
+from ccmlib.node import Node
+
+
+@pytest.mark.parametrize("stress_options, initial_tablets, datacenter, rf, expected", [
+    # typical case
+    (['write', 'n=0', 'no-warmup', '-schema', 'replication(factor=1)', '-rate', 'threads=1'], 8, None, 3,
+     ['write', 'n=0', 'no-warmup', '-schema', 'replication(strategy=NetworkTopologyStrategy,datacenter1=1,initial_tablets=8)', '-rate', 'threads=1']),
+    # no replication
+    (['write', 'n=3500K', '-rate', 'threads=50', '-schema', 'compaction(strategy=SizeTieredCompactionStrategy)'], 8, None, 3,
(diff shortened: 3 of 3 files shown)


ccmlib/node.py
```diff
@@ -197,6 +198,9 @@ def is_scylla(self):
     def is_docker():
         return False
 
+    def tablets_enabled(self) -> bool:
+        return 'tablets' in self.cluster._config_options.get('experimental_features', [])
```
bhalevy, line 202: nit: this function should be implemented in cluster.py and be called here

ccmlib/node.py
```diff
@@ -1277,7 +1281,10 @@ def stress_process(self, stress_options, **kwargs):
                 # specify used jmx port if not already set
                 if not [opt for opt in stress_options if opt.startswith('jmx=')]:
                     stress_options.extend(['-port', 'jmx=' + self.jmx_port])
-
+        if self.tablets_enabled():
+            stress_options = self._set_keyspace_initial_tablets(
+                stress_options, datacenter=self.data_center, rf=min(len(self.cluster.nodes), 3)
```
bhalevy, line 1286: maybe be the default rf should be calculated this way in `_set_keyspace_initial_tablets`
so the logic could be reused in the future if and when called from other places

ccmlib/node.py
```diff
@@ -1295,6 +1302,41 @@ def stress(self, stress_options=None, capture_output=False, **kwargs):
...
+        default_replication = f"replication(strategy=NetworkTopologyStrategy,{datacenter}={rf},initial_tablets={initial_tablets})"
+        try:
+            idx = stress_options.index('-schema')
+            replication = stress_options[idx + 1]
+            if 'replication' in replication:
+                if 'NetworkTopologyStrategy' in replication:
+                    # if NTP already set, just prepend initial_tablets
+                    replication = re.sub(r'replication\(.+?\)',
+                                         lambda m: m.group(0).replace(')', f',initial_tablets={initial_tablets})'), replication)
+                else:
```
bhalevy, line 1327: shouldn't we check for `SimpleStrategy`?
what if it's `EverywhereStrategy`?
Maybe the funtion should throw an error in this case

soyacz: currently anything but NetworkStrategy will be overridden completely. I don't know much about EverywhereStrategy - is it available and can be used by user keyspaces?

ccmlib/node.py
soyacz: yes, at first I wanted also this way, but to ease unit testing i made it this way. I moved this calculation to `cluster` object. WDYT?

ccmlib/node.py
fruch: and more, can you pass initial_tablets on `EverywhereStrategy` and `SimpleStrategy` ?

setup.py
```diff
@@ -16,7 +16,7 @@
 
 setup(
     name='ccm',
-    version='2.0.5',
+    version='2.0.6',
```
fruch, line 19: why changing this ? i.e. we are not uploading CCM anywhere, no point of changing this

tests/test_set_keyspace_initial_tablets.py
```diff
@@ -0,0 +1,30 @@
+import pytest
+from typing import List, Optional
+
+from ccmlib.node import Node
+
+
+@pytest.mark.parametrize("stress_options, initial_tablets, datacenter, rf, expected", [
+    # typical case
+    (['write', 'n=0', 'no-warmup', '-schema', 'replication(factor=1)', '-rate', 'threads=1'], 8, None, 3,
+     ['write', 'n=0', 'no-warmup', '-schema', 'replication(strategy=NetworkTopologyStrategy,datacenter1=1,initial_tablets=8)', '-rate', 'threads=1']),
```
fruch, line 10: looking at those example, I don't think we should be changing the `strategy`, that might be changing the test logic more then we intend.

SimpleStrategy should accept `initial_tablets`, and if not, I think it's a bug

ccmlib/node.py
soyacz: `SimpleStrategy` fails with `ConfigurationException: Unrecognized strategy option {initial_tablets} passed to org.apache.cassandra.locator.SimpleStrategy for keyspace keyspace1` 
`EverywhereStrategy` - I can create but it ignores tablets option (it is mentioned in describe keyspace, but `initial_tablets` don't change anything in sstables structure).

setup.py
soyacz: is this causing any problems? 
I mean, when I reinstall ccm I always see 2.0.5 and i wonder if there was some change...

tests/test_set_keyspace_initial_tablets.py
soyacz: `initial_tablets` are not planned for `SimpleStrategy`, at least not in near future. 

I don't see any other option than change this to `NetworkTopologyStrategy`.
Shall we split this change to first adapt dtests to NTS?

setup.py
fruch: It's irrelevant to this change

Since we don't have releases.

We can open an issue for starting doing so, I would love putting CCM up in pypi, it can help us in lots of places.

But for now we stick to branches.

soyacz: fixed.
I also consider puting ccm to pypi as a good idea.

ccmlib/cluster.py
```diff
@@ -720,3 +720,11 @@ def error(self, message):
     @staticmethod
     def is_docker():
         return False
+
+    @property
+    def tablets_enabled(self) -> bool:
+        return 'tablets' in self._config_options.get('experimental_features', [])
+
+    @property
```
fruch, line 728: I think we should do the c-s default, which is 1

ccmlib/node.py
fruch: EverywhereStrategy is used in 3 dtest modules only, non of them uses c-s, we should check if c-s even supports using it

ccmlib/cluster.py
soyacz: true. Fixed.

fruch (approved):
LGTM

ccmlib/node.py
```diff
@@ -1295,6 +1300,41 @@ def stress(self, stress_options=None, capture_output=False, **kwargs):
         except KeyboardInterrupt:
             pass
 
+    @staticmethod
+    def _set_keyspace_initial_tablets(stress_options: List[str], initial_tablets:int = 8,
```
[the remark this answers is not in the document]
fruch: Is there some reference to how one should pick this number ?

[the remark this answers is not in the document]
avikivity: Minimum: 1*(total number of vcpus in the cluster) (suitable for up to 100 tables)
Maximum: 33*(total number of vcpus in the cluster) (suitable for a small number of tables)

For ccm you can treat the boundaries as fairly fluid. Better to have too many than too few. Eventually it will just be a hint and the load balancer will recover from bad inputs.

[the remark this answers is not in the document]
soyacz: ccm usecase is running scylla on local machines, usually they are not too powerful (and run mutliple scylla nodes) and clusters are not loaded too much with data.
Also, this is just 'initial' number of tablets and in future split/merge will be implemented and this setting will not have that much impact I believe.

I can leave as is currently or implement some logic behind like:
`self._smp * 8` and then to match the nearest bigger power of 2 (IIRC tablets number must be always power of 2):
```
import math

def multiply_and_find_nearest_power_of_two(smp):
     smp *= 8
     smp = math.pow(2, math.ceil(math.log(smp, 2)))
     return smp

```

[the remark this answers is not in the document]
fruch: @soyacz where this 8 is coming from ? @avikivity suggested thet the maximum should be used.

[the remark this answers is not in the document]
avikivity: Note you shouldn't use smp directly, rather the sum of all smp values across the cluster (smp * nodes).

And it doesn't need to be exact.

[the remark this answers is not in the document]
soyacz: I updated the code with higher number (smp * 32, coerced up to power of 2)

fruch (approved):
LGTM

fruch:
@bhalevy @avikivity, care to give it a final review ?

ccmlib/node.py
```diff
@@ -1277,7 +1279,14 @@ def stress_process(self, stress_options, **kwargs):
                 # specify used jmx port if not already set
                 if not [opt for opt in stress_options if opt.startswith('jmx=')]:
                     stress_options.extend(['-port', 'jmx=' + self.jmx_port])
-
+        if self.cluster.tablets_enabled:
+            smp = sum([node._smp or 1 for node in self.cluster.nodes.values()])
+            # default initial tablets: 32 * `sum of SMPs in cluster` coerced to power of 2 (suitable for small number of tables)
+            initial_tablets = int(math.pow(2, math.ceil(math.log(smp*32, 2))))
+            self.debug("tablets enabed, adjusting stress options by replication strategy change.")
```
avikivity, line 1286: enabled

ccmlib/node.py
```diff
@@ -1277,7 +1279,14 @@ def stress_process(self, stress_options, **kwargs):
...
+            stress_options = self._set_keyspace_initial_tablets(
+                stress_options, datacenter=self.data_center, rf=1, initial_tablets=initial_tablets
```
avikivity, line 1288: Why rf=1? seems unrelated here.

fruch: It's just used to setup a default which matches cassandra stress default which if you don't define the schema option it would be SimpleStrategy with rf=1

I agree it's a bit weird to have it in the arguments if it's a default.

soyacz: updated

ccmlib/node.py
soyacz: updated

PR #512, opened 2023-09-20
ccmlib/common: support yaml syntax in updateconf command values

fruch:
since in some place we need to be able to pass in a list
for example in `experimental_features`, ccm didn't support it
in the `ccm updateconf` command, now it would support it
but passing it via the command line, like the following:

```bash
> ccm updateconf experimental_features:[udf,tablets]
```

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -769,11 +769,7 @@ def normalize_interface(itf):
 def parse_settings(args):
     settings = {}
     for s in args:
-        if is_win():
-            # Allow for absolute path on Windows for value in key/value pair
-            splitted = s.split(':', 1)
-        else:
-            splitted = s.split(':')
+        splitted = s.split(':', 1)
         if len(splitted) != 2:
             raise ArgumentError("A new setting should be of the form 'key: value', got " + s)
         key = splitted[0].strip()
@@ -784,10 +780,7 @@ def parse_settings(args):
         elif val.lower() == "false":
             val = False
         else:
-            try:
-                val = int(val)
-            except ValueError:
-                pass
+            val = yaml.safe_load(val)
         splitted = key.split('.')
         if len(splitted) == 2:
             try:
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -1,6 +1,10 @@
-from ccmlib.common import scylla_extract_mode, LockFile
 import tempfile
 import os
+
+import pytest
+import yaml
+
+from ccmlib.common import scylla_extract_mode, LockFile, parse_settings
 
 
 def test_scylla_extract_mode():
@@ -69,3 +73,25 @@ def test_lockfile_retain_status_by_default():
     assert lf.acquire(blocking=False)[0] is True
     assert lf.read_status() == 'some_status_1'
     lf.release()
+
+def test_parse_settings():
+    res = parse_settings(["number:12", "string:somthing", "bool:false"])
+    assert res == {'bool': False, 'number': 12, 'string': 'somthing'}
+
+    res = parse_settings(["nested.number:12", "nested.string:somthing", "nested.bool:false"])
+    assert res == {'nested': {'bool': False, 'number': 12, 'string': 'somthing'}}
+
+    # a fix that need to be backported from upstream
+    # https://github.com/riptano/ccm/commit/37afe6ab86fe03d5be7d0cf7a328dccac035a9a0
+    # res = parse_settings(["double.nested.number:12", "double.nested.string:somthing", "double.nested.bool:false"])
+    # assert res == {'double': {'nested': {'bool': False, 'number': 12, 'string': 'somthing'}}}
+
+    res = parse_settings(["experimental_features:[udf,tablets]"])
+    assert res == {'experimental_features': ['udf', 'tablets']}
+
+    res = parse_settings(["experimental_features:['udf',\"tablets\"]"])
+    assert res == {'experimental_features': ['udf', 'tablets']}
+
+    # would break if incorrect yaml format is passed in the value
+    with pytest.raises(yaml.parser.ParserError):
+        parse_settings(["experimental_features:['udf',\"tablets\""])
--- a/ccmlib/cmds/cluster_cmds.py
+++ b/ccmlib/cmds/cluster_cmds.py
@@ -830,7 +830,10 @@ class ClusterUpdateconfCmd(Cmd):
         return "Update the cassandra config files for all nodes"
 
     def get_parser(self):
-        usage = "usage: ccm updateconf [options] [ new_setting | ...  ], where new_setting should be a string of the form 'compaction_throughput_mb_per_sec: 32'; nested options can be separated with a period like 'client_encryption_options.enabled: false'"
+        usage = ("usage: ccm updateconf [options] [ new_setting | ...  ], "
+                 "where new_setting should be a string of the form 'key:val'; "
+                 "where 'key' is the name of an option, 'val' is the YAML representation of its value, "
+                 "nested options can be separated with a period like 'client_encryption_options.enabled: false'")
         parser = self._get_default_parser(usage, self.description())
         parser.add_option('--no-hh', '--no-hinted-handoff', action="store_false",
                           dest="hinted_handoff", default=True, help="Disable hinted handoff")
@@ -870,7 +873,10 @@ class ClusterUpdatedseconfCmd(Cmd):
         return "Update the dse config files for all nodes"
 
     def get_parser(self):
-        usage = "usage: ccm updatedseconf [options] [ new_setting | ...  ], where new_setting should be a string of the form 'max_solr_concurrency_per_core: 2'; nested options can be separated with a period like 'cql_slow_log_options.enabled: true'"
+        usage = ("usage: ccm updatedseconf [options] [ new_setting | ...  ], "
+                 "where new_setting should be a string of the form 'key:val'; "
+                 "where 'key' is the name of an option, 'val' is the YAML representation of its value, "
+                 "nested options can be separated with a period like 'client_encryption_options.enabled: false'")
         parser = self._get_default_parser(usage, self.description())
         return parser
 
--- a/ccmlib/cmds/node_cmds.py
+++ b/ccmlib/cmds/node_cmds.py
@@ -590,7 +590,10 @@ class NodeUpdateconfCmd(Cmd):
         return "Update the cassandra config files for this node (useful when updating cassandra)"
 
     def get_parser(self):
-        usage = "usage: ccm node_name updateconf [options] [ new_setting | ...  ], where new_setting should be a string of the form 'compaction_throughput_mb_per_sec: 32'"
+        usage = ("usage: ccm node_name updateconf [options] [ new_setting | ...  ], "
+                 "where new_setting should be a string of the form 'key:val'; "
+                 "where 'key' is the name of an option, 'val' is the YAML representation of its value, "
+                 "nested options can be separated with a period like 'client_encryption_options.enabled: false'")
         parser = self._get_default_parser(usage, self.description())
         parser.add_option('--no-hh', '--no-hinted-handoff', action="store_false",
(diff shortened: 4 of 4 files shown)


ccmlib/common.py
```diff
@@ -785,9 +786,12 @@ def parse_settings(args):
             val = False
         else:
             try:
-                val = int(val)
-            except ValueError:
+                val = ast.literal_eval(str(val))
```
tchaikov, line 789: what if we have something like: `hello: "world"` here? before this change, the `val` is `'"world"'`, after this change the `val` would be `'world'`. and what if we really want to set a value like `'1e10'`. i still cannot tell if this is a valid concern after reading 
https://github.com/scylladb/scylla-ccm/blob/69118c7185e5b97a1d923084356f5b4dc4fbec8d/ccmlib/cmds/cluster_cmds.py#L832-L841 and https://github.com/scylladb/scylla-ccm/blob/69118c7185e5b97a1d923084356f5b4dc4fbec8d/ccmlib/cmds/cluster_cmds.py#L872-L875

or do we accept `foo: [1, False, "hello"]` ?

actually, i was about to suggest using `json.loads()` but i just realized that we are not actually expecting json encoded data here. it's kind of freestyle format. but i still think it would be better to stick to a well-defined format / encoding, so we don't need to worry about these things.

but if we don't really care. i'd suggest at least handle more types of exceptions, see https://docs.python.org/3/library/ast.html#ast.literal_eval

fruch: I think

> what if we have something like: `hello: "world"` here? before this change, the `val` is `'"world"'`, after this change the `val` would be `'world'`. and what if we really want to set a value like `'1e10'`. i still cannot tell if this is a valid concern after reading

regarding the example  `hello: "world"` in yaml it's equivalent to `hello: world`, so in less you'll want something complex like multi line string, this should be o.k.

90% of the cases we or need a string or need a number in configurations
this fix helps to handle the case we need a list (and it would also gonna work for mapping)

> 
> https://github.com/scylladb/scylla-ccm/blob/69118c7185e5b97a1d923084356f5b4dc4fbec8d/ccmlib/cmds/cluster_cmds.py#L832-L841
> 
> and
> https://github.com/scylladb/scylla-ccm/blob/69118c7185e5b97a1d923084356f5b4dc4fbec8d/ccmlib/cmds/cluster_cmds.py#L872-L875
> 
> or do we accept `foo: [1, False, "hello"]` ?

this should work

> 
> actually, i was about to suggest using `json.loads()` but i just realized that we are not actually expecting json encoded data here. it's kind of freestyle format. but i still think it would be better to stick to a well-defined format / encoding, so we don't need to worry about these things.
> 
> but if we don't really care. i'd suggest at least handle more types of exceptions, see https://docs.python.org/3/library/ast.html#ast.literal_eval

I rather not handle those, let it explode, i.e. if user put in a giant literal or that complex one, that it would hit recursion limit, I don't want to handle it gracefully, and silently put it as a string (which is probably never what he would expect in this case).

tchaikov: > I think
> 
> > what if we have something like: `hello: "world"` here? before this change, the `val` is `'"world"'`, after this change the `val` would be `'world'`. and what if we really want to set a value like `'1e10'`. i still cannot tell if this is a valid concern after reading
> 
> regarding the example `hello: "world"` in yaml it's equivalent to `hello: world`, so in less you'll want something complex like multi line string, this should be o.k.

how does yaml come into play? are we consuming `val` later with `yaml.load()` ?

> 
> 90% of the cases we or need a string or need a number in configurations this fix helps to handle the case we need a list (and it would also gonna work for mapping)
> 
> > https://github.com/scylladb/scylla-ccm/blob/69118c7185e5b97a1d923084356f5b4dc4fbec8d/ccmlib/cmds/cluster_cmds.py#L832-L841
> > 
> > and
> > https://github.com/scylladb/scylla-ccm/blob/69118c7185e5b97a1d923084356f5b4dc4fbec8d/ccmlib/cmds/cluster_cmds.py#L872-L875
> > 
> > or do we accept `foo: [1, False, "hello"]` ?
> 
> this should work

it works with this patch. but again, do we accept it? or put another way, do we expect it?

> 
> > actually, i was about to suggest using `json.loads()` but i just realized that we are not actually expecting json encoded data here. it's kind of freestyle format. but i still think it would be better to stick to a well-defined format / encoding, so we don't need to worry about these things.
> > but if we don't really care. i'd suggest at least handle more types of exceptions, see https://docs.python.org/3/library/ast.html#ast.literal_eval
> 
> I rather not handle those, let it explode, i.e. if user put in a giant literal or that complex one, that it would hit recursion limit, I don't want to handle it gracefully, and silently put it as a string (which is probably never what he would expect in this case).

if we want to fail the test or whatever using ccm if user feeds it with something malformed, why are we handling `ValueError`? because it is a more typical use case?

tchaikov: > > > I think
> > > > what if we have something like: `hello: "world"` here? before this change, the `val` is `'"world"'`, after this change the `val` would be `'world'`. and what if we really want to set a value like `'1e10'`. i still cannot tell if this is a valid concern after reading
> > > 
> > > 
> > > regarding the example `hello: "world"` in yaml it's equivalent to `hello: world`, so in less you'll want something complex like multi line string, this should be o.k.
> > 
> > 
> > how does yaml come into play? are we consuming `val` later with `yaml.load()` ?
> 
> no, if we put it as string in a yaml file, with or without the quotes, it would be parsed the same, the code at scylla end gonna get it without quotes (like any yaml parser would).

i am trying to understand what we are expecting. as there is no canonical document on this. if we want to parse a yaml, then we should go with `yaml.load()` instead of trying to parse it as a python literal. it's both misleading and more importantly, wrong.

> 
> > > 90% of the cases we or need a string or need a number in configurations this fix helps to handle the case we need a list (and it would also gonna work for mapping)
> > > > https://github.com/scylladb/scylla-ccm/blob/69118c7185e5b97a1d923084356f5b4dc4fbec8d/ccmlib/cmds/cluster_cmds.py#L832-L841
> > > > 
> > > > and
> > > > https://github.com/scylladb/scylla-ccm/blob/69118c7185e5b97a1d923084356f5b4dc4fbec8d/ccmlib/cmds/cluster_cmds.py#L872-L875
> > > > 
> > > > or do we accept `foo: [1, False, "hello"]` ?
> > > 
> > > 
> > > this should work
> > 
> > 
> > it works with this patch. but again, do we accept it? or put another way, do we expect it?
> 
> it's a valid python list, it's a valid json, and would map to valid yaml, I see no reason to block it.
> 
> > > > actually, i was about to suggest using `json.loads()` but i just realized that we are not actually expecting json encoded data here. it's kind of freestyle format. but i still think it would be better to stick to a well-defined format / encoding, so we don't need to worry about these things.
> > > > but if we don't really care. i'd suggest at least handle more types of exceptions, see https://docs.python.org/3/library/ast.html#ast.literal_eval
> > > 
> > > 
> > > I rather not handle those, let it explode, i.e. if user put in a giant literal or that complex one, that it would hit recursion limit, I don't want to handle it gracefully, and silently put it as a string (which is probably never what he would expect in this case).
> > 
> > 
> > if we want to fail the test or whatever using ccm if user feeds it with something malformed, why are we handling `ValueError`? because it is a more typical use case?
> 
> because the the main case, passing some string, which isn't python literal. and not that I'm trying it, this main case of just passing a string is broken

fruch: > i am trying to understand what we are expecting. as there is no canonical document on this. if we want to parse a yaml, then we should go with yaml.load() instead of trying to parse it as a python literal. it's both misleading and more importantly, wrong.

There is a command line option to update configuration, 
passing a whole yaml structure doesn't yield itself to a command line very easily

Now it's the support configuration option which are string or int, now we need a list.

fruch: > i am trying to understand what we are expecting. as there is no canonical document on this. if we want to parse a yaml, then we should go with yaml.load() instead of trying to parse it as a python literal. it's both misleading and more importantly, wrong.

There is a command line option to update configuration, 
passing a whole yaml structure doesn't yield itself to a command line very easily

Now it's the support configuration option which are string or int, now we need a list.

tchaikov: > There is a command line option to update configuration,

i did read the code implementing  the configuration and referenced it in the previous comment. not sure if that's the one you are referring here?

```
In [1]: import yaml
In [2]: yaml.load('aa')
Out[2]: 'aa'
```

i think "aa" is also a yaml structure, and that's exactly we are passing to the command as the "val" in it.

fruch: I thought you are talking about parsing whole commandline as yaml...
but yes using `yaml.safe_load` instead of `ast.literal_eval`, works a bit better, since you don't need escaping.

I also added a unit test, to cover the different cases this should be supporting

tchaikov: thanks for adding the test. the reason that i prefer `yaml.safe_load` is that we are not loading a python literal. it's more correct, not just more convenient.

tests/test_common.py
```diff
@@ -69,3 +73,24 @@ def test_lockfile_retain_status_by_default():
...
     lf.release()
+
+def test_parse_settings():
+    res = parse_settings(["number:12", "string:somthing", "bool:false"])
+    assert res == {'bool': False, 'number': 12, 'string': 'somthing'}
+
+    res = parse_settings(["nested.number:12", "nested.string:somthing", "nested.bool:false"])
+    assert res == {'nested': {'bool': False, 'number': 12, 'string': 'somthing'}}
+
+    # a fix that need to be backported from upstream
```
tchaikov, line 84: how about filing an issue?

ccmlib/common.py
```diff
@@ -784,10 +781,7 @@ def parse_settings(args):
         elif val.lower() == "false":
             val = False
         else:
-            try:
-                val = int(val)
-            except ValueError:
-                pass
+            val = yaml.safe_load(str(val))
```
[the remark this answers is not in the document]
fruch: update those usages with what ?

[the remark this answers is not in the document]
fruch: done

tests/test_common.py
fruch: I've pasted there the commit, we don't really need it for anything (at least not yet)

tests/test_common.py
```diff
@@ -69,3 +73,24 @@ def test_lockfile_retain_status_by_default():
...
+    # res = parse_settings(["double.nested.number:12", "double.nested.string:somthing", "double.nested.bool:false"])
+    # assert res == {'double': {'nested': {'bool': False, 'number': 12, 'string': 'somthing'}}}
+
+    res = parse_settings(["experimental_features:[udf,tablets]"])
+    assert res == {'experimental_features': ['udf', 'tablets']}
+
+    res = parse_settings(["experimental_features:['udf',\"tablets\"]"])
+    assert res == {'experimental_features': ['udf', 'tablets']}
+
+    # would break if incorrect yaml format is pass in the value
```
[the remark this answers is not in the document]
fruch: fixed

ccmlib/common.py
```diff
@@ -16,6 +16,7 @@
 import tempfile
 import logging
 import pathlib
+import ast
```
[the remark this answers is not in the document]
fruch: remove

ccmlib/common.py
```diff
@@ -784,10 +780,7 @@ def parse_settings(args):
         elif val.lower() == "false":
             val = False
         else:
-            try:
-                val = int(val)
-            except ValueError:
-                pass
+            val = yaml.safe_load(str(val))
```
tchaikov, line 783: > no need to cast `val` to `str`.

this comment is not addressed. could you please reconsider it?

tests/test_common.py
tchaikov: okay, don't feel strong here. just thought that an issue is easier to track than a comment in code which is not even marked with `FIXME` or `TODO`.

ccmlib/common.py
fruch: removed

tchaikov (approved):
lgtm.

PR #492, opened 2023-08-07
scylla_extract_mode: fix regex pattern to match mode

bhalevy:
The current regular expression failed to match
the mode for e.g
`https://downloads.scylladb.com/unstable/scylla/master/relocatable/latest/scylla-debug-unified-5.4.0~dev-0.20230801.37b548f46365.x86_64.tar.gz`

See https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-debug/247/consoleText
```
[2023-08-06T05:45:02.790Z] SCYLLA_UNIFIED_PACKAGE=/jenkins/workspace/scylla-master/dtest-debug/scylla/build/debug/dist/tar/scylla-debug-unified-5.4.0~dev-0.20230801.37b548f46365.x86_64.tar.gz
```

https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-debug/247/artifact/logs-full.debug.052/dtest-gw0.log
```
07:54:08,036 790     errors                         ERROR    conftest.py         :208  | test_cluster_expansion_with_cdc[Single_cluster]: test failed:
...
            try:
>               self.wait_for_binary_interface(from_mark=from_mark, process=self._process_scylla, timeout=t)

../scylla/.local/lib/python3.11/site-packages/ccmlib/scylla_node.py:318:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../scylla/.local/lib/python3.11/site-packages/ccmlib/node.py:537: in wait_for_binary_interface
    self.watch_log_for("Starting listening for CQL clients", **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <ccmlib.scylla_node.ScyllaNode object at 0x7f4150514410>
exprs = 'Starting listening for CQL clients', from_mark = 0, timeout = 420
```

timeout should have been 900 seconds for debug mode if it was matched properly.

PR #516, opened 2023-10-16
scylla_cluster: load_from_repository: set default timeouts

bhalevy:
Commit 4f619b75ab048366f3a0f7971450d00485e910bc (scylladb/scylla-ccm#492)
Improved the regular expression used in `scylla_extract_mode` to better
detect the scylla_mode given the package url.

However, when we install from a local path, like
`/jenkins/workspace/scylla-master/dtest-debug/scylla/build/debug/dist/tar/scylla-debug-unified-5.4.0~dev-0.20231013.055f0617064d.x86_64.tar.gz`,
The regular expression fails to detect the mode since it falsely start the match at `/scylla-master`.

This change restricts the match to the last path component.

As seen in https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-debug/258/artifact/logs-full.debug.041/dtest-gw0.log
```
07:20:43,371 790     errors                         ERROR    conftest.py         :225  | test_cluster_expansion_with_cdc[Single_cluster]: test failed:
...
>               self.wait_for_binary_interface(from_mark=from_mark, process=self._process_scylla, timeout=t)
...
self = <ccmlib.scylla_node.ScyllaNode object at 0x7fbd89790f10>
exprs = 'Starting listening for CQL clients', from_mark = 0, timeout = 420
```

When `cassandra_version` is passed to `ScyllaCluster.__init__`
we set `self.scylla_mode = None` and the default timeouts are not
adjusted.

`scylla_mode` is set correctly later on in `load_from_repository`.
Call `__set_default_timeouts` again to adjust the timeout based
on the valid `scylla_mode`.

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -42,11 +42,12 @@ class ScyllaCluster(Cluster):
 
         self.started = False
         self.force_wait_for_cluster_start = (force_wait_for_cluster_start != False)
-        self.default_wait_other_notice_timeout = 120 if self.scylla_mode != 'debug' else 600
-        self.default_wait_for_binary_proto = 420 if self.scylla_mode != 'debug' else 900
+        self.__set_default_timeouts()
         self._scylla_manager = None
         self.skip_manager_server = skip_manager_server
         self.scylla_version = cassandra_version
+
+        self.debug(f"ScyllaCluster: cassandra_version={cassandra_version} docker_image={docker_image} install_dir={install_dir} scylla_mode={self.scylla_mode} default_wait_other_notice_timeout={self.default_wait_other_notice_timeout} default_wait_for_binary_proto={self.default_wait_for_binary_proto}")
 
         super(ScyllaCluster, self).__init__(path, name, partitioner,
                                             install_dir, create_directory,
@@ -71,7 +72,14 @@ class ScyllaCluster(Cluster):
     def load_from_repository(self, version, verbose):
         install_dir, version = scylla_repository.setup(version, verbose)
         install_dir, self.scylla_mode = common.scylla_extract_install_dir_and_mode(install_dir)
+        assert self.scylla_mode is not None
+        self.__set_default_timeouts()
+        self.debug(f"ScyllaCluster: load_from_repository: install_dir={install_dir} scylla_mode={self.scylla_mode} default_wait_other_notice_timeout={self.default_wait_other_notice_timeout} default_wait_for_binary_proto={self.default_wait_for_binary_proto}")
         return install_dir, version
+
+    def __set_default_timeouts(self):
+        self.default_wait_other_notice_timeout = 120 if self.scylla_mode != 'debug' else 600
+        self.default_wait_for_binary_proto = 420 if self.scylla_mode != 'debug' else 900
 
     # override get_node_jmx_port for scylla-jmx
     # scylla-jmx listens on the unique node address (127.0.<cluster.id><node.id>)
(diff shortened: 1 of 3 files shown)


fruch (approved):
LGTM

ccmlib/scylla_cluster.py
```diff
@@ -98,8 +98,7 @@ def start_nodes(self, nodes=None, no_wait=False, verbose=False, wait_for_binary_
             wait_for_binary_proto = self.force_wait_for_cluster_start
         if wait_other_notice is None:
             wait_other_notice = self.force_wait_for_cluster_start
-        self.debug(f"start_nodes: no_wait={no_wait} wait_for_binary_proto={wait_for_binary_proto} wait_other_notice={wait_other_notice} force_wait_for_cluster_start={self.force_wait_for_cluster_start}")
-        self.started=True
+        self.debug(f"start_nodes: no_wait={no_wait} wait_for_binary_proto={wait_for_binary_proto} wait_other_notice={wait_other_notice} force_wait_for_cluster_start={self.force_wait_for_cluster_start} scylla_mode={self.scylla_mode} default_wait_other_notice_timeout={self.default_wait_other_notice_timeout} default_wait_for_binary_proto={self.default_wait_for_binary_proto}")
```
fruch, line 101: don't we have enough prints showing the modes ?

why do we need it on each node starting ?

ccmlib/scylla_cluster.py
```diff
@@ -98,8 +98,7 @@ def start_nodes(self, nodes=None, no_wait=False, verbose=False, wait_for_binary_
             wait_for_binary_proto = self.force_wait_for_cluster_start
         if wait_other_notice is None:
             wait_other_notice = self.force_wait_for_cluster_start
-        self.debug(f"start_nodes: no_wait={no_wait} wait_for_binary_proto={wait_for_binary_proto} wait_other_notice={wait_other_notice} force_wait_for_cluster_start={self.force_wait_for_cluster_start}")
-        self.started=True
```
fruch, line 102: why did you remove this ?

fruch (changes requested):
return "self.started=True" back

ccmlib/scylla_cluster.py
bhalevy: this was inadvertent

bhalevy:
> return "self.started=True" back

done

bhalevy:
BTW setting self.started=True seems to be happening too early.
I'd consider doing that only before the function returns successfully

ccmlib/scylla_cluster.py
bhalevy: We have a printout in dtest_setup.py `init_default_config` and or printed the apparently correct mode:
```
06:51:00,723 790     dtest_setup                    DEBUG    dtest_setup.py      :694  | test_cluster_expansion_with_cdc[Single_cluster]: Scylla mode is 'debug'
```

But timeout==420 is an evidence for the mode not set up properly in `ScyllaCluster.__init__`

bhalevy: I can move the printout to the constructor so it's printed once per cluster instantiation.

bhalevy: or print the extra information only `if not self.started`

bhalevy:
Besides fixing this regex, I think that the issue with scylla_mode in `ScyllaCluster.__init__` could be this:
https://github.com/scylladb/scylla-ccm/blob/78b5d266dca3daa209406ab14265fe61479c51e5/ccmlib/scylla_cluster.py#L33-L35

While we pass a valid `cassandra_version` in this case:
https://github.com/scylladb/scylla-dtest/blob/a9b0676fc11eeab144d6d8d5fd402b4db6505dee/dtest_setup.py#L805-L809

bhalevy:
@fruch please re-review

bhalevy:
https://github.com/scylladb/scylla-ccm/commit/168057e1e344f7648b34930e87393e834d191112 also fixes https://github.com/scylladb/scylla-ccm/pull/516#issuecomment-1766222735 by setting the default timeouts again after ScyllaCluster.scylla_mode is correctly extracted in `load_from_repository`

fruch (approved):
LGTM

bhalevy:

PR #517, opened 2023-10-19
scylla_node: wait_for_compactions: move override to class ScyllaNode

bhalevy:
Change a1624b5c471c8653e3855e0bfa3ea06422e95f2a added the method to class `NodeUpgrader` instead of `ScyllaNode` by mistake so it is a no-op and `test_lcs_sstable_promotion` still fails in debug mode occasionally, like
https://jenkins.scylladb.com/view/master/job/scylla-master/job/dtest-debug/261/testReport/compaction_additional_test/TestLCSSSTablePromotion/Run_Dtest_Parallel_Cloud_Machines___FullDtest___full_split044___test_lcs_sstable_promotion/
```
ccmlib.node.TimeoutError: Waiting for compactions timed out after 300 seconds with 26 pending tasks remaining.
```

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1583,6 +1583,12 @@ class ScyllaNode(Node):
         stdout, _ = sstable_stats['']
         return json.loads(stdout)['sstables']
 
+    def wait_for_compactions(self, idle_timeout = None):
+        if idle_timeout is None:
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM

bhalevy:
@fruch can you please backport to all branches?
(And I cannot add labels, so we also need backport labels for 5.4 and soon, 2024.1)

Issue #518, opened 2023-10-20
Add support for populating multi-rack datacenter(s)


bhalevy:
Currently cluster.prepare supports 2 flavors of the `nodes` parameter:
1. An integer - specifying the number of nodes in a single datacenter cluster
2. A list of integers - specifying the number of nodes in each datacenter in a multi-dc cluster

For testing multi-rack datacenter I suggest adding a third notation:
A dictionary in the form of `{ dc{i}: nodes, ... }`
where nodes, like today, could specify the number of nodes in that dc as an integer (compatible with today's `[x, y, z, ...]` notation)
or they could be a list, where the would specify the number of nodes in each rack (racks automatically named as `RAC{i}` for backward compatibility)
or they could be a dictionary with `{ rac{i}: num_nodes, ... }` specifying the number of nodes in each named rack (integer only in this case)

bhalevy:
Cc @vladzcloudius

PR #519, opened 2023-10-20
cluster: prepare: support multi rack

bhalevy:
Support a `rack` member in `Node`
and allow specifying datacenter racks by extending the `nodes` parameter to `prepare` supported notations as follows:

    # nodes can be provided in multiple notations, determining the cluster topology:
    # 1. int - specifying the number of nodes in a single DC, single RACK cluster
    # 2. list[int] - specifying the number of nodes in a multi DC, single RACK per DC cluster
    #    The datacenters are automatically named as dc{i}, starting from 1, the rack is named RAC1
    # 3 dict[str, int|list|dict] - specifying the number of nodes in a multi DC, single or multi RACK per DC cluster
    #   The dictionary keys explicitly identify each datacenter name, and the value is either:
    #   a. int - the number of nodes in the DC.
    #   b. list[int] - the number of nodes in each rack in the datacenter.
    #   c. dict[str, int] - the number of nodes in each named rack in the datacenter.

The motivation for doing so is for adding dtests testing mult-rack topologies.

Refs https://github.com/scylladb/scylla-dtest/issues/3681
Fixes scylladb/scylla-ccm#518

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -244,6 +244,12 @@ class Cluster(object):
         node._save()
         return self
 
+    # nodes can be provided in multiple notations, determining the cluster topology:
+    # 1. int - specifying the number of nodes in a single DC, single RACK cluster
+    # 2. list[int] - specifying the number of nodes in a multi DC, single RACK per DC cluster
+    #    The datacenters are automatically named as dc{i}, starting from 1, the rack is named RAC1
+    #    For example, [3, 2] would translate to the following topology {'dc1': {'RAC1': 3}, 'dc2': {'RAC1': 2}}
+    #    Where 3 nodes are populated in dc1/RAC1, and 2 nodes are populated in dc2/RAC1
     def populate(self, nodes, debug=False, tokens=None, use_vnodes=False, ipprefix=None, ipformat=None):
         if ipprefix:
             self.ipprefix = ipprefix
@@ -253,21 +259,34 @@ class Cluster(object):
             self.ipformat = ipformat
         elif not self.ipformat:
             self.ipformat = self.ipprefix + "%d"
-        node_count = nodes
-        dcs = []
         self.use_vnodes = use_vnodes
-        if isinstance(nodes, list):
+        topology = OrderedDict()
+        if isinstance(nodes, int):
+            topology[None] = OrderedDict([(None, nodes)])
+        elif isinstance(nodes, list):
+            for i in range(0, len(nodes)):
+                dc = f"dc{i + 1}"
+                n = nodes[i]
+                topology[dc] = OrderedDict([(None, n)])
+        else:
+            raise common.ArgumentError(f'invalid nodes type {type(nodes)}: {nodes}')
+        node_count = 0
+        dcs = list(topology.keys())
+        node_locations = []
+        for dc, racks in topology.items():
+            assert dc is None or isinstance(dc, str)
+            for rack, n in racks.items():
+                assert rack is None or isinstance(rack, str)
+                assert isinstance(n, int)
+                node_count += n
+                for _ in range(n):
+                    node_locations.append((dc, rack))
+        if dcs != [None]:
             self.set_configuration_options(values={'endpoint_snitch': self.snitch})
-            node_count = 0
-            i = 0
-            for c in nodes:
-                i = i + 1
-                node_count = node_count + c
-                for x in range(0, c):
-                    dcs.append('dc%d' % i)
+        self.use_vnodes = use_vnodes
 
         if node_count < 1:
-            raise common.ArgumentError(f'invalid node count {nodes}')
+            raise common.ArgumentError(f'invalid topology {topology}')
 
         for i in range(1, node_count + 1):
             if f'node{i}' in list(self.nodes.values()):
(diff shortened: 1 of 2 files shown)


ccmlib/cluster.py
```diff
@@ -646,14 +682,14 @@ def __update_topology_files(self):
             self.__update_topology_using_rackdc_properties()
 
     def __update_topology_using_toplogy_properties(self):
-        dcs = [('default', 'dc1')]
+        dcs = [('default', 'dc1', 'r1')]
```
fruch, line 685: Why r1 and not RAC1 ?

It contradicts the default in other places of the code.

Also we should be careful not to change this default, it affects multiple users of CCM, not just dtest.

fruch:
I could guess the reason we want it,

But it would be nice to mention why we need this change.

(And why the way we do it in dtest currently isn't good enough)

fruch:
> I could guess the reason we want it,
> 
> But it would be nice to mention why we need this change.
> 
> (And why the way we do it in dtest currently isn't good enough)
> 

Now I've seen the issue asking for this, but we should reference it.

ccmlib/cluster.py
```diff
@@ -646,14 +682,14 @@ def __update_topology_files(self):
...
         for node in self.nodelist():
             if node.data_center is not None:
-                dcs.append((node.address(), node.data_center))
+                dcs.append((node.address(), node.data_center, node.rack or 'r1'))
 
         content = ""
-        for k, v in dcs:
-            content = f"{content}{k}={v}:r1\n"
```
bhalevy, line 656: This is the reason I used `r1`, for backward compatibility.

ccmlib/cluster.py
bhalevy: See my comment below

soyacz: anayway it is used for `PropertyFileSnitch` snitch (don't mix with `GossipingPropertyFileSnitch`) - so it is never used anyway.

ccmlib/cluster.py
```diff
@@ -251,18 +262,43 @@ def populate(self, nodes, debug=False, tokens=None, use_vnodes=False, ipprefix=N
...
+        elif isinstance(nodes, list):
             self.set_configuration_options(values={'endpoint_snitch': self.snitch})
             node_count = 0
             i = 0
             for c in nodes:
                 i = i + 1
                 node_count = node_count + c
                 for x in range(0, c):
-                    dcs.append('dc%d' % i)
+                    dcs.append(('dc%d' % i, None))
```
soyacz, line 277: dcs append here 3 element tuple, while below 2 element

ccmlib/cluster.py
```diff
@@ -251,18 +262,43 @@ def populate(self, nodes, debug=False, tokens=None, use_vnodes=False, ipprefix=N
...
+        elif isinstance(nodes, dict):
```
soyacz, line 278: another approach could be always normalize input to `dict[str, dict]` form and cover only that case - could improve readability how result cluster will look and reduce some complexity.

soyacz (commented):
One thing to fix and some comments

ccmlib/cluster.py
bhalevy: how 3?
`'dc%d' % i` and `None`

ccmlib/cluster.py
bhalevy: it's possible, but it will break compatibility with todays's default case of rack=None.
It"s nor so terrible, but I was careful not change anything that might break tests or applications using this code

ccmlib/cluster.py
soyacz: I confused, sorry

bhalevy:
In https://github.com/scylladb/scylla-ccm/commit/bbfa12bdf289ffa9dd12ec07a5d36a08a77fc2ec:
- use OrderedDict to represent the topology in a canonical way

bhalevy:
fixed CI error

ccmlib/cluster.py
```diff
@@ -238,10 +239,22 @@ def add(self, node, is_seed, data_center=None):
...
+    # 1. int - specifying the number of nodes in a single DC, single RACK cluster
+    # 2. list[int] - specifying the number of nodes in a multi DC, single RACK per DC cluster
+    #    The datacenters are automatically named as dc{i}, starting from 1, the rack is named RAC1
+    # 3.a dict[str, int] - specifying the number of nodes in a multi DC, single RACK per DC cluster
+    #     The dictionary keys explicitly identify each datacenter name, and the value is the number of nodes in the DC.
+    # 3.b dict[str, list[int]] - specifying the number of nodes in a multi, multi RACK cluster
+    #     The dictionary keys explicitly identify each datacenter name, and the value is the number of nodes in each RACK in the DC.
+    #     Racks are automatically named as RAC{i}, starting from 1
+    # 3.c dict[str, dict[str, int]] - specifying the number of nodes in a multi DC, multi RACK cluster
+    #     The dictionary keys explicitly identify each datacenter and rack names and the values are the number of nodes in each RACK in the DC.
```
soyacz, line 257: would be nice to have examples somewhere as `dict[str, dict[str, int]]` doesn't look friendly

bhalevy: added examples in next version

soyacz (approved):
LGTM

fruch (approved):
LGTM

PR #523, opened 2023-11-08
scylla_node: watch_rest_for_alive: wait for others to be considered normal token owners

bhalevy:
It is not enough for node to know about other nodes' tokens, as they might not be reflected in the token_metadata map.

Instead, check the `/storage_service/host_id` api that provides a list of nodes that are normal token owners and ready to be used by queries.

Refs https://github.com/scylladb/scylladb/issues/15146

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -100,13 +100,15 @@ class ScyllaCluster(Cluster):
             node._update_pid(p)
 
     def start_nodes(self, nodes=None, no_wait=False, verbose=False, wait_for_binary_proto=None,
-              wait_other_notice=None, jvm_args=None, profile_options=None,
+              wait_other_notice=None, wait_normal_token_owner=None, jvm_args=None, profile_options=None,
               quiet_start=False):
         if wait_for_binary_proto is None:
             wait_for_binary_proto = self.force_wait_for_cluster_start
         if wait_other_notice is None:
             wait_other_notice = self.force_wait_for_cluster_start
-        self.debug(f"start_nodes: no_wait={no_wait} wait_for_binary_proto={wait_for_binary_proto} wait_other_notice={wait_other_notice} force_wait_for_cluster_start={self.force_wait_for_cluster_start}")
+        if wait_normal_token_owner is None and wait_other_notice:
+            wait_normal_token_owner = True
+        self.debug(f"start_nodes: no_wait={no_wait} wait_for_binary_proto={wait_for_binary_proto} wait_other_notice={wait_other_notice} wait_normal_token_owner={wait_normal_token_owner} force_wait_for_cluster_start={self.force_wait_for_cluster_start}")
         self.started=True
 
         p = None
@@ -137,7 +139,8 @@ class ScyllaCluster(Cluster):
                 p = node.start(update_pid=False, jvm_args=jvm_args,
                                profile_options=profile_options, no_wait=no_wait,
                                wait_for_binary_proto=wait_for_binary_proto,
-                               wait_other_notice=wait_other_notice)
+                               wait_other_notice=wait_other_notice,
+                               wait_normal_token_owner=False)
                 started.append((node, p, mark))
                 marks.append((node, mark))
 
@@ -156,7 +159,8 @@ class ScyllaCluster(Cluster):
             for old_node, _ in marks:
                 for node, _, _ in started:
                     if old_node is not node:
-                        old_node.watch_rest_for_alive(node, timeout=self.default_wait_other_notice_timeout)
+                        old_node.watch_rest_for_alive(node, timeout=self.default_wait_other_notice_timeout,
+                                                      wait_normal_token_owner=wait_normal_token_owner)
 
         return started
 
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1364,10 +1364,10 @@ class ScyllaNode(Node):
             while time.time() < endtime:
                 live = set()
                 response = requests.get(url=url_live)
-                if response.text:
+                if response.status_code == requests.codes.ok:
                     live = set(response.json())
                 response = requests.get(url=url_joining)
-                if response.text:
+                if response.status_code == requests.codes.ok:
                     live = live - set(response.json())
                 # Verify that node knows not only about the existance of the
                 # other node, but also its tokens:
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1348,18 +1348,22 @@ class ScyllaNode(Node):
     def watch_rest_for_alive(self, nodes, timeout=120):
         """
         Use the REST API to wait until this node detects that the nodes listed
-        in "nodes" become fully operational and knows of its tokens.
+        in "nodes" become fully operational as normal token owners.
         This is similar to watch_log_for_alive but uses ScyllaDB's REST API
         instead of the log file and waits for the node to be really useable,
         not just "UP" (see issue #461)
         """
         logging.getLogger('urllib3.connectionpool').disabled = True
         try:
-            tofind = nodes if isinstance(nodes, list) else [nodes]
-            tofind = set([node.address() for node in tofind])
+            nodes_tofind = nodes if isinstance(nodes, list) else [nodes]
+            tofind = set([node.address() for node in nodes_tofind])
+            tofind_host_id_map = dict([(node.address(), node.hostid()) for node in nodes_tofind])
+            found = set()
+            found_host_id_map = dict()
             url_live = f"http://{self.address()}:10000/gossiper/endpoint/live"
             url_joining = f"http://{self.address()}:10000/storage_service/nodes/joining"
             url_tokens = f"http://{self.address()}:10000/storage_service/tokens/"
+            url_host_ids = f"http://{self.address()}:10000/storage_service/host_id"
             endtime = time.time() + timeout
             while time.time() < endtime:
                 live = set()
@@ -1370,20 +1374,37 @@ class ScyllaNode(Node):
                 if response.status_code == requests.codes.ok:
                     live = live - set(response.json())
                 # Verify that node knows not only about the existance of the
-                # other node, but also its tokens:
+                # other node, but also its host_id as a normal token owner:
                 if tofind.issubset(live):
                     # This node thinks that all given nodes are alive and not
                     # "joining", we're almost done, but still need to verify
                     # that the node knows the others' tokens.
                     check = tofind
-                    tofind = set()
+                    have_no_tokens = set()
                     for n in check:
                         response = requests.get(url=url_tokens+n)
                         if response.text == '[]':
-                            tofind.add(n)
-                if not tofind:
-                    return
+                            have_no_tokens.add(n)
+                    if not have_no_tokens:
+                        # and that the node knows that the others' are normal token owners.
+                        host_id_map = dict()
+                        response = requests.get(url=url_host_ids)
+                        if response.status_code == requests.codes.ok:
+                            for r in response.json():
+                                host_id_map[r['key']] = r['value']
+                        # Verify that the other nodes are considered normal token owners on this node
+                        # and their host_id matches the host_id the client knows about
+                        normal = set([addr for addr, id in host_id_map.items() \
+                                    if addr in tofind_host_id_map and \
+                                    (id == tofind_host_id_map[addr] or not tofind_host_id_map[addr])])
+                        tofind = tofind.difference(normal)
+                        if not tofind:
+                            return
+                        # Update cumulative maps for debugging
+                        found = found.union(normal)
+                        found_host_id_map.update(host_id_map)
                 time.sleep(0.1)
+            self.debug(f"watch_rest_for_alive: tofind={tofind} found={found}: tofind_host_id_map={tofind_host_id_map} found_host_id_map={found_host_id_map}")
             raise TimeoutError(f"watch_rest_for_alive() timeout after {timeout} seconds")
         finally:
             logging.getLogger('urllib3.connectionpool').disabled = False
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -284,7 +284,8 @@ class ScyllaNode(Node):
             time.sleep(sleep_time)
         return bool(self.grep_log(f"{bootstrap_message}|{resharding_message}", from_mark=from_mark))
 
-    def _start_scylla(self, args, marks, update_pid, wait_other_notice,
+    def _start_scylla(self, args, marks, update_pid,
+                      wait_other_notice, wait_normal_token_owner,
                       wait_for_binary_proto, ext_env):
         log_file = os.path.join(self.get_path(), 'logs', 'system.log')
         # In case we are restarting a node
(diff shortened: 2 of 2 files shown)


ccmlib/scylla_node.py
```diff
@@ -1333,18 +1333,19 @@ def rollback(self, upgrade_to_version):
...
         instead of the log file and waits for the node to be really useable,
         not just "UP" (see issue #461)
         """
         logging.getLogger('urllib3.connectionpool').disabled = True
         try:
-            tofind = nodes if isinstance(nodes, list) else [nodes]
-            tofind = set([node.address() for node in tofind])
+            nodes_tofind = nodes if isinstance(nodes, list) else [nodes]
+            tofind = set([node.address() for node in nodes_tofind])
+            tofind_host_id_map = dict([(node.address(), node.node_hostid if hasattr(self, 'node_hostid') else None) for node in nodes_tofind])
```
bhalevy, line 1345: I think we can retrieve the host_id via the api if it's yet available (the ccmlib node.py method uses nodetool that is it ready yet since scylla-jmx was not started yet)

bhalevy:
@nyh, @fruch please review

ccmlib/scylla_node.py
```diff
@@ -1330,45 +1330,69 @@ def upgrade(self, upgrade_to_version):
...
 
+    def hostid(self, timeout=60, force_refresh=False):
+        if hasattr(self, 'node_hostid') and not force_refresh:
+            return self.node_hostid
+        node_address = self.address()
+        url = f"http://{node_address}:10000/storage_service/hostid"
+        response = requests.get(url=url)
+        if response.status_code == requests.codes.ok:
+            self.debug(f"/storage_service/hostid: {response.json()}")
+            for r in response.json():
```
fruch, line 1341: This API can return multiple host_ids ?

On which cases ?

fruch:
Seems o.k.

But I'll want to take a ride in at least in gating

fruch:
@bhalevy 

giving it a go:
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/134/ (OSS)
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/135/ (OSS + raft)
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/136/ (Enterprise)
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/137/ (Enterprise + raft)

ccmlib/scylla_node.py
bhalevy: It's name is misleading, it usually returns multiple endpoints/hostsids, of all nodes which are normal token owners, unless it's a single node cluster.

fruch: a bit sad there's no API for getting this node host_id...

fruch:
> @bhalevy
> 
> giving it a go:
> 
> * [ ]  https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/134/ (OSS)
> * [ ]  https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/135/ (OSS + raft)
> * [ ]  https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/136/ (Enterprise)
> * [ ]  https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/137/ (Enterprise + raft)

@bhalevy seems like we are getting None as host_id all over the place

ccmlib/scylla_node.py
bhalevy: Oops, I looked at the jmx code and there is one I forgot about (that being used by `nodetool info` indirectly via jmx.

bhalevy: fixed in v3

bhalevy:
> @bhalevy seems like we are getting None as host_id all over the place

@fruch where?  can you please give a concrete example?

fruch:
> > @bhalevy seems like we are getting None as host_id all over the place
> 
> should be fixed now (at least it passes for me locally, for example `update_cluster_layout_tests.py::TestUpdateClusterLayout::test_simple_removenode_3[api-True]`)

running gating again:
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/138/ (OSS)
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/139/ (OSS + raft)
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/140/ (Enterprise)
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/141/ (Enterprise + raft)

bhalevy:
- scylla_node: start_scylla: reset node_hostid

so to force reload of the host_id after restart as the node may wake up with a different hsot_id (after being wiped)

bhalevy:
https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/142/

bhalevy:
Depends on https://github.com/scylladb/scylla-dtest/pull/3716

bhalevy:
In [54332a9](https://github.com/scylladb/scylla-ccm/commit/54332a98fa0ebc4616a1615af338eb2164afb35d):
- scylla_node: hostid: handle errors
- scylla_node: watch_rest_for_alive: check response.status_code
- scylla_node: watch_rest_for_alive: wait for others to be considered normal token owners:
  - extends current function by waiting on tokens first, then on nrmal token ownership
- scylla_node: watch_rest_for_alive: add wait_normal_token_owner param
  - for backward compatibility
  - applied also from ScyllaCluster::start_nodes

fruch:
giving one last run, this is some that gonna happen on every single test:

- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/143/ (OSS)
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/144/ (OSS + raft)
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/145/ (Enterprise)
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/146/ (Enterprise + raft)

@bhalevy also keep in mind it might break non gating tests...
we'll need to keep that in mind following this merge

bhalevy:
> @bhalevy also keep in mind it might break non gating tests...
> we'll need to keep that in mind following this merge

Right. That's why I added the `wait_normal_token_owner`.

fruch:
> > @bhalevy also keep in mind it might break non gating tests...
> > we'll need to keep that in mind following this merge
> 
> Right. That's why I added the `wait_normal_token_owner` flag.

And which tests are gonna use that ? Only one we will be unstable?

bhalevy:
> > > @bhalevy also keep in mind it might break non gating tests...
> > > we'll need to keep that in mind following this merge
> > 
> > 
> > Right. That's why I added the `wait_normal_token_owner` flag.
> 
> And which tests are gonna use that ? Only one we will be unstable?

Yes.  This category of tests monitor starting nodes' log looking for particular events.
In most cases we already call start with `wait_other_notice=False, wait_for_binary_protocol=False`, but in cases we didn't do that we'll have to figure out what's the right thing to do.

bhalevy:
> giving one last run, this is some that gonna happen on every single test:
> 
> * [ ]  https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/143/ (OSS)
> * [ ]  https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/144/ (OSS + raft)
> * [ ]  https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/145/ (Enterprise)
> * [ ]  https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/146/ (Enterprise + raft)

There's only https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/146/testReport/junit/cql_tests/TestTruncate/FullDtest___full_split000___test_cql_query_filtering_without_indexes/
which seems unrelated.  Need to analyse it further

bhalevy:
This is a test issue: https://github.com/scylladb/scylla-dtest/issues/3717

bhalevy:
@nyh since you're the last to change `watch_rest_for_alive`, any comments before we merge this change?

fruch (approved):
LGTM

bhalevy:
@nyh please review (since you were the last one to change this area)

ccmlib/scylla_node.py
```diff
@@ -316,6 +316,8 @@ def _start_scylla(self, args, marks, update_pid, wait_other_notice,
                 raise NodeError(f"Error starting node {self.name}",
                                 self._process_scylla)
 
+        self.node_hostid = None
```
nyh, line 320: Nitpick: Would have been useful to get the explanation in the commit message as a comment, not (just) as a commit message.

ccmlib/scylla_node.py
```diff
@@ -1336,12 +1336,16 @@ def rollback(self, upgrade_to_version):
...
+        try:
+            node_address = self.address()
+            url = f"http://{node_address}:10000/storage_service/hostid/local"
+            response = requests.get(url=url)
+            if response.status_code == requests.codes.ok:
+                self.node_hostid = response.json()
+                return self.node_hostid
+        except Exception as e:
+            self.error(f"Failed to get hostid using {url}: {e}")
+            pass
```
nyh, line 1355: ‎This "pass" isn't needed.

ccmlib/scylla_node.py
```diff
@@ -1373,20 +1377,33 @@ def watch_rest_for_alive(self, nodes, timeout=120):
...
                 # Verify that node knows not only about the existance of the
-                # other node, but also its tokens:
+                # other node, but also its host_id as a normal token owner:
                 if tofind.issubset(live):
                     # This node thinks that all given nodes are alive and not
                     # "joining", we're almost done, but still need to verify
                     # that the node knows the others' tokens.
                     check = tofind
-                    tofind = set()
+                    have_no_tokens = set()
```
nyh, line 1393: Nitpick: The original code, after checking that node X does have tokens, removed it from the "tofind" list, so no need to check it again in the next iteration. Your rewrite lost this optimization. But I agree it's not a very important optimization.

ccmlib/scylla_cluster.py
```diff
@@ -100,13 +100,15 @@ def __update_pids(self, started):
             node._update_pid(p)
 
     def start_nodes(self, nodes=None, no_wait=False, verbose=False, wait_for_binary_proto=None,
-              wait_other_notice=None, jvm_args=None, profile_options=None,
+              wait_other_notice=None, wait_normal_token_owner=None, jvm_args=None, profile_options=None,
```
nyh, line 103: To be honest, I don't understand why anybody will want to set this wait_normal_token_owner thing - in what situation would you want not to wait?
And I don't understand how anyone would ever understand what it means without a comment explaining it.
Also, why is the default "None" and not True or False??

nyh: I see below that you have in start_node() a comment about this option. My other questions on it still stand.

nyh (changes requested):

ccmlib/scylla_cluster.py
bhalevy: > To be honest, I don't understand why anybody will want to set this wait_normal_token_owner thing - in what situation would you want not to wait?

It's for backward compatibility purposes in case this change breaks existing dtests that rely on the previous timing (like they need to inject some even after they node is noticed as UP and before its considered a normal token owner.

This was the case I fixed in https://github.com/scylladb/scylla-dtest/pull/3716.
We can follow these lines also in other test cases and just pass `wait_other_notice=False`, but I think this option would provide a faster path that will just fallback to the existing behavior.

> And I don't understand how anyone would ever understand what it means without a comment explaining it.

I'll spread the doc comment around more broadly.

> Also, why is the default "None" and not True or False??

Like the oter args, `None` let's the test know that it is unset, and then it is calculated in the function body based on the state of other parameters.  It's hard to achieve that with literal default initialization.

ccmlib/scylla_node.py
bhalevy: True, but I find it clarifying, as it documents the fact that the error is ignored on purpose.
Otherwise, a naive reader might suspect there is a missing `raise`.

ccmlib/scylla_node.py
```diff
@@ -1330,6 +1331,17 @@ def upgrade(self, upgrade_to_version):
...
+    def hostid(self, timeout=60, force_refresh=False):
+        if self.node_hostid and not force_refresh:
+            return self.node_hostid
+        node_address = self.address()
+        url = f"http://{node_address}:10000/storage_service/hostid/local"
+        response = requests.get(url=url)
+        if response.status_code == requests.codes.ok:
+            self.node_hostid = response.json()
+            return self.node_hostid
+        return None
```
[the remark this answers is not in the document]
bhalevy: Hmm, the timeout parameter is inherited from the default implementation of this method in the Node class.
(Unfortunately there is no `override` keyword in Python, that I know of).
There, it is passed to the `nodetool` call, but here it has no purpose.
Still, we're required to implement the same signature to override the method.

ccmlib/scylla_node.py
bhalevy: added in next version

ccmlib/scylla_node.py
bhalevy: dropped `pass` in next version

ccmlib/scylla_node.py
bhalevy: I restored the trimming of `tofind` in the next version (in a different way)

bhalevy:
In [5fb352a](https://github.com/scylladb/scylla-ccm/commit/5fb352ad877b3b66f0850bee903b58c4eae3eaf5):
- scylla_node: override hostid() using the REST api
  - used timeout arg for requests.get
- scylla_node: start_scylla: reset node_hostid
  - added code comment explaining why we do that
- scylla_node: hostid: handle errors
  - dropped `pass` statement
- scylla_node: watch_rest_for_alive: wait for others to be considered normal token owners
  - trim `tofind` as nodes are found
  - but maintain cummulative `found` set and `found_host_id_map` dict for debugging purposes
- scylla_node: watch_rest_for_alive: add wait_normal_token_owner param
  - Document `wait_normal_token_owner` param also in `watch_rest_for_alive`

bhalevy:
> @nyh please re-review. I addressed your review comments.

@nyh ping, when do you think you can re-review?

ccmlib/scylla_node.py
```diff
@@ -1373,20 +1377,37 @@ def watch_rest_for_alive(self, nodes, timeout=120):
...
+                                host_id_map[r['key']] = r['value']
+                        # Verify that the other nodes are considered normal token owners on this node
+                        # and their host_id matches the host_id the client knows about
+                        normal = set([addr for addr, id in host_id_map.items() \
+                                    if addr in tofind_host_id_map and \
+                                    (id == tofind_host_id_map[addr] or not tofind_host_id_map[addr])])
+                        tofind = tofind.difference(normal)
+                        if not tofind:
+                            return
+                        # Update cummulative maps for debugging
```
nyh, line 1418: spelling error: cummulative

ccmlib/scylla_cluster.py
nyh: I remembered the None thing is a Python convention, you're right.

I'm not a fan of dtest's approach of having convenience functions that take 10 parameters that nobody understand instead of just "doing the right thing" but I agree that functionally, the extra parameter doesn't hurt. So if the dtest developers are fine with it, then so be it.

nyh (approved):

bhalevy:
- fixed spelling mistake and rebased
@fruch please merge when CI passed

bhalevy:
All checks have passed, please merge

bhalevy:
@fruch can you please backport to 5.4 and 2024.1?

mykaul:
This kinda reminds a (very) long discussion with the operator team, where they did not know when can they safely do a rolling restart of Scylla pods - and the request was a simple API to let them know 'yes, it's real, this node is really up and serving traffic and is part of the cluster and you can safely move to the next node and restart it'.
See https://github.com/scylladb/scylladb/issues/8275

PR #525, opened 2023-11-12
switch to use ruamel.yaml

fruch:
since ruamel.yaml can preserve yaml comment
so we can have the original yaml files
of scylla

### Testing

- [x] :clock1: https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/462/ (Gossip - OSS)
- [x] :clock1: https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/463/ (Gossip - Enterprise)
- [x] :clock1: https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/464/ (Raft - OSS)
- [x] :clock1: https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/465/ (Raft - Enterprise)
- [x] :clock1: https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/466/ (Tablets - OSS)
- [x] :clock1: https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/467/ (Tablets - Enterprise)

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -8,13 +8,15 @@ import time
 from collections import OrderedDict, defaultdict
 from concurrent.futures import ThreadPoolExecutor
 
-import yaml
+from ruamel.yaml import YAML
 
 from ccmlib import common, repository
 from ccmlib.node import Node, NodeError
 from ccmlib.common import logger
 from ccmlib.utils.version import parse_version
 
+yaml = YAML()
+yaml.default_flow_style = False
 
 class Cluster(object):
 
@@ -675,7 +677,7 @@ class Cluster(object):
             cluster_config['sni_proxy_listen_port'] = self.sni_proxy_listen_port
 
         with open(filename, 'w') as f:
-            yaml.safe_dump(cluster_config, f)
+            yaml.dump(cluster_config, f)
 
     def __update_pids(self, started):
         for node, p, _ in started:
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -17,12 +17,15 @@ from datetime import datetime
 import locale
 from collections import namedtuple
 
-import yaml
+from ruamel.yaml import YAML
 
 from ccmlib import common
 from ccmlib.cli_session import CliSession
 from ccmlib.repository import setup
 from ccmlib.utils.version import parse_version
+
+yaml = YAML()
+yaml.default_flow_style = False
 
 class Status():
     UNINITIALIZED = "UNINITIALIZED"
@@ -135,7 +138,7 @@ class Node(object):
         node_path = os.path.join(path, name)
         filename = os.path.join(node_path, 'node.conf')
         with open(filename, 'r') as f:
-            data = yaml.safe_load(f)
+            data = yaml.load(f)
         try:
             itf = data['interfaces']
             initial_token = None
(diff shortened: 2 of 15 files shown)


ccmlib/cluster.py
```diff
@@ -8,13 +8,15 @@
 from collections import OrderedDict, defaultdict
 from concurrent.futures import ThreadPoolExecutor
 
-import yaml
+from ruamel.yaml import YAML
```
tchaikov, line 11: IIUC, a typical use case of preserving the comment in YAML file is when care about comment of an existing yaml file, but why would we care about the comment? do we read read the edited yaml file, or do we plan to use some script to extract the comments out of the edited yaml file and process them?

tchaikov: i see. you are using ccm as a testbed for this feature we would be interested in, before replacing `yaml` with `ruamel.yam` in scylla-machine-image. but i don't understand how the test drive in this project helps to build the confidence of using it in another. is scylla-ccm sharing the same pipeline with scylla-machine-image?

i don't feel strong either way, as long as functionality wise, `ruamel.yam` is a drop-in replacement of `yaml`.

ccmlib/node.py
```diff
@@ -19,13 +19,16 @@
...
 
-import yaml
+from ruamel.yaml import YAML
 
 from ccmlib import common
 from ccmlib.cli_session import CliSession
 from ccmlib.repository import setup
 from ccmlib.utils.version import parse_version
 
+yaml = YAML()
```
soyacz, line 29: safe_load was used, shouldn't then add `yaml=YAML(typ='safe') `?

fruch: maybe we could, not sure it's really critical to this project

`safe_load` was used cause pyyaml was issues warning on not using it.

soyacz (approved):
LGTM

ccmlib/cluster.py
fruch: it is a drop in replacement, 
and the main reason for this one, is to align all tests infrastructure to be able to return scylla.yaml as is, as much as possible.

it's doesn't have too much value on dtest or other users of ccm, since we don't even save those fails as artifacts (even that we probably should)

tchaikov (approved):
i'd say it's one of steps to migrate our CI/CD from `yaml` to `ruamel.yaml`.  what if the trial failed, shall we revert the changes? i don't know =)

fruch:
> i'd say it's one of steps to migrate our CI/CD from `yaml` to `ruamel.yaml`. what if the trial failed, shall we revert the changes? i don't know =)

yes we can, but this change on it's own shouldn't be breaking anything (as far as I know)
i.e. I did tested it with dtest (and it was passing CCM CI tests)

if it proven to causing issue, we'll consider/handle it as in any other change.

and every trial, start with one small step, this PR is small enough :)

cezarmoise:
maybe the readme needs to be updated as well
https://github.com/scylladb/scylla-ccm/blob/076a4068c879da661cdda5df0e623a211e5b7282/README.md?plain=1#L59

fruch:
> maybe the readme needs to be updated as well
> 
> https://github.com/scylladb/scylla-ccm/blob/076a4068c879da661cdda5df0e623a211e5b7282/README.md?plain=1#L59

that readme probably need rewriting anyhow, I touch it about in that requirement section, which is mostly isn't needed anymore now

those are from days you could install ccm as a package, now you can and it's requirements are set inside `setup.py`

PR #526, opened 2023-11-13
ccmlib/scylla_node: pass `scylla.yaml` to `scylla sstable`

fruch:
* since we want to use this for encrypted sstable, we need to pass in the scylla.yaml

* since some of the output has binary data inside it, we need to make the call using bytes and not text

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1384,7 +1384,7 @@ class ScyllaNode(Node):
                 return str(candidate)
         raise ValueError(f"gnutls.config wasn't found in any path: {candidates}")
 
-    def run_scylla_sstable(self, command, additional_args=None, keyspace=None, datafiles=None, column_families=None, batch=False):
+    def run_scylla_sstable(self, command, additional_args=None, keyspace=None, datafiles=None, column_families=None, batch=False, text=True):
         """Invoke scylla-sstable, with the specified command (operation) and additional_args.
 
         For more information about scylla-sstable, see https://docs.scylladb.com/stable/operating-scylla/admin-tools/scylla-sstable.html.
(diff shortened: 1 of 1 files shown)


ccmlib/scylla_node.py
```diff
@@ -1404,6 +1404,14 @@ def run_scylla_sstable(self, command, additional_args=None, keyspace=None, dataf
...
         if additional_args is None:
             additional_args = []
+
+        additional_args += ['--scylla-yaml-file', str(Path(self.get_conf_dir()) / common.SCYLLA_CONF)]
+        if keyspace:
+            additional_args += ['--keyspace', keyspace]
+
+        if column_families and len(column_families) == 1:
+            additional_args += ['--table', column_families[0]]
+
```
tchaikov, line 1414: i'd suggest create yet another wrapper method around `run_scylla_sstable()` and pass `addition_args` to with the expected argument instead, like

```py
def dump_sstable_scylla_metadata(self, keyspace, column_family):
    # ...
    additional_args = ['--keyspace', keyspace,
                       '--table', column_family]
    scylla_metadata = self.run_scylla_sstable('dump-scylla-metadata',
                                               additional_args,
                                               keyspace=keyspace,
                                               column_families=[column_family])
```

fruch: including the yaml file ?

fruch: I've moved it to a new function, to minimize the changes to the other functions

denesb (approved):

ccmlib/scylla_node.py
```diff
@@ -1583,6 +1585,41 @@ def dump_sstable_stats(self,
         stdout, _ = sstable_stats['']
         return json.loads(stdout)['sstables']
 
+    def dump_sstable_scylla_metadata(self,
+                           keyspace: str,
+                           column_family: str,
+                           datafiles: Optional[List[str]] = None) -> List[Dict[str, Any]]:
+        """dump scylla metadate `scylla sstable dump-scylla-metadata`
```
tchaikov, line 1592: s/metadate/metadata/

fruch: fixed

PR #530, opened 2023-11-21
scylla_repository: expect -SNAPSHOT in scylla-manager reloc URL

tchaikov:
it turns out the latest scylla-manager's reloc tarball's URL contains contains "-SNAPSHOT" instead of "-dev-" in it. so let's update the test to reflect the fact.

the name should match with the name specified by
scylla-manager/dist/.goreleaser.yaml 's `name_template`.


fruch (approved):
Lgtm

Maybe we should check for both ?

tchaikov:
> Lgtm
> 
> Maybe we should check for both ?

i'd prefer not. IMHO, the uncertainty in test is a source of the confusion. that's why i also checked scylla-manager's CI.

PR #532, opened 2023-11-24
ccmlib/scylla_node: add ks, cf params to wait_for_compactions()

tchaikov:
before this change, wait_for_compactions() waits until all pending compactions are finished. but sometimes, we just want to

1. disable a certain $ks.$cf with column_family/autocompaction RESTful API, and
2. wait until all pending compaction tasks of the specified $ks.$cf is done, then
3. check the sstables of that table

column_family/autocompaction does not stop all compaction jobs a scylla instance, it just stops the specified table. so there are chances that we are still observing compaction tasks even if all the compactions of the specified table are stopped.

so, in this change, to cater the needs of the use case above, we add optional parameters to Node and ScyllaNode, so that we can specify the table we are interested in, instead of waiting for all tables.

also, instead of assigning `n` to `pending_task` before comparing them to detect the unexpected increasing number of pending tasks, we compare them before the assignment. otherwise, the branch is unreachable even if the `pending_tasks` is increasing.

Refs https://github.com/scylladb/scylladb/issues/16170

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -745,30 +745,66 @@ class Node(object):
         else:
             return False
 
-    def wait_for_compactions(self, idle_timeout=300):
+    def _parse_pending_tasks(self, output, keyspace, column_family):
+        # "nodetool compactionstats" prints the compaction stats like:
+        # pending tasks: 42
+        # - ks1.cf1: 13
+        # - ks1.cf2: 19
+        # - ks2.cf1: 10
+        head_pattern = r"pending tasks:\s*(?P<tasks>\d+)"
+        tail_pattern = re.compile(r'\-\s+(\w+)\.(\w+):\s*(\d+)')
+        head, *tail = output.strip().split('\n')
+
+        if keyspace is None and column_family is None:
+            matched = re.search(head_pattern, head)
+            if not matched:
+                raise RuntimeError(f"Cannot find 'pending tasks' in nodetool output.\nOutput: {output}")
+            return int(matched.group('tasks'))
+
+        # if keyspace or column_family is specified, check active tasks
+        # of the specified ks.cf instead
+        def matches(expected, actual):
+            if expected is None:
+                return True
+            return expected == actual
+
+        total = 0
+        for line in tail:
+            ks, cf, num = tail_pattern.search(line).groups()
+            if matches(keyspace, ks) and matches(column_family, cf):
+                total += int(num)
+        return total
+
+    def wait_for_compactions(self,
+                             keyspace=None,
+                             column_family=None,
+                             idle_timeout=300):
+        """Wait for all compactions to finish on this node.
+
+        :param keyspace: only wait for the compactions performed for specified
+            keyspace. if not specified, all keyspaces are waited
+        :param column_family: only wait for the compactions performed for
+            specified column_family. if not specified, all keyspaces are waited
+        :param idle_timeout: the time in seconds to wait for progress.
+            Total time to wait is undeteremined, as long as we observe forward
+            progress.
         """
-        Wait for all compactions to finish on this node.
-        idle_timeout is the time in seconds to wait for progress.
-        Total time to wait is undeteremined, as long as we observe forward progress.
-        """
-        pending_tasks = None
+        pending_tasks = -1
         last_change = None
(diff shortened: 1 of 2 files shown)


ccmlib/node.py
```diff
@@ -745,30 +745,59 @@ def stop(self, wait=True, wait_other_notice=False, other_nodes=None, gently=True
...
-    def wait_for_compactions(self, idle_timeout=300):
+    def _get_pending_tasks(self, stats, keyspace, column_family):
+        assert keyspace is not None, 'keyspace is not specified'
+        assert column_family is not None, 'column_family is not specified'
+        # "nodetool compactionstats" prints the compaction stats like:
+        # pending tasks: 42
+        # - ks1.cf1: 13
+        # - ks1.cf2: 19
+        # - ks2.cf1: 10
+        pattern = re.compile(r'\-\s+([^:]+):\s*(\d+)')
```
bhalevy, line 756: why not just use the regex to parse ks and cf for us to save the string format below?

ccmlib/node.py
```diff
@@ -745,30 +745,59 @@ def stop(self, wait=True, wait_other_notice=False, other_nodes=None, gently=True
...
+        for line in stats.strip().split('\n')[1:]:
+            ks_cf, num = pattern.search(line).groups()
+            if ks_cf == f'{keyspace}.{column_family}':
+                return int(num)
+        return 0
+
+    def wait_for_compactions(self,
+                             keyspace=None,
+                             column_family=None,
+                             idle_timeout=300):
```
bhalevy, line 766: new optional params should be added as last since existing call sites may not specify the idle_timeput by name and just invoke `node.wait_for_compactions(120)`

ccmlib/node.py
bhalevy: Also, specifying just the keyspace seems like a valid use case, so we can sum all the compaction jobs from all lines matching the given keyspace.

ccmlib/node.py
tchaikov: i agree you point in general. but because we always put keyspace and column_family as the first two argument, i took extra efforts to check all the callers in scylla-dtest and in scylla-ccm, none of the existing callers calls with positional argument.

ccmlib/node.py
tchaikov: > why not just use the regex to parse ks and cf for us to save the string format below?

this is the first version i had. but i changed to existing one. because the more regex the more convoluted. and why is it better than what we have? we are concatenating the ks and cf everywhere.

tchaikov: > Also, specifying just the keyspace seems like a valid use case, so we can sum all the compaction jobs from all lines matching the given keyspace.

can do. but i don't want to over-engineering. the existing implementation is able to address the need, and it is extensible.

bhalevy: It would be different if it was a utility function in dtest itself.  But since it is a library it should be extended in a more general way for future use cases, even if there's no immediate need.
I don't think the extension is that difficult and you're already touching the code it can be puggybacked on that relatively easily.

tchaikov: i want to invest my time in something we really need. if we have a real world use case, i would be happy to implement this. can we implement this feature when we need it?

tchaikov: done.

tchaikov:
v2:

- tally the total number of pending compactions with specified ks and/or cf

bhalevy (approved):

tchaikov:
@scylladb/qa-maintainers hello maintainers, could you help queue this change?

fruch (approved):
LGTM

For such a change I would rather have one run of the gating to make sure we are not breaking any gating tests

fruch:
@tchaikov, here this is a run for you:
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/180/ (OSS)
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/181/ (OSS + raft)
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/182/ (Enterprise)
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/183/ (Enterprise + raft)

if all of them are passing with this change, I'll merge.

PR #533, opened 2023-11-27
scylla_cluster: mark only first node as seed

bhalevy:
Starting from version 4.3 / 2021.1,
only the first node should be a seed in Scylla

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -222,9 +222,15 @@ class Cluster(object):
     def cassandra_version(self):
         return self.version()
 
+    def auto_seed(self, node):
+        # For legacy purposes, every new node is configured as seed by default.
+        return True
+
     def add(self, node, is_seed, data_center=None, rack=None):
         if node.name in self.nodes:
             raise common.ArgumentError(f'Cannot create existing node {node.name}')
+        if is_seed is None:
+            is_seed = self.auto_seed(node)
         self.nodes[node.name] = node
         if is_seed:
             self.seeds.append(node)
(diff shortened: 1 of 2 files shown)


fruch:
we should test is with all of gating

bhalevy:
@kbr-scylla there are some interesting failures with this change.
https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/52/testReport/

This is interesting: https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/52/testReport/update_cluster_layout_tests/TestUpdateClusterLayout/FullDtest___full_split001___test_add_new_node_while_adding_data_1_streaming_/
I'm not entirely convicened it didn't hit a pre-existing topology bug with RF=1.
I'm able to reprodice locally, even with consistent-topology-changes.

I can't explain:
https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/52/testReport/internode_ssl_test/TestInternodeSSL/FullDtest___full_split004___test_putget_with_internode_rack_ssl/
https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/52/testReport/internode_ssl_test/TestInternodeSSL/FullDtest___full_split006___test_putget_with_internode_ssl_with_dc_compression/
https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/52/testReport/internode_ssl_test/TestInternodeSSL/FullDtest___full_split006___test_putget_with_internode_rack_ssl_with_dc_compression/

fruch:
@bhalevy you think this is needed ?

bhalevy:
> @bhalevy you think this is needed ?

It would reflect reality better.

@kostja WDYT?

kostja:
I thikn it's fine either way and normally shouldn't make any difference in tests. 

What are the failures? the logs are no longer there.

bhalevy:
> I thikn it's fine either way and normally shouldn't make any difference in tests.
> 
> What are the failures? the logs are no longer there.

I don't remember.

The nain problem this change addresses in conforming to our own documentation, so we test what we tell customers to do.

I can rebare and rerun the tests to see what's the current state of the union.

PR #534, opened 2023-11-28
Remove --collectd-hostname from node command line invocation

mykaul:
We'd like to deprecate collectd, we use Prometheus these days. 

Refs: https://github.com/scylladb/scylla-enterprise/issues/3628


fruch (approved):
LGTM

PR #536, opened 2023-12-07
ccmlib: stop passing auto_bootstrap

tchaikov:
we stopped handling auto_bootstrap since
scylladb/scylladb@3b1ff90 and the behavior is like it's true, so stop passing this setting down.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -104,7 +104,8 @@ class Node(object):
         self.name = name
         self.cluster = cluster
         self.status = Status.UNINITIALIZED
-        self.auto_bootstrap = auto_bootstrap
+        # auto_bootstrap is deprecated
+        _ = auto_bootstrap
         self.network_interfaces = {'thrift': common.normalize_interface(thrift_interface),
                                    'storage': common.normalize_interface(storage_interface),
                                    'binary': common.normalize_interface(binary_interface)}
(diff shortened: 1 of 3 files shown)


tchaikov:
see also https://github.com/scylladb/scylla-dtest/pull/3769

tchaikov:
being tested at https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/55

ccmlib/node.py
```diff
@@ -150,7 +151,9 @@ def load(path, name, cluster):
             binary_interface = None
             if 'binary' in itf and itf['binary'] is not None:
                 binary_interface = tuple(itf['binary'])
-            node = cluster.create_node(data['name'], data['auto_bootstrap'], tuple(itf['thrift']), tuple(itf['storage']), data[
+            # auto_bootstrap is deprecated
+            auto_bootstrap = True
+            node = cluster.create_node(data['name'], auto_bootstrap, tuple(itf['thrift']), tuple(itf['storage']), data[
```
kbr-scylla, line 156: I guess we cannot remove the param from `create_node` because it might break users? (public api)

kbr-scylla (approved):
LGTM but ccm code is unfamiliar to me, you should ask someone who works with ccm regularly / maintains it to review.

ccmlib/node.py
```diff
@@ -104,7 +104,8 @@ def __init__(self, name, cluster, auto_bootstrap, thrift_interface, storage_inte
         self.name = name
         self.cluster = cluster
         self.status = Status.UNINITIALIZED
-        self.auto_bootstrap = auto_bootstrap
+        # auto_bootstrap is deprecated
+        _ = auto_bootstrap
         self.network_interfaces = {'thrift': common.normalize_interface(thrift_interface),
```
fruch, line 109: Can we issue a warning if someone is using it ?

ccmlib/node.py
fruch: Yeah even the main user is dtest

There are usage across our driver forks as well

fruch (approved):
Lgtm

Not sure why this was part of the API in the first place, and not a 2nd stage to configure a specific node before starting it

avelanarius:
This change breaks Java Driver 4.x CI with Cassandra 4.x - in one test node bootstrap starts timing out and now requires a very large timeout (120 seconds)

Issue #540, opened 2023-12-10
Only the first occurrence  `--experimental-features` in `SCYLLA_EXT_OPTS` is used


eliransin:
ccm don't seem to support Scylla command line options that can be repeated with different values,
when trying to use it with:
`export SCYLLA_EXT_OPTS="--experimental-features consistent-topology-changes --experimental-features tablets"`

Only the first option is given to Scylla (evident in node logs and actual enabled functionality).

Bouncheck:
I've encountered the same problem when writing tablets tests for java-driver.  It would be great if this got resolved.

Issue #539, opened 2023-12-10
Scylla startup fails with an error if "-c" or "-m" is used in `SCYLLA_EXT_OPTS`


eliransin:
It looks like ccm is unaware of the abbreviations for `--smp` and `--memory` (`-c` and `-m`) , this causes scylla startup to
fail with an error about same command line option for smp or/and memory being given twice.

Way to reproduce:
try to create and start a scylla cluster using `export SCYLLA_EXT_OPTS="-c 2"`

Scylla will fail with an error about duplicate command line options.

cezarmoise:
@fruch should this be closed due to https://github.com/scylladb/scylla-ccm/pull/558?

fruch:
> @fruch should this be closed due to #558?

Maybe, I have test this specific case.

PR #541, opened 2023-12-10
Fixes for handling `SCYLLA_EXT_OPTS` correctly

eliransin:
This mini-series implements two fixes for handling `SCYLLA_EXT_OPTS`:
1. It adds support for `-c` and `-m` options which are abbreviations for the `--smp` and `--memory` options and handles them correctly.
2. It adds support for multiple occurrences of  `--experimental-features` with correct handling.

The individual issues:
#539 and #540 describes the specific problems.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -549,7 +549,7 @@ class ScyllaNode(Node):
 
         MB = 1024 * 1024
 
-        def process_opts(opts):
+        def process_opts(opts, args_translation_map = {}):
             ext_args = OrderedDict()
             opts_i = 0
             while opts_i < len(opts):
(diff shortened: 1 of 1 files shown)


tchaikov:
I will review this change once I am back from my PTO.

eliransin:
@tchaikov ping

ccmlib/scylla_node.py
```diff
@@ -570,11 +570,12 @@ def process_opts(opts):
                         opts_i += 1
                     val = ' '.join(vals)
                 if key not in ext_args and not key.startswith("--scylla-manager"):
-                    ext_args[key] = val
+                    ext_args[args_translation_map.get(key, key)] = val
```
tchaikov: i'd suggest normalize `key` even before the `if key not in ext_args and not key.startswith("--scylla-manager")` check, and do this unconditially. like:
```py
key = args_translation_map.get(key, key)
if key not in ext_args and not key.startswith("--scylla-manager"):
    ext_args[key] = val
```

for better readaiblity.

ccmlib/scylla_node.py
```diff
@@ -549,8 +549,15 @@ def start(self, join_ring=True, no_wait=False, verbose=False,
 
         MB = 1024 * 1024
 
-        def process_opts(opts, args_translation_map = {}):
+        def process_opts(opts, args_translation_map = {}, allowed_duplicates_set = set()):
             ext_args = OrderedDict()
+            def should_add_option(key):
```
tchaikov, line 554: the nested function is more readable, but it also renders this method more complicated, how about this?

```py
if key.startswith("--scylla-manager"):
    continue

if key in allowed_duplicates_set:
    ext_args.get(key, []).append(val)
elif key not in ext_args:
    ext_args[key] = val
```

for following reasons:

1. the combinations of conditions is getting convoluted in this function, better off simplifying them
2. `dict.set_default()` helps to consolidate the logic, but it hurts the readability.
3. we should not use list for all values of options. the boxing-unboxing hurts the readability.

tchaikov:
@eliransin hi Eliran, sorry for the latency. left a couple comments.

fruch:
closing in favor of #558

PR #542, opened 2023-12-14
Remove --collectd option

mykaul:
We'd like to deprecate collectd, we use Prometheus these days. Refs: scylladb/scylla-enterprise#3628


fruch (approved):
Lgtm

PR #543, opened 2023-12-17
node: stress_object: strip off commas from value strings

bhalevy:
Some resharding_test dtest, e.g. `test_resharding_counter` fails locally for me:
```
    def _run_stress(self, op_cnt, stress_cmd):
        res = self.node.stress_object(stress_cmd)
        if not isinstance(res, dict):
            raise Exception('Error running cassandra-stress: {}'.format(res))
        assert res['total errors'] == 0
>       assert res['total partitions'] >= op_cnt
E       assert 10.0 >= 10000
```

When printing the cassandra-stress stdout I saw that it uses a comma as a 1000s separator:
```
Total partitions          :     10,000 [COUNTER_READ: 10,000]
```

This is probably related to the LOCALE on my machine.

This change explicitly sets the locale to en_US
tha defines comma as the thousands separator
and dot as the decimal point.

The origin local is reset when the function is done.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1381,24 +1381,25 @@ class Node(object):
 
     @staticmethod
     def _set_stress_val(key, val, res):
-        # Set locale to the user's default value (usually specified in the LANG)
-        locale.setlocale(locale.LC_NUMERIC, '')
+        def parse_num(s):
+            return locale.atof(s.replace(',', ''))
+
         if "[" in val:
             p = re.compile(r'^\s*([\d\.\,]+\d?)\s*\[.*')
             m = p.match(val)
             if m:
-                res[key] = locale.atof(m.group(1))
+                res[key] = parse_num(m.group(1))
             p = re.compile(r'^.*READ:\s*([\d\.\,]+\d?)[^\d].*')
             m = p.match(val)
             if m:
-                res[key + ":read"] = locale.atof(m.group(1))
+                res[key + ":read"] = parse_num(m.group(1))
             p = re.compile(r'.*WRITE:\s*([\d\.\,]+\d?)[^\d].*')
             m = p.match(val)
             if m:
-                res[key + ":write"] = locale.atof(m.group(1))
+                res[key + ":write"] = parse_num(m.group(1))
         else:
             try:
-                res[key] = locale.atof(val)
+                res[key] = parse_num(val)
             except ValueError:
                 res[key] = val
 
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1382,7 +1382,7 @@ class Node(object):
     @staticmethod
     def _set_stress_val(key, val, res):
         def parse_num(s):
-            return locale.atof(s.replace(',', ''))
+            return float(s.replace(',', ''))
 
         if "[" in val:
             p = re.compile(r'^\s*([\d\.\,]+\d?)\s*\[.*')


fruch (commented):
It's not clear why the locale is even involved in this code

I would remove it completely

bhalevy:
> It's not clear why the locale is even involved in this code
> 
> I would remove it completely

I think that the intention was to try to match the locale settings cassandara-stress uses to print its output numbers.
Some locales, like `de_DE` use comma as a decimal point and dot as the thousands separator.

bhalevy:
Yup, setting the locale explicitly to en_US works well as it defines
the thousands separator to comma.

ccmlib/node.py
```diff
@@ -1405,18 +1403,32 @@ def _set_stress_val(key, val, res):
...
-            if start:
-                m = p.match(line)
-                if m:
-                    Node._set_stress_val(m.group(1).strip().lower(), m.group(2).strip(), res)
-            else:
-                if line == 'Results:':
-                    start = True
+        prev_locale, encoding = locale.getlocale(locale.LC_NUMERIC)
+        # Set locale to en_US to simplify printing and parsing of numbers
+        locale.setlocale(locale.LC_NUMERIC, 'en_US')
```
fruch, line 1408: if we hard code to a specific one, 

we can just replace(',',''), and remove all locale related code

setting the locale like that, might break other code we aren't even aware of, like third party code.

fruch:
@bhalevy this change also introduce 3 more items into the object, which the unittests doesn't know about (and fails)

bhalevy:
> @bhalevy this change also introduce 3 more items into the object, which the unittests doesn't know about (and fails)

I dropped this part in the latest version ([265d8fe](https://github.com/scylladb/scylla-ccm/commit/265d8fe9fb24bf0d4f15f307ff0e0fcf1a654600)).
I thought this could help debug future issues but its not crucial right now.

bhalevy:
Hmm
```
=================================== FAILURES ===================================
________________ TestScyllaRelocatableCluster.test_node_stress _________________

self = <tests.test_scylla_relocatable_cluster.TestScyllaRelocatableCluster object at 0x7fc0b4a16820>
relocatable_cluster = <ccmlib.scylla_cluster.ScyllaCluster object at 0x7fc0b3d2ecd0>

    def test_node_stress(self, relocatable_cluster):
        node1, *_ = relocatable_cluster.nodelist()
        node1: Node
        ret = node1.stress(['write', 'n=10'])
        assert '10 [WRITE: 10]' in ret.stdout
        assert 'END' in ret.stdout
    
>       ret = node1.stress_object(['write', 'n=10'])

tests/test_scylla_relocatable_cluster.py:36: 
ccmlib/node.py:1408: in stress_object
    locale.setlocale(locale.LC_NUMERIC, 'en_US')

category = 1, locale = 'en_US'

    def setlocale(category, locale=None):
    
        """ Set the locale for the given category.  The locale can be
            a string, an iterable of two strings (language code and encoding),
            or None.
    
            Iterables are converted to strings using the locale aliasing
            engine.  Locale strings are passed directly to the C lib.
    
            category may be given as one of the LC_* values.
    
        """
        if locale and not isinstance(locale, _builtin_str):
            # convert to string
            locale = normalize(_build_localename(locale))
>       return _setlocale(category, locale)
E       locale.Error: unsupported locale setting

/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/locale.py:608: Error
```

fruch:
> Hmm
> ```
> =================================== FAILURES ===================================
> ________________ TestScyllaRelocatableCluster.test_node_stress _________________
> 
> self = <tests.test_scylla_relocatable_cluster.TestScyllaRelocatableCluster object at 0x7fc0b4a16820>
> relocatable_cluster = <ccmlib.scylla_cluster.ScyllaCluster object at 0x7fc0b3d2ecd0>
> 
>     def test_node_stress(self, relocatable_cluster):
>         node1, *_ = relocatable_cluster.nodelist()
>         node1: Node
>         ret = node1.stress(['write', 'n=10'])
>         assert '10 [WRITE: 10]' in ret.stdout
>         assert 'END' in ret.stdout
>     
> >       ret = node1.stress_object(['write', 'n=10'])
> 
> tests/test_scylla_relocatable_cluster.py:36: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> ccmlib/node.py:1408: in stress_object
>     locale.setlocale(locale.LC_NUMERIC, 'en_US')
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> 
> category = 1, locale = 'en_US'
> 
>     def setlocale(category, locale=None):
>     
>         """ Set the locale for the given category.  The locale can be
>             a string, an iterable of two strings (language code and encoding),
>             or None.
>     
>             Iterables are converted to strings using the locale aliasing
>             engine.  Locale strings are passed directly to the C lib.
>     
>             category may be given as one of the LC_* values.
>     
>         """
>         if locale and not isinstance(locale, _builtin_str):
>             # convert to string
>             locale = normalize(_build_localename(locale))
> >       return _setlocale(category, locale)
> E       locale.Error: unsupported locale setting
> 
> /opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/locale.py:608: Error
> ```
> 

As I said, we don't know which locale are available where we're gonna run, we shouldn't be setting it inside a library code.

bhalevy:
Ok, I'll go back to square one

bhalevy:
@fruch please merge

ccmlib/node.py
```diff
@@ -1381,24 +1381,25 @@ def _set_keyspace_initial_tablets(stress_options: List[str], initial_tablets:int
 
     @staticmethod
     def _set_stress_val(key, val, res):
-        # Set locale to the user's default value (usually specified in the LANG)
-        locale.setlocale(locale.LC_NUMERIC, '')
+        def parse_num(s):
+            return locale.atof(s.replace(',', ''))
```
fruch, line 1385:
```suggestion
            return float(s.replace(',', ''))
```

fruch: we don't need to use locale reading a float.

fruch (approved):
LGTM

PR #545, opened 2023-12-24
node: _parse_pending_tasks: handle tail_pattern mismatch

bhalevy:
`nodetool compactionstats` may contain more lines after the pending tasks section.

For exameple:
```
pending tasks: 2
- system_schema.tables: 1
- system_schema.columns: 1

id                                   compaction type keyspace      table   completed total unit progress
8e1f2d90-a252-11ee-a7f4-1bf9ae4e6ffd COMPACTION      system_schema columns 1         640   keys 0.16%
Active compaction remaining time :        n/a
```

Break from the `_parse_pending_tasks` loop once the tail_pattern regex mismatches the line to prevent the following error, as seen for example in https://jenkins.scylladb.com/job/scylla-master/job/dtest-debug/270/testReport/resharding_test/TestReshardingTombstonesSingleNode/Run_Dtest_Parallel_Cloud_Machines___FullDtest___full_split004___test_disable_tombstone_removal_during_reshard_1_SizeTieredCompactionStrategy_15_/
```
        for line in tail:
>           ks, cf, num = tail_pattern.search(line).groups()
E           AttributeError: 'NoneType' object has no attribute 'groups'
```

This fix is on top of ae645ad1

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -747,7 +747,8 @@ class Node(object):
         else:
             return False
 
-    def _parse_pending_tasks(self, output, keyspace, column_family):
+    @staticmethod
+    def _parse_pending_tasks(output, keyspace, column_family):
         # "nodetool compactionstats" prints the compaction stats like:
         # pending tasks: 42
         # - ks1.cf1: 13
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -773,7 +773,10 @@ class Node(object):
 
         total = 0
         for line in tail:
-            ks, cf, num = tail_pattern.search(line).groups()
+            m = tail_pattern.search(line)
+            if not m:
+                break
+            ks, cf, num = m.groups()
             if matches(keyspace, ks) and matches(column_family, cf):
                 total += int(num)
         return total
--- a/tests/test_internal_functions.py
+++ b/tests/test_internal_functions.py
@@ -1,0 +1,40 @@
+from ccmlib.node import Node
+
+def test_parse_pending_tasks():
+    def verify_result(output, keyspace, column_family, expected):
+        n = Node._parse_pending_tasks(output, keyspace, column_family)
+        assert n == expected
+
+    def verify_cases(output, cases):
+        for ks, cf, expected in cases:
+            verify_result(output, ks, cf, expected)
+
+    for output in [
+        '''
+        pending tasks: 6
+        - system_schema.tables: 1
+        - system_schema.columns: 2
+        - keyspace1.standard1: 3
+        ''',
+        '''
+        pending tasks: 6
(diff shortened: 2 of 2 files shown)


fruch:
@bhalevy since when it this change that causes this ?

ccmlib/node.py
```diff
@@ -772,7 +772,11 @@ def matches(expected, actual):
 
         total = 0
         for line in tail:
-            ks, cf, num = tail_pattern.search(line).groups()
+            print(f"'{line}'")
```
fruch, line 775: using logging...

fruch: also maybe we want a unit test for this parse function

bhalevy:
> @bhalevy since when it this change that causes this ?

As I wrote in the description, it's from https://github.com/scylladb/scylla-ccm/commit/ae645ad1f2202870394ae59065a0b2f48686907a which is currently only in master AFAICT

ccmlib/node.py
bhalevy: > using logging...

sorry, left over from debugging, already removed

bhalevy:
[84f473d](https://github.com/scylladb/scylla-ccm/commit/84f473d99304a5c7f3dba676bd7b72300ae036d2):
- make `_parse_pending_tasks` a static method
- added unit test for it

ccmlib/node.py
```diff
@@ -772,7 +773,10 @@ def matches(expected, actual):
 
         total = 0
         for line in tail:
-            ks, cf, num = tail_pattern.search(line).groups()
+            m = tail_pattern.search(line)
+            if not m:
```
tchaikov, line 777: if m is None:


More specific this way.

tchaikov (approved):
Thank you for fixing the issue!

ccmlib/node.py
bhalevy: Why do we care?
`not m` is simpler and sensible

tchaikov: i don't feel strong either way. It's like `m == nullptr` versus `!m` in c/c++ world. The latter works as well, but one cannnot deduce that `m` is a pointer by reading the code.

bhalevy: Right, and I prefer the boolean operator rather than comparing to the null value where it's not needed explicitly, like when you to distinguish between an arg default value of None vs. a valid value of False or [].
In the latter case, None would mean unset, not just False.

tchaikov: yeah, I don't insist in this case. But, personally , I prefer explicit even at the expense of more verbosity.

bhalevy:
@fruch please re-review / merge

fruch (approved):
Lgtm

PR #546, opened 2023-12-24
scylla_node: run_scylla_sstable: print stderr on failure

bhalevy:
If `scylla sstable` fails, the standard printouts for `subprocess.CalledProcessError` are insufficient to determine what went wrong.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -29,7 +29,7 @@ from ccmlib.common import CASSANDRA_SH, BIN_DIR, wait_for, copy_directory
 from ccmlib import common
 from ccmlib.node import Node, NodeUpgradeError
 from ccmlib.node import Status
-from ccmlib.node import NodeError
+from ccmlib.node import NodeError, ToolError
 from ccmlib.node import TimeoutError
 from ccmlib.scylla_repository import setup, get_scylla_version
 from ccmlib.utils.version import parse_version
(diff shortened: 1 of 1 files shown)


ccmlib/scylla_node.py
```diff
@@ -1469,7 +1469,13 @@ def do_invoke(sstables):
...
             common_args = [scylla_path, "sstable", command] + additional_args
             env = self._get_environ()
-            res = subprocess.run(common_args + sstables, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=text, check=True, env=env)
+            res = subprocess.run(common_args + sstables, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=text, check=False, env=env)
+            if res.returncode:
+                msg = f"{common_args + sstables} failed:\n"
+                for l in res.stderr.split('\n'):
+                    msg += f"> {l}\n"
+                self.debug(msg)
+                res.check_returncode()
```
fruch, line 1478: Mayne use `ToolError`, like in nodetool and c-s calls ? it raise with the stdout and stderr

you can even use `handle_external_tool_process` directly as in `stress()`, just need to switch to `subprocess.Popen` instead of `subprocess.run`

bhalevy: Thanks.  Done in next version.

bhalevy:
@fruch please re-review / merge

fruch (approved):
Lgtm

PR #549, opened 2024-01-10
cluster: add add_seed method

bhalevy:
To be used after replacing nodes.
We cannot add the replacing nodes to seeds right away since it's considered invalid to add a replacing node as a seed (probably for historical reasons).
So we need to add it only after it's finished
replace bootstrapping.


bhalevy:
ping @fruch please review/merge

fruch (approved):
LGTM

PR #550, opened 2024-01-12
Revert "ccmlib: stop passing auto_bootstrap"

tchaikov:
Reverts scylladb/scylla-ccm#536

according to Piotr Grabowski

> This change breaks Java Driver 4.x CI with Cassandra 4.x - in one test node bootstrap starts timing out and now requires a very large timeout (120 seconds)

so revert the offending change.


avelanarius (approved):

fruch (approved):
Lgtm

PR #551, opened 2024-01-15
scylla_node: use str.split('=') when appropriate

tchaikov:
instead of chaining replace() and split(), let's just use split(). simpler and more readable this way


fruch (approved):
LGTM

PR #552, opened 2024-01-16
ccmlib/common: check for scylla built with CMake as well 

tchaikov:
cmake generates scylla at paths like `build/Debug/scylla` instead of
`build/debug/scylla`, so check for them as well.

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -591,11 +591,15 @@ def isScylla(install_dir):
 
         raise ArgumentError('Undefined installation directory')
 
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM

I'm guessing in some part in dtest itself would need adapting

tchaikov:
> LGTM
> 
> I'm guessing in some part in dtest itself would need adapting

yes, i am working on that end also.

tchaikov:
@scylladb/qa-maintainers  hello maintainers, could you help merge this change?

soyacz (approved):
LGTM

PR #556, opened 2024-01-25
scylla_node: print `check_socket_available` failure to log

fruch:
in cases there are failure in dtest runs, those print was going to stdout, which is hard to track, and might not be save at all in when using `pytest-xdist`

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -24,7 +24,7 @@ import glob
 import re
 import requests
 
-from ccmlib.common import CASSANDRA_SH, BIN_DIR, wait_for, copy_directory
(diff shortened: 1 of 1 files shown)


juliayakovlev (approved):

ccmlib/scylla_node.py
```diff
@@ -533,10 +533,10 @@ def start(self, join_ring=True, no_wait=False, verbose=False,
                     try:
                         common.check_socket_available(itf)
                     except Exception as msg:
-                        print(f"{msg}. Looking for offending processes...")
+                        print_if_standalone(f"{msg}. Looking for offending processes...", debug_callback=logging.error)
```
tchaikov, line 536: not in the scope of this change, but we might need to only catch `OSError` and check its `errno` before looking for the offending process.

tchaikov (approved):
lgtm

PR #557, opened 2024-01-28
improvement(setup): improve caching of local tarballs

juliayakovlev:
If the test version is unstable (not release, maybe private branch) and installation folder exists, we want to check if this version was downloaded nd installed already in the past and was changed.
In this case the version should be downloaded and installed again. Compare hash of saved version (it is saved in the 'scylla-core-package/source.txt' file) and hash of a new package.
If it is same - skip the download. If not - remove existing folder and download again.

Tests:
 - [x] https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/133/ (branch-5.4)
 - [x] https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/134/ (master)
 - [x] https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/135/ (using ARTIFACT_WEB_URL)
 - [x] local run


ccmlib/scylla_repository.py
```diff
@@ -244,7 +245,15 @@ def setup(version, verbose=True, skip_downloads=False):
...
     version_dir = version_directory(version) if not skip_downloads else None
 
-    if len(type_n_version) == 2 and version_dir is None:
+    # If the test version is unstable (not release, maybe private branch) and installation folder exists,
+    # we want to check if this version was downloaded nd installed already in the past and was changed.
+    # In this case the version should be downloaded and installed again.
+    # Compare hash of saved version (it is saved in the 'scylla-core-package/source.txt' file) and hash of a new package.
+    # If it is same - skip the download. If not - remove existing folder and download again.
+    # TODO: is it good condition?
+    validate_by_hash = version_dir is not None and type_n_version[0] != "release"
```
juliayakovlev, line 254: @fruch  is ti good enough condition?

ccmlib/scylla_repository.py
```diff
@@ -303,6 +312,26 @@ def setup(version, verbose=True, skip_downloads=False):
     if skip_downloads:
         return directory_name(version), packages
 
+    if validate_by_hash:
+        if packages.scylla_unified_package:
+            new_hash = get_url_hash(packages.scylla_unified_package)
+        else:
+            # Old releases, there are no Scylla unified_package
+            # TODO: what's in this case?
+            new_hash = ""
```
juliayakovlev, line 321: @fruch is it real case?

ccmlib/scylla_repository.py
fruch: I would save one call for the release case

ccmlib/scylla_repository.py
fruch: add a check to packages.scylla_unified_package, and skip the hash checking if it doesn't exist

ccmlib/utils/download.py
```diff
@@ -172,7 +172,8 @@ def get_url_hash(url: str) -> str:
     """
 
     if os.path.exists(url):  # if file/dir is local, hash based on the path
-        return hashlib.md5(url).hexdigest()
+        # Strings must be encoded before hashing
+        return hashlib.md5(url.encode('utf-8')).hexdigest()
```
fruch, line 176: it might worth adding a unit test for this case (we have test for downloading, but not for using local file)

juliayakovlev: added

ccmlib/scylla_repository.py
juliayakovlev: added support of non-unified packages

ccmlib/common.py
```diff
@@ -973,3 +974,17 @@ def print_if_standalone(*args, debug_callback=None, end='\n', **kwargs):
...
+    current_hash = ""
+
+    # If source file does not exists - we can not to check the hash of the existing package
+    if source_file.exists():
+        with open(source_file, 'r') as f:
+            lines = f.readlines()
+        # get hash from file
+        for line in lines:
+            if line.startswith("hash="):
+                current_hash = line.replace("hash=", "").strip()
```
tchaikov, line 989: is it possible that the `source_file` does not contain a line of "hash=.."? is it possible that it contains multiple lines of "hash=.."? if both answers are "no", i'd suggest put something like:
```py
if not source_file.exists():
    return ""
with open(source_file, 'r') as f:
    for line in f:
        if line.startswith("hash="):
            return line.replace("hash=", "").strip()
throw ArgumentError(f"{source_file} does not contains the hash")
```

less indent this way. and less tolerant for sure.

ccmlib/scylla_repository.py
```diff
@@ -303,6 +311,32 @@ def setup(version, verbose=True, skip_downloads=False):
     if skip_downloads:
         return directory_name(version), packages
 
+    if validate_by_hash and packages:
+        """
+        Validate if packages hash was changed and the new package(s) have to be downloaded 
+        """
```
tchaikov, line 317: nit, i'd suggest use a comment instead of a multi-line string for comment. as it cannot be used as doc-string.

ccmlib/common.py
juliayakovlev: Few lines with "hash" is not possible because we replace whole file and not add the lines.

No "hash" it is possible. Due to this I added the comment:

https://github.com/scylladb/scylla-ccm/pull/557/files#:~:text=%23%20Current%20hash%20may,start%20downloading%20again

```
# Current hash may be None. It may be due to uncompleted downloading and installation.
            # Or because of it is old installation that was not saved the hash yet.
            # In any case the new download and installation should be performed.
            # For this goal we need to remove existing folder and start downloading again
```

ccmlib/scylla_repository.py
```diff
@@ -303,6 +311,32 @@ def setup(version, verbose=True, skip_downloads=False):
...
+
+            new_hash = get_url_hash(package[1])
+            package_dir = map_field_to_dir_name[package[0]]
+            current_hash = get_installed_scylla_package_hash(source_file=Path(version_dir) / package_dir / SOURCE_FILE_NAME)
+            # Current hash may be None. It may be due to uncompleted downloading and installation.
+            # Or because of it is old installation that was not saved the hash yet.
+            # In any case the new download and installation should be performed.
+            # For this goal we need to remove existing folder and start downloading again
+            if not (new_hash and current_hash) or new_hash != current_hash:
+                # remove version_dir
```
tchaikov, line 335: IIUC, `new_hash` should always be a non-empty string, right? so, probably we can simplify this a little bit. how about:

```py
for package in zip(packages._fields, packages):
    # ...
    if current_hash is None or new_hash != current_hash:
        rmdirs(version_dir)
        version_dir = None
        break
```

tchaikov (commented):
left a couple comments.

ccmlib/common.py
juliayakovlev: the source.txt file - version will be download

ccmlib/scylla_repository.py
fruch: yep, if we had a linter of any kind on this project, it would have complain about this

ccmlib/scylla_repository.py
fruch: so yes let's keep it

tests/test_scylla_repository.py
```diff
@@ -115,6 +118,62 @@ def test_setup_unstable_master_new_url(self):
...
+        start_time = time.time()
+        cdir, version = scylla_setup(version="unstable/master:2023-04-03T22:38:18Z", verbose=True, skip_downloads=False)
+        end_time = time.time()
+        assert (end_time - start_time) > 5
+        assert '2023-04-03T22_38_18Z' in cdir
+        assert version == '5.3.0-dev'
+
+    def test_get_local_tarball_hash(self):
+        this_path = Path().resolve()
+        url_hash = get_url_hash(url=str(this_path / "tests" / "test_data" / "scylla_unified_master_2023_04_03.tar.gz"))
```
fruch, line 174: where is that file ? in the the source code ? 

let make sure it's not a full sized release, the a minimal empty `tar.gz`

fruch: and yeah, CI show there no such file:
```
FAILED tests/test_scylla_repository.py::TestScyllaRepositoryRelease::test_get_local_tarball_hash - requests.exceptions.MissingSchema: Invalid URL '/home/runner/work/scylla-ccm/scylla-ccm/tests/test_data/scylla_unified_master_2023_04_03.tar.gz': No scheme supplied. Perhaps you meant https:///home/runner/work/scylla-ccm/scylla-ccm/tests/test_data/scylla_unified_master_2023_04_03.tar.gz?
```

ccmlib/scylla_repository.py
juliayakovlev: done

ccmlib/common.py
tchaikov: i see. then still good to return early in that case.

tests/test_scylla_repository.py
juliayakovlev: I took is from unstable. It's size 166M

ccmlib/scylla_repository.py
tchaikov: i see. the `!` and `not` in the condition expression makes my eyes hurt. hence this suggestion. if we can truest the 'ETag' returned by S3, then it should not be empty.

tests/test_scylla_repository.py
juliayakovlev: > and yeah, CI show there no such file:
> 
> ```
> FAILED tests/test_scylla_repository.py::TestScyllaRepositoryRelease::test_get_local_tarball_hash - requests.exceptions.MissingSchema: Invalid URL '/home/runner/work/scylla-ccm/scylla-ccm/tests/test_data/scylla_unified_master_2023_04_03.tar.gz': No scheme supplied. Perhaps you meant https:///home/runner/work/scylla-ccm/scylla-ccm/tests/test_data/scylla_unified_master_2023_04_03.tar.gz?
> ```

Yeah, I did not added it. fixing

juliayakovlev: @fruch 
Oh, it is too big. It's is allowed 100m. I'll take any tarball. It is not must to be scylla package, right?

juliayakovlev: fixed

tests/tests/test_data/scylla_unified_master_2023_04_03.tar.gz
fruch, line 1: @juliayakovlev as I've said, if we can make it even smaller, I would rather not had 24mb binary on a git repository

and if we need something at this size for some reason, that it would fetch or generated during the test

juliayakovlev: > @juliayakovlev as I've said, if we can make it even smaller, I would rather not had 24mb binary on a git repository
> 
> and if we need something at this size for some reason, that it would fetch or generated during the test

replaced it with small 131K. is it ok?

fruch: it's better than 24Mb

ccmlib/common.py
juliayakovlev: Why do you think we need to return and stop the test? Maybe there is the reason that I did not think?

fruch: I think @tchaikov is talking more on the style, i.e. readability of it, nothing logical you are missing here. 
i.e. no need for exception, but the style suggestion still can be relevert

ccmlib/scylla_repository.py
fruch: if `get_url_hash` always returns string, there no need to check for emptiness of it return value.

also `get_installed_scylla_package_hash` doesn't return None (at least not now), it always return a string, could be empty, so using `is None` wouldn't work as one might expect.

so `current_hash and new_hash != current_hash` should be suffice
if `get_url_hash` might return None
so `current_hash and new_hash and new_hash != current_hash` should do the trick.

juliayakovlev: I am not sure if we can trust. 'ETag' is not mandatory. https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html

@fruch please, let you opinion

ccmlib/common.py
juliayakovlev: @tchaikov I changed it, let me know if it is more clear

ccmlib/scylla_repository.py
juliayakovlev: @tchaikov I changed it, let me know if it is more clear

ccmlib/common.py
tchaikov: thanks. indeed. i was concerned about the readability. the code was and still correct.

juliayakovlev:
@fruch can you review, please?

fruch (approved):
LGTM

@juliayakovlev can you also update dtest readme, and remove the comment about this issue?

PR #558, opened 2024-02-05
support multiple identical args in SCYLLA_EXT_OPTS

fruch:
since we have cases multiple `--experimental-features` needs to be passed via `SCYLLA_EXT_OPTS`, adding support for key with multiple values in `process_opts()`

fixes: #540
closes: #541

## Testing - to be on the safe side:
- [x] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/340/ (OSS)
- [x] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/341/ (Enterprise)

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -579,8 +579,8 @@ class ScyllaNode(Node):
                         vals.append(opts[opts_i])
                         opts_i += 1
                     val = ' '.join(vals)
-                if key not in ext_args and not key.startswith("--scylla-manager"):
-                    ext_args[key] = val
+                if not key.startswith("--scylla-manager"):
+                    ext_args.setdefault(key, []).append(val)
             return ext_args
 
         # Lets search for default overrides in SCYLLA_EXT_OPTS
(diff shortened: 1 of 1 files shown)


nyh (approved):
Looks good to me. I assumed you tested it actually works? :-)

ccmlib/scylla_node.py
```diff
@@ -609,7 +609,10 @@ def process_opts(opts):
             elif k != '--memory':
                 args.append(k)
                 if v:
-                    args.append(v)
+                    args.append(v[0])
+                    for val in v[1:]:
+                        args.append(k)
+                        args.append(val)
```
tchaikov, line 615: if we really want to unify the argument to be multiple values, i'd suggest:
```py
elif k != '--memory':
    for val in v:
        args += [k, val]
```

tchaikov:
might want to reference #541.

copy @eliransin , hi Eliran, in case you are still interested in this issue.

ccmlib/scylla_node.py
fruch: yeah that a bit clearer

fruch:
> Looks good to me. I assumed you tested it actually works? :-)

it works for the cases reported, i.e. multiple `--experimental-features`

ccmlib/scylla_node.py
fruch: used that, and it's much better, thanks.

tchaikov (approved):
lgtm

Bouncheck (approved):
Looks to be working for me. My test had issues, but the cluster launches with system.tablets present and I see both experimental features in system.log

fruch:
tests showed, that a few places were needed to change and expect a list, not assume an int.

fixed those and re-running gating

fruch:
Code is failing in gating

Need to check why it's still breaking Scylla boot

bhalevy:
@nyh by the way, why don't we support comma-separated list arg in scylla --experimental-features?

nyh:
I don't know (I don't think I wrote that code?), we definitely could. But we need to also keep supporting the current approach (multiple instances of the option), because it's convenient in scripts.

fruch:
> Code is failing in gating
> 
> Need to check why it's still breaking Scylla boot

found the issue, the args without value, was passing empty space as value (it wasn't clear from logs)
now fixe, and giving one more go in gating

PR #559, opened 2024-02-11
scylla_node: pass -Dcom.scylladb.apiPort=10000 when running nodetool

tchaikov:
this pass the setting of "api_port" to nodetool-wrapper, so that it can override the "-p" option with the port number. otherwise "scylla nodetool" would not be able to access the API service by connecting to the JMX server.


tchaikov:
the change on the nodetool side can be found at https://github.com/scylladb/scylla-tools-java/pull/380

fruch (approved):
LGTM

denesb (approved):

PR #560, opened 2024-02-12
scylla_node: pass -Dcom.scylladb.apiPort=10000 when running nodetool

tchaikov:
this pass the setting of "api_port" to nodetool-wrapper, so that it can override the "-p" option with the port number. otherwise "scylla nodetool" would not be able to access the API service by connecting to the JMX server.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -757,12 +757,15 @@ class ScyllaNode(Node):
         else:
             self.jmx_pid = None
 
-    def nodetool(self, *args, **kwargs):
+    def nodetool(self, cmd, capture_output=True, wait=True, timeout=None, verbose=True):
         """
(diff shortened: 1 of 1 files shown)


tchaikov:
v2:

- replicate the function signature of the parent class: the previous version of #559 failed to handle this case.

tchaikov:
being tested at https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/138/

denesb (approved):

denesb:
> being tested at https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/140/

@fruch the BYO passed, can you please queue this?

fruch (approved):
LGTM

PR #562, opened 2024-02-23
ccmlib/scylla_node: use workdir option

margdoc:
Use the workdir option to set all subdirectories instead of setting them separately.
This change allows to set options dependent on the work directory, for example, maintenance-socket.

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1127,13 +1127,17 @@ class ScyllaNode(Node):
                 self.get_base_cassandra_version() >= 1.2):
             _, data['native_transport_port'] = self.network_interfaces['binary']
 
+        # Use "workdir,W" instead of "workdir", because scylla defines this option this way
+        # and dtests compares names of used options with the names defined in scylla.
+        data['workdir,W'] = self.get_path()
+        # This option is set separately from the workdir to keep backward compatibility with scylla java tools
+        # such as sstablelevelreset which needs this option to get the path to the data directory.
         data['data_file_directories'] = [os.path.join(self.get_path(), 'data')]
+        # The default path of commitlog subdirectory is workdir/commitlog.
+        # This option is set separately from the workdir to keep backward compatibility,
+        # because some dtests use `commitlogs` value to get the commitlog directory.
         data['commitlog_directory'] = os.path.join(self.get_path(),
                                                    'commitlogs')
-        data['hints_directory'] = os.path.join(self.get_path(), 'hints')
-        data['saved_caches_directory'] = os.path.join(self.get_path(),
-                                                      'saved_caches')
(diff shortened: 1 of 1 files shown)


kbr-scylla:
Please prove that this doesn't break dtests by providing a Jenkins run incorporating this change that runs at least the next-gating subset of dtests

ccmlib/scylla_node.py
```diff
@@ -1127,13 +1127,11 @@ def update_yaml(self):
                 self.get_base_cassandra_version() >= 1.2):
             _, data['native_transport_port'] = self.network_interfaces['binary']
 
-        data['data_file_directories'] = [os.path.join(self.get_path(), 'data')]
+        data['workdir,W'] = self.get_path()
+        # The default path of commitlog subdirectory is workdir/commitlog.
+        # This option is set seperately from the workdir to keep backward compatibility with `commitlogs` value.
```
kbr-scylla, line 1132: s/seperately/separately

I'd also add why we need this backwards compatibility --- because some dtests depend on it

margdoc:
v2:
* discard setting maintenance-socket option (it will be done in scylla)

margdoc:
v3:
* fixed comment and added comment about dtests.

margdoc:
> Please prove that this doesn't break dtests by providing a Jenkins run incorporating this change that runs at least the next-gating subset of dtests

Ok, this PR is a draft until Jenkins run is provided.

ccmlib/scylla_node.py
```diff
@@ -1127,13 +1127,12 @@ def update_yaml(self):
                 self.get_base_cassandra_version() >= 1.2):
             _, data['native_transport_port'] = self.network_interfaces['binary']
 
-        data['data_file_directories'] = [os.path.join(self.get_path(), 'data')]
+        data['workdir,W'] = self.get_path()
```
fruch, line 1130: `'workdir,W'` is this a typo ? what kind of a configuration key is that ?

margdoc: "workdir,W" is used instead of "workdir", because scylla defines this option this way and dtests compare the names of used options with the names defined in scylla.

margdoc:
v4:
* added comment about the use of "workdir,W"
* use `data_file_directories` option

kbr-scylla (approved):

margdoc:
Successful Jenkins run with these changes: https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/2103/

margdoc:
@fruch ping

fruch (approved):
LGTM,

but we are in a code freeze, so only CI stability can be merged, they would need to wait

kbr-scylla:
Merge freeze is over, @fruch can you merge?

Issue #563, opened 2024-03-08
`node.start(wait_other_notice=True, ...)` may return too early because implementation of `wait_other_notice` may observe past state


kbr-scylla:
`wait_other_notice` in ccmlib/scylla_node.py currently works by looking at the /gossiper/live endpoint API
https://github.com/scylladb/scylla-ccm/blob/bd519ed12cccd8186dedf8613656f21dd156354e/ccmlib/scylla_node.py#L334
https://github.com/scylladb/scylla-ccm/blob/bd519ed12cccd8186dedf8613656f21dd156354e/ccmlib/scylla_node.py#L1361

If other nodes see the node as UP, and the node sees other nodes as UP, the check will succeed.

But consider the following:
- 2 nodes A, B
- kill node A
- start node A with `wait_other_notice=True`

The `start` call will often return quickly, but not because node B noticed that A restarted
but because node B didn't even notice that A was killed yet!

So node B may still consider A as UP because it was UP before it was killed; B's failure detector just didn't react to that event yet.

That's why the old implementation in ccmlib/node.py used **log markers**, it watched other nodes' logs **from a point after the node was restarted**, effectively serializing the events.

This bug was introduced in https://github.com/scylladb/scylla-ccm/pull/462.

cc @temichus

kbr-scylla:
We could fix it by combining the old behavior (with log markers) with the new behavior (watching REST). For example -- we could wait for the "is now UP" message first, then use `watch_rest_for_alive`.

Although, looking at Scylla code, I don't really understand why "is now UP" is not sufficient; it happens strictly after modifying live_endpoints set in gossiper, which is used by the /gossiper/live REST API.
Perhaps it's necessary for some older versions of Scylla.
And I see that `watch_rest_for_alive` contains additional logic for watching for NORMAL too, so that is still useful.

kbr-scylla:
Example failure:
https://github.com/scylladb/scylla-dtest/pull/4042/files

the test sometimes fails with
```
        # starting node1 - it should reconnect and run as is
        node1.start(wait_other_notice=True, wait_for_binary_proto=True)
        #  due to CL=1 it does not make sense to read while not all nodes are UP
        # for node in cluster.nodelist():
        #     wait_for_nodes_status(node, ["UN", "UN", "UN"])
    
>       result = list(session.execute("SELECT * FROM cf"))

update_cluster_layout_tests.py:1136: 
[...]

PR #564, opened 2024-03-12
scylla_cluster, scylla_node: watch_log_for_alive before watch_rest_for_alive

bhalevy:
https://github.com/scylladb/scylla-ccm/pull/462 introduced watch_rest_for_alive that replaced the calls to watch_log_for_alive on the scylla node(s) start path.

But a node is killed and then restarted, and other nodes miss the kill event, `watch_rest_for_alive` will consider that node already as up as seen by the other nodes, while previously, watch_log_for_alive, waited until other nodes discovered this node as up again, based on markes taken right before (re)starting that node.

This change brings this call back.

Fixes scylladb/scylla-ccm#563

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -156,9 +156,10 @@ class ScyllaCluster(Cluster):
                                    verbose=verbose, from_mark=mark)
 
         if wait_other_notice:
-            for old_node, _ in marks:
+            for old_node, mark in marks:
                 for node, _, _ in started:
                     if old_node is not node:
+                        old_node.watch_log_for_alive(node, from_mark=mark, timeout=self.default_wait_other_notice_timeout)
                         old_node.watch_rest_for_alive(node, timeout=self.default_wait_other_notice_timeout,
                                                       wait_normal_token_owner=wait_normal_token_owner)
 
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -332,8 +332,9 @@ class ScyllaNode(Node):
                 self.wait_for_binary_interface(from_mark=from_mark, process=self._process_scylla, timeout=t)
 
         if wait_other_notice:
-            for node, _ in marks:
+            for node, mark in marks:
                 t = self.cluster.default_wait_other_notice_timeout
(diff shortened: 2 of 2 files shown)


bhalevy:
- [ ] BYO test for test_simple_kill_remained_node_while_decommissioning, 100 times: https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/188/

nyh:
Oops, @fruch @roydahan it appears I was given commit permissions in the dtest repository, but not in the ccm repository. So someone else will need to merge this PR. I approved it.

fruch:
I kind of agree with @nyh, that I don't really like it.
and we should try to shy away from relying on the logs for this, (we are keep being accused by @kosta and other, why dtest and ccm depends on the logs, while this kind of yet again proved we don't have better alternatives, and no one agree even what they are)

anyhow this change affects all of the test, and we should do a sweep with gating for it before merging

fruch:
we'll wait for the results:

- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/344/ (OSS)
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/345/ (Enterprise)

kosta:
> we are keep being accused by @kosta

Not by me, do mean another Kosta?

bhalevy:
> > we are keep being accused by @kosta
> 
> Not by me, do mean another Kosta?

@kostja :)

fruch:
> > we are keep being accused by @kosta
> 
> Not by me, do mean another Kosta?

Sorry, now you have something to accuse me for, too.

fruch:
> we'll wait for the results:
> 
> - [x] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/344/ (OSS)
> - [x] https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/345/ (Enterprise)

Runs were green, merging it in

kostja:
accuse or not, scylla-ccm should retire just like the java based nodetool which thanks to @denesb bit the dust. sorry.

PR #566, opened 2024-03-22
github/workflows/nix: extract RELOC_VERSION as an env variable

tchaikov:
encode RELOC_VERSION in the key of binary, so we can use different binary for testing. there is chance that newer ccm would like to use newer scylla executables for testing.

this series is more a refactory, and it does not change the version of scylla we are using, but in a follow up change, we will use a different `RELOC_VERSION` for testing newer scylla.


tchaikov:
@scylladb/qa-maintainers hello qa-maintainers, could you help review this change?

fruch (approved):
LGTM

PR #567, opened 2024-03-25
github/workflows/integration-tests: extract RELOC_VERSION as an env variable

tchaikov:
encode `RELOC_VERSION` in the key of binary, so we can use different binary for testing. there is chance that newer ccm would like to use newer scylla executables for testing.

this series is more a refactory, and it does not change the version of scylla we are using, but in a follow up change, we will use a different `RELOC_VERSION` for testing newer scylla.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -10,6 +10,9 @@ on:
     branches:
      - next*
 
+env:
+  RELOC_VERSION: "2023-04-13T13:31:00Z"
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM

fruch:
@tchaikov please check why CI didn't run at all

denesb (approved):

tchaikov:
changelog:

- because it's not allowed to define an env var with another env var in github workflow, apply the fix like #568

tchaikov:
> @tchaikov please check why CI didn't run at all

@fruch sorry, it was a regression. should be fixed by #568

tchaikov:
@fruch ping?

tchaikov:
next step, i will extract the setting of scylla version into another yaml file as a dedicated action shared by these two existing action, so we don't need to repeat it in two workflows.

PR #569, opened 2024-03-27
github/workflows: extract RELOC_VERSION into a workflow

tchaikov:
instead of repeating the RELOC_VERSION in two different workflows, let's extract it into a workflow which will drive the existing two workflow and feed them with the RELOC_VERSION.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -1,0 +1,23 @@
+name: ci
+
+on:
+  push:
+    branches:
+     - master
+     - next*
+
+  pull_request:
+    branches:
+     - next*
+
+jobs:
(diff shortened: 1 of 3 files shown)


tchaikov:
the "Expected — Waiting for status to be reported" messages are expected. as these two jobs have been renamed. we would need to update the repo settings in github to reflect this change.

tchaikov:
@scylladb/qa-maintainers hello maintainers, could you help review this change? i intended to improve the maintainability of the workflow by consolidating the triggering criteria and by moving the cached scylla version to a single file.

roydahan (approved):

fruch (approved):
LGTM

PR #570, opened 2024-03-27
common: treat code compiled with coverage as debug

fruch:
since there are some slowdowns with coverage enabled binary we temporary treat it as debug, so all test that were tuned down for debug in dtest, would be have the same as they do for debug.

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -661,6 +661,12 @@ def scylla_extract_install_dir_and_mode(install_dir):
                 f.close()
             except:
                 pass
+
+    is_gcov = subprocess.run(['bash', '-c', f'strings {install_dir}/libexec/scylla | grep -i GCOV_'], stderr=subprocess.DEVNULL,
+                  stdout=subprocess.DEVNULL, check=False)
+    if is_gcov.returncode == 0:
+        scylla_mode = "debug"
+
     return install_dir, scylla_mode
 
 


fruch:
@benipeled 

feel free to take it for a ride in jenkins

I've test this locally run test_bulk_round_trip_with_timeouts x30 in parralell 

```
pytest --scylla-version=cov --log-cli-level=debug -n 3 --count=30 cqlsh_tests/cqlsh_copy_tests.py::TestCqlshCopy::test_bulk_round_trip_with_timeouts 
```

before this change 80% of the runs would fail, after it all of them passed

roydahan:
LGTM, but now I'm rethinking if we need to run in CI both this and the regular dtest.
My main concern is that something pass CI like this and then will fail in next for the regular run.

fruch:
> LGTM, but now I'm rethinking if we need to run in CI both this and the regular dtest. My main concern is that something pass CI like this and then will fail in next for the regular run.

The assumption the coverage doesn't slow down anything, is probably wrong, and you'll still have this issue even if we'll move this part into dtest and do that for specific places

roydahan:
> The assumption the coverage doesn't slow down anything, is probably wrong, and you'll still have this issue even if we'll move this part into dtest and do that for specific places

There was not such assumption, the assumption is that the tests should handle or be able to run on any machine, slow or fast.
We know it's not the case for all existing tests, but once we introduce this feature in CI, it'll start "enforcing" it for people.

benipeled:
@fruch seems like it works for `test_bulk_round_trip_with_timeouts` [0] but raised a few other failures
https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/2192

https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/2192/testReport/cqlsh_tests.cqlsh_copy_tests/TestCqlshCopy/Tests___dtest___test_bulk_round_trip_with_timeouts/

fruch:
@benipeled move this code into dtest, closing

PR #571, opened 2024-04-29
ccmlib: drop cassandra-cli and thrift support

tchaikov:
cassandra-cli was removed in Cassandra 2.2.0. and it was replaced by
cqlsh in newer releases. actually, scylla-tools does not build or
package this command line tool at all. so let's drop the related helpers
for accessing this tool. 

thrift was deprecated in both ScyllaDB and Cassandra. so there is
no need to pass it around. to be compatible with existing users of
the ccmlib, the thrift related parameters are preserved in public
intefaces, but they are not passed down anymore. in some places,
we enforce that the host on which thrift protocol is served should
be identical to that of binary. and actually, scylla always use
the same host address for both thrift and binary protocols. so we
replace the address like `self.network_interfaces['thrift'][0]`
with `self.network_interfaces['binary'][0]`.

Refs https://github.com/scylladb/scylladb/issues/3811
Refs https://github.com/scylladb/scylladb/issues/18416

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ After that execute:
 
     ccm start
 
-That will start 3 nodes on IP 127.0.0.[1, 2, 3] on port 9160 for thrift, port
+That will start 3 nodes on IP 127.0.0.[1, 2, 3] on port 9042 for native transport, port
 7000 for the internal cluster communication and ports 7100, 7200 and 7300 for JMX.
 You can check that the cluster is correctly set up with
 
@@ -297,9 +297,9 @@ how to use ccmlib follows:
     cluster.populate(3).start()
     [node1, node2, node3] = cluster.nodelist()
 
-    # do some tests on the cluster/nodes. To connect to a node through thrift,
+    # do some tests on the cluster/nodes. To connect to a node through native protocol,
     # the host and port to a node is available through
-    #   node.network_interfaces['thrift']
+    #   node.network_interfaces['binary]
 
     cluster.flush()
     node2.compact()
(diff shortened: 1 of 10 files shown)


tchaikov:
```
ERROR tests/test_scylla_cmds.py::TestCCMClusterSniProxy::test_cqlsh[ccm_reloc_cluster] - subprocess.TimeoutExpired: Command '['./ccm', 'start', '--sni-proxy', '--sni-port', '8443', '--wait-for-binary-proto']' timed out after 600 seconds
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
```

still looking.

tchaikov:
@scylladb/qa-maintainers hi maintainers, could you help review this change?

fruch (approved):

fruch:
It might have some falloffs on drivers tests, but will address then there

tchaikov:
> It might have some falloffs on drivers tests, but will address then there

@fruch hi Israel, thank you for your review and approval. do you mean the tests in dtest?

fruch:
> > It might have some falloffs on drivers tests, but will address then there
> 
> @fruch hi Israel, thank you for your review and approval. do you mean the tests in dtest?

Not dtest, other users of ccm are the java and python drivers, but we probably should skip thrift even if used in those

tchaikov:
thanks. i see. since `start_rpc` option is off by default, see https://github.com/scylladb/scylladb/blob/af0c0ee8af48f1691a8f8a379a442da5eeeb5f8a/db/config.cc#L802-L803. and i searched 

- python-driver: https://github.com/search?q=repo%3Ascylladb%2Fpython-driver+start_rpc&type=code
- https://github.com/search?q=repo%3Ascylladb%2Fjava-driver%20start_rpc&type=code

neither of them actually enables this option. so i think we are safe.

PR #572, opened 2024-04-30

tchaikov:

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -776,16 +776,23 @@ class ScyllaNode(Node):
             self.jmx_pid = None
 
     def nodetool(self, cmd, capture_output=True, wait=True, timeout=None, verbose=True):
-        if not self.has_jmx:
+        try:
+            # nodetool subcommand was added in 5.5, so check if it is available
+            # before using it.
+            nodetool = [os.path.join(self.get_bin_dir(), "scylla"), "nodetool"]
+            subprocess.check_call(nodetool + ["help"], stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT)
             if self.is_docker():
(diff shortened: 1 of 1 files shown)


denesb (approved):

tchaikov:
changelog

- silence the output from the stdout and stderr

tchaikov:
this change should addrses the test failure like https://jenkins.scylladb.com/job/scylla-master/job/byo/job/dtest-byo/271/testReport/junit/nodetool_additional_test/TestNodetool/FullDtest___full_split001___test_describering/

tchaikov:
@scylladb/qa-maintainers hello maintainers, could you help review and merge this change?

bhalevy (approved):
LGTM

Cc @denesb @fruch

fruch (approved):
LGTM

fruch:
@tchaikov 

had to yank this out of next,

see: https://jenkins.scylladb.com/job/scylla-master/job/next/7565/testReport/junit/

soyacz:
> @tchaikov
> 
> had to yank this out of next,
> 
> see: https://jenkins.scylladb.com/job/scylla-master/job/next/7565/testReport/junit/

I think this uncovers some issues with the new nodetool (see https://github.com/scylladb/scylladb/issues/18468)

denesb:
Looks like there is yet another missing native nodetool command: resetlocalschema. I'm working on the implementation.

tchaikov:
> @tchaikov
> 
> had to yank this out of next,
> 
> see: https://jenkins.scylladb.com/job/scylla-master/job/next/7565/testReport/junit/

thank you @fruch . will implement a workaround on dtest first, see https://github.com/scylladb/scylla-dtest/pull/4204. and then try to get this in after it lands.

PR #573, opened 2024-05-01

tchaikov:

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -858,8 +858,17 @@ class Node(object):
         if not isinstance(nodetool, list):
             nodetool = [nodetool]
         # see https://www.oracle.com/java/technologies/javase/8u331-relnotes.html#JDK-8278972
-        nodetool.extend(['-h', host, '-p', str(self.jmx_port), '-Dcom.sun.jndi.rmiURLParsing=legacy'])
-        nodetool.extend(cmd.split())
+        args = ['-h', host, '-p', str(self.jmx_port), '-Dcom.sun.jndi.rmiURLParsing=legacy']
+        if len(nodetool) > 1:
+            nodetool.extend(cmd.split())
+            # put args at the end of command line options, as "scylla nodetool"
+            # expects them as options of the subcommand
+            nodetool.extend(args)
+        else:
+            # while java-based tool considers them as options of the nodetool
+            # itself
+            nodetool.extend(args)
+            nodetool.extend(cmd.split())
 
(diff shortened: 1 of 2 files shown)


mykaul:
That reminds me of the functionality of the wrapper script we've had. Why not use it?

tchaikov:
@mykaul hi Yaniv, as @fruch put, it was dropped. see https://github.com/scylladb/scylla-tools-java/pull/386/files

tchaikov:
this change is pending on a up-to-date build. as https://jenkins.scylladb.com/job/scylla-staging/job/tchaikov/job/gating-dtest-release/9/testReport/junit/nodetool_additional_test/TestNodetool/test_resetlocalschema_nodetool_1_2_/ was still testing a commit without https://github.com/scylladb/scylladb/pull/18470

tchaikov:
tested at https://jenkins.scylladb.com/job/scylla-staging/job/tchaikov/job/gating-dtest-release/11/

tchaikov:
@denesb  and @fruch could you help review this change?

tchaikov:
it's a resurrection of #572. which was queued but then dequeued due to a regression. but the regression was addressed by https://github.com/scylladb/scylladb/pull/18470.

fruch (approved):
LGTM

bhalevy:
Was https://github.com/scylladb/scylla-ccm/commit/7d067c039951772a33a7a04b4dc4c760913e0d5e dequeued?

bhalevy:
@tchaikov please retest all next_gating tests and resubmit

tchaikov:
> @tchaikov please retest all next_gating tests and resubmit

i am actually testing it at https://jenkins.scylladb.com/job/scylla-staging/job/tchaikov/job/gating-dtest-release/16/. but some tests are failing even without this patch. see https://jenkins.scylladb.com/job/scylla-staging/job/tchaikov/job/gating-dtest-release/17/ . still looking.

bhalevy:
> > @tchaikov please retest all next_gating tests and resubmit
> 
> i am actually testing it at https://jenkins.scylladb.com/job/scylla-staging/job/tchaikov/job/gating-dtest-release/16/. but some tests are failing even without this patch. see https://jenkins.scylladb.com/job/scylla-staging/job/tchaikov/job/gating-dtest-release/17/ . still looking.

Apparently it's due to the https://github.com/scylladb/scylla-dtest/pull/4186 fiasco.
The test doesn't recognize that consistent-topology-changes is on (now by default).

bhalevy:
@tchaikov please also test rolling_upgrade_test.py

PR #574, opened 2024-05-06

tchaikov:

also, the native nodetool expect `-h` and `--port` as parameters of the subcommand, not the parameters of `scylla nodetool`, it complains like:

```
E               ccmlib.node.ToolError: Subprocess /jenkins/workspace/scylla-master/gating-dtest-release/scylla/.dtest/dtest-7r2n368l/test/node1/bin/scylla nodetool -h 127.0.17.1 -p 10000 snapshot exited with non-zero status; exit status: 100;
E               stderr: error: unrecognized operation argument: expected one of ({cleanup, compact, disablebackup, disablebinary, disablegossip, enablebackup, enablebinary, enablegossip, flush, gettraceprobability, help, settraceprobability, statusbackup, statusbinary, statusgossip, version}), got --host
```

but the java-based tool looks for these options before the subcommand, so put them conditionally.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -858,8 +858,17 @@ class Node(object):
         if not isinstance(nodetool, list):
             nodetool = [nodetool]
         # see https://www.oracle.com/java/technologies/javase/8u331-relnotes.html#JDK-8278972
-        nodetool.extend(['-h', host, '-p', str(self.jmx_port), '-Dcom.sun.jndi.rmiURLParsing=legacy'])
-        nodetool.extend(cmd.split())
+        args = ['-h', host, '-p', str(self.jmx_port), '-Dcom.sun.jndi.rmiURLParsing=legacy']
+        if len(nodetool) > 1:
+            nodetool.extend(cmd.split())
+            # put args at the end of command line options, as "scylla nodetool"
+            # expects them as options of the subcommand
+            nodetool.extend(args)
+        else:
+            # while java-based tool considers them as options of the nodetool
+            # itself
+            nodetool.extend(args)
+            nodetool.extend(cmd.split())
 
         return self._do_run_nodetool(nodetool, capture_output, wait, timeout, verbose)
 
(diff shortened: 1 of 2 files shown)


tchaikov:
- dtest-full:  https://jenkins.scylladb.com/job/scylla-staging/job/tchaikov/job/gating-dtest-release/18/

tchaikov:
@fruch and @yaronkaikov could you update the git repo setting to make the "ci / CCM Integration tests / build (3.8) " and ci / CCM Integration tests / build (3.11)" as the "Require status checks to pass before merging", and deselect the stale ones?

fruch:
> @fruch and @yaronkaikov could you update the git repo settings to make the "ci / CCM Integration tests / build (3.8) " and ci / CCM Integration tests / build (3.11)" as the "Require status checks to pass before merging", and deselect the stale ones?

updated it

fruch (approved):
LGTM

bhalevy:
@tchaikov it looks like this patch is insufficient since scylla nodetool returns status 1 with `--help`, rather than 0.

Cc @denesb

fruch:
> @tchaikov it looks like this patch is insufficient since scylla nodetool returns status 1 with `--help`, rather than 0.
> 
> Cc @denesb 

I think it was fixed in
https://github.com/scylladb/seastar/pull/2213

Maybe you are running code that didn't got it yet ?

bhalevy:
> > @tchaikov it looks like this patch is insufficient since scylla nodetool returns status 1 with `--help`, rather than 0.
> > Cc @denesb
> 
> I think it was fixed in [scylladb/seastar#2213](https://github.com/scylladb/seastar/pull/2213)
> 
> Maybe you are running code that didn't got it yet ?

It's not me apparently. It's everybody, since the seastar submodule in scylla wasn't updated yet.

bhalevy:
Ref https://github.com/scylladb/scylladb/pull/18555

bhalevy:
@tchaikov next time please take care of the papertrail and document the dependencies and a merge plan, so we can make sure that changes in different repos are upstreamed in the right order.

PR #575, opened 2024-05-06
ccmlib/scylla_node: dump_sstable_stats to handle broken utf-8

fruch:
in some cases the output of `scylla sstable` can be with illegal utf-8, we should just skip them, and it would work as expected.


tchaikov (approved):
lgtm

PR #576, opened 2024-05-08
Remove cli cluster command

Lorak-mmk:
It's implementation was already removed in https://github.com/scylladb/scylla-ccm/commit/06a2adf303c292239b9a84b6a674824e0204797c , but the command was still present in the list, causing an error when printing ccm's help.


tchaikov (approved):
lgtm

fruch (approved):
LGTM

PR #578, opened 2024-06-05
ccmlib/utils: add inotify wrapper

tchaikov:
a typical session looks like:

```console
$ while true; do
    sleep 1
    echo 1 >> /tmp/foobar
  done
```

in the meanwhile:

```py
from inotify import INotify, Flag

notify = INotify()
with notify.watch("/tmp/foobar", Flag.MODIFY):
  for e in notify.read():
    print(e)
```

### DIFF
--- a/ccmlib/utils/inotify.py
+++ b/ccmlib/utils/inotify.py
@@ -1,0 +1,137 @@
+import errno
+import fcntl
+import io
+import os
+import select
+import struct
+
+from contextlib import contextmanager
+from ctypes import CDLL, get_errno, c_int
+from ctypes.util import find_library
+from dataclasses import dataclass
+from enum import IntEnum
+
+from termios import FIONREAD
+from time import sleep
+from typing import Callable, Generator, List, Optional, Tuple
+
+
+class Libc:
+    """a delegate class for accessing libc functions not exposed by CPython
+    """
+    def __init__(self) -> None:
+        self.libc = CDLL(find_library('c'), use_errno=True)
+
+    @staticmethod
+    def _retry_on_eintr(func: Callable) -> Callable:
+        """Wrapper around a libc function and retries on EINTR."""
+        def wrapper(*args, **kwargs):
+            while True:
+                rc = func(*args, **kwargs)
+                if rc != -1:
+                    return rc
+                ec = get_errno()
+                if ec != errno.EINTR:
+                    raise OSError(errno, os.strerror(ec))
+
+        return wrapper
+
+    def __getattr__(self, name: str) -> Callable:
+        return self._retry_on_eintr(getattr(self.libc, name))
+
+
+class Flag(IntEnum):
+    """IN_* flags"""
+    ACCESS = 0x00000001
+    MODIFY = 0x00000002
+    ATTRIB = 0x00000004
+    CLOSE_WRITE = 0x00000008
+    CLOSE_NOWRITE = 0x00000010
+    OPEN = 0x00000020
+    MOVED_FROM = 0x00000040
+    MOVED_TO = 0x00000080
+    CREATE = 0x00000100
+    DELETE = 0x00000200
+    DELETE_SELF = 0x00000400
+    MOVE_SELF = 0x00000800
+
+    UNMOUNT = 0x00002000
+    Q_OVERFLOW = 0x00004000
+    IGNORED = 0x00008000
+
+    ONLYDIR = 0x01000000
+    DONT_FOLLOW = 0x02000000
+    EXCL_UNLINK = 0x04000000
+    MASK_ADD = 0x20000000
+    ISDIR = 0x40000000
+    ONESHOT = 0x80000000
+
+
+@dataclass
+class Event:
+    """represents a notification"""
+    wd: int
+    flags: List[Flag]
+    cookie: int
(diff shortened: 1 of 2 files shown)


tchaikov:
inspired by #577

fruch:
@tchaikov 

can you add a test, or some example of usage for this, it's not exactly clear what exactly do we need it for...

for using in polling log files ?

tchaikov:
> @tchaikov
> 
> can you add a test, or some example of usage for this, it's not exactly clear what exactly do we need it for...

indeed. will do.

> 
> for using in polling log files ?

yes, it's exactly the intended use case.

tchaikov:
v2:

- added tests for inotify wrapper.

fruch (approved):
LGTM

fruch:
@tchaikov 

I'll merge it once tests pass

anyhow with the next steam, we need to be careful, and make sure we run all tests (all variant, and enterprise)
since it something that affect all of the tests

tchaikov:
> need to be careful with type annotations, we still support python 3.8 (cause of usage in drivers)
> 
> ```
> tests/test_utils_inotify.py:1: in <module>
>     from ccmlib.utils.inotify import INotify, Flag
> ccmlib/utils/inotify.py:71: in <module>
>     class Event:
> ccmlib/utils/inotify.py:74: in Event
>     flags: list[Flag]
> E   TypeError: 'type' object is not subscriptable
> ```

ack.

tchaikov:
v2:

- use typing annotations compatible with Python 3.8

nyh:
> > for using in polling log files ?
> 
> yes, it's exactly the intended use case.

Inotify is useful when you have a lot of files, and not sure when any of them will change - it's possible that a file will change in the next second, or maybe it won't change in a year. This isn't usually the case in log files - you have just one or a small number of those (one per node), and usually each of them changes all the time (our logs are very chattery). So we need to constantly read from each of them looking if the log message we cared about appeared.

Moreover in a modern machine, code which looks like:

```
while True:
   check_the_end_of_the_log()
   sleep(0.1)
```
Is very efficient (there is a file system cache) and doesn't really needs optimization.

Can you give an example of a problem - usability or performance problem - that you're trying to solve?

fruch:
> > > for using in polling log files ?
> > 
> > 
> > yes, it's exactly the intended use case.
> 
> Inotify is useful when you have a lot of files, and not sure when any of them will change - it's possible that a file will change in the next second, or maybe it won't change in a year. This isn't usually the case in log files - you have just one or a small number of those (one per node), and usually each of them changes all the time (our logs are very chattery). So we need to constantly read from each of them looking if the log message we cared about appeared.
> 
> Moreover in a modern machine, code which looks like:
> 
> ```
> while True:
>    check_the_end_of_the_log()
>    sleep(0.1)
> ```
> 
> Is very efficient (there is a file system cache) and doesn't really needs optimization.
> 
> Can you give an example of a problem - usability or performance problem - that you're trying to solve?

I think the issue at hand was the the loop we have was doing that loop with sleeps which were hardcoded
and were too big for on specific test, now those times can be configured.

I think @tchaikov is aiming into something, that would need those sleeps at all. 
I don't think we were having performance issue at hand, or having big amount of files to be watched

but yes, might be a bit of an overkill.
also, we should be think/moving to ways to test things with depending on the logs output. and maybe no entrancing more into it.
so far we did succeeded much in it, and every boot has a mix flavor of using API calls and waiting for logs, for all bunch of compatibility reasons.

tchaikov:
@fruch @nyh do we have any objections? this PR was approved. and i made changes per the request of reviewer.

fruch:
> @fruch @nyh do we have any objections? this PR was approved. and i made changes per the request of reviewer. 

There is still the concern it might be an overkill 

But I'm willing to experiment on how it would be used and as said , we'll need to run all of the tests to validate

nyh:
> I think the issue at hand was the the loop we have was doing that loop with sleeps which were hardcoded and were too big for on specific test, now those times can be configured.

Sleeps are a problem in a lot of dtests, but it has nothing to do with **logs**. As I mentioned, because we enable a lot of crap in our log, our logs constantly get written, so inotify will return almost immediately anyway. If we use sleep(0.1) when looking at logs, that would have been good enough - nobody would have complained the sleep was too short or too long. I think inotify is a complete overkill for this use case. But I guess it's not wrong.

On the other hand, sleep in log searching is good if you iteratively keep reading from the **same** mark (and not always read just the new data from the last read). In that case, waking up every millisecond and reading all the data from the original mark is a big waste. If we expect the message to appear only after 5 seconds (usually), then sleeping 5 seconds and then checking is much more efficient. Of course even better is to read in every iteration just from the end of the previous iteration. But this must be done carefully (you need to at least cut the log on full lines and not in the middle of a line), and I don't know if we're careful in that. @fruch maybe you know.

fruch:
@nyh 

All is correct, but someone did complain about 0.1 sleep is too long, for some use case, and we introduced a way to lower it down for that specific test.

That thread ignited @tchaikov direction here.

As I've said I'm o.k. to give it a try, it might prove useful in more place.

Nowadays we have quite good ways to test such a thing with all tests, i.e. we have quite stright criteria for such a change (that all tests pass :))

PR #579, opened 2024-06-06
ccmlib/common: print out process if it's listen given address

tchaikov:
when we start, for instance, a scylla node, we use `check_socket_available()` to see if the given address can be bound, if not an exception is raised. but we still have no idea who is listening on this address.

in this change, we print out the process in this case, so that we can understand the problem better. in general, it's either an infra issue, or, it could be caused by a buggy test.

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -16,11 +16,13 @@ import time
 import tempfile
 import logging
 import pathlib
+from datetime import datetime
 from itertools import zip_longest
 from typing import Callable, Optional, TextIO, Union, List
 from pathlib import Path
 
 import yaml
+import psutil
 from boto3.session import Session
 from botocore import UNSIGNED
 from botocore.client import Config
@@ -721,6 +723,32 @@ def validate_install_dir(install_dir):
             raise ArgumentError(f'{install_dir} does not appear to be a cassandra or dse installation directory')
 
 
+def pid_listening_on(addr: str, port: int):
+    '''get pid of the process listening on the given port'''
+    pid = -1
+    if lsof := shutil.which('lsof'):
+        result = subprocess.run([lsof, '-t', '-i', f'tcp@{addr}:{port}'],
+                                stdout=subprocess.PIPE,
+                                check=True)
+        if result.stdout:
+            pid = int(result.stdout)
+    elif ss := shutil.which('ss'):
+        result = subprocess.run([ss,
+                                 '--no-header',
+                                 '--listening',
+                                 '--processes',
+                                 '--tcp',
+                                 f'src = inet:{addr}:{port}'],
+                                stdout=subprocess.PIPE,
+                                check=True)
+        matched = re.search(r'pid=(\d+)', result.stdout)
+        if matched is not None:
+            pid = int(matched.group(1))
+    else:
+        logger.info("neither lsof nor ss was found")
+    return pid
+
+
 def check_socket_available(itf):
     info = socket.getaddrinfo(itf[0], itf[1], socket.AF_UNSPEC, socket.SOCK_STREAM)
     if not info:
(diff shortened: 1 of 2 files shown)


tchaikov:
this change was inspired by https://jenkins.scylladb.com/job/scylla-6.0/job/next/47/testReport/junit/nodetool_additional_test/TestNodetool/Build___x86___dtest_with_topology_changes___test_status/

fruch (commented):
I don't know how useful it would be, if we don't log the scylla processes 
when scylla starts

tchaikov:
> I don't know how useful it would be, if we don't log the scylla processes when scylla starts
> 
> we really want to identify which test that process belong to, we know it's scylla...

does the parent process helps?

tchaikov:
> @tchaikov
> 
> I would recommend giving it a ride, at least in gating, to see it's not breaking anything

running at https://jenkins.scylladb.com/job/scylla-master/job/byo/job/dtest-byo/342/

fruch:
> > I don't know how useful it would be, if we don't log the scylla processes when scylla starts
> > we really want to identify which test that process belong to, we know it's scylla...
> 
> does the parent process help? also, the timestamp should help.

not much as well

in this situation you have a rouge scylla process up (my guess)
you have 5 test processes running in the same time, dtest/ccm should have killed the cluster at the end of the test

each time a scylla process gonna start there this print:
```
21:04:15,154 741     ccm                            DEBUG    cluster.py          :754  | test_load_older_snapshot_and_refresh: node1: Starting scylla: args=['/jenkins/workspace/scylla-6.0/gating-dtest-release-with-consistent-topology-changes/scylla/.dtest/dtest-7hc6kw6e/test/node1/bin/scylla', '--options-file', '/jenkins/workspace/scylla-6.0/gating-dtest-release-with-consistent-topology-changes/scylla/.dtest/dtest-7hc6kw6e/test/node1/conf/scylla.yaml', '--log-to-stdout', '1', '--abort-on-seastar-bad-alloc', '--abort-on-lsa-bad-alloc', '1', '--abort-on-internal-error', '1', '--api-address', '127.0.51.1', '--smp', '2', '--memory', '1024M', '--developer-mode', 'true', '--default-log-level', 'info', '--overprovisioned', '--prometheus-address', '127.0.51.1', '--unsafe-bypass-fsync', '1', '--kernel-page-cache', '1', '--commitlog-use-o-dsync', '0', '--max-networking-io-control-blocks', '1000'] wait_other_notice=True wait_for_binary_proto=True
``

you need a way to correlate between test failure listing on a specific address (and you have the process id of who holds it)
to the test started that scylla process, so you can investigate why clearing up of the process didn't work in that test.

tchaikov:
@fruch i think this is the at least what we can do at this moment. if there is anything else i can do, i would be happy to make it happen.

fruch:
> @fruch i agree this is not a turn-key solution, but i think this is at least what we can do at this moment. if there is anything else i can do, i would be happy to make it happen.

I think adding a print of the scylla process, when we start it, would make this PR a bit more useful, and we would be able to hunt down which test have start the process

fruch:
> v2:
> 
> * print out scylla instance's pid after launching it.
> 
> @fruch could you take another look?

I don't see this log print... it's on it's own commit ?

tchaikov:
> > v2:
> > 
> > * print out scylla instance's pid after launching it.
> > 
> > @fruch could you take another look?
> 
> I don't see this log print... it's on it's own commit ?

sorry, it was on another branch. now included in this PR.

fruch:
> > > v2:
> > > 
> > > * print out scylla instance's pid after launching it.
> > > 
> > > @fruch could you take another look?
> > 
> > 
> > I don't see this log print... it's on it's own commit ?
> 
> sorry, it was on another branch. now included in this PR.

yeah looks better

fruch (approved):
LGTM

PR #580, opened 2024-06-12
ccmlib/scylla_node: use reachable link

tchaikov:
before this change, we are using an unreachable hyperlink in the docstring.

after this change, the unreachable hyperlink is replaced with a reachable one.


fruch (approved):
LGTM

PR #581, opened 2024-06-18
ComparableScyllaVersion: improve matching 

tchaikov:
so that we can compare a given ComparableScyllaVersion by its major version, minor version or patch version, when we are interested in features introduced at different major version, minor version, or a certain patch version.

### DIFF
--- a/ccmlib/utils/version.py
+++ b/ccmlib/utils/version.py
@@ -28,16 +28,16 @@ SEMVER_REGEX = re.compile(
 class ComparableScyllaVersion:
     """Accepts and compares known 'non-semver' and 'semver'-like Scylla versions."""
 
-    def __init__(self, version_string: str):
+    def __init__(self, version_string: str) -> None:
         parsed_version = self.parse(version_string)
-        self.v_major = int(parsed_version[0])
-        self.v_minor = int(parsed_version[1])
-        self.v_patch = int(parsed_version[2])
-        self.v_pre_release = parsed_version[3] or ''
-        self.v_build = parsed_version[4] or ''
+        self.v_major = int(parsed_version["major"])
+        self.v_minor = int(parsed_version["minor"])
+        self.v_patch = int(parsed_version["patch"])
+        self.v_pre_release = parsed_version["prerelease"] or ''
+        self.v_build = parsed_version["build"] or ''
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


tchaikov:
@scylladb/qa-maintainers hello maintainers, could you help review this changeset?

vponomaryov (approved):
LGTM

fruch (approved):

fruch:
@tchaikov 


We don't have mergify setup for this repo

But keep in mind that if using this in backports, you'll need to backport this  too

ccmlib/utils/version.py
```diff
@@ -28,16 +28,16 @@
...
     """Accepts and compares known 'non-semver' and 'semver'-like Scylla versions."""
 
-    def __init__(self, version_string: str):
+    def __init__(self, version_string: str) -> None:
         parsed_version = self.parse(version_string)
-        self.v_major = int(parsed_version[0])
-        self.v_minor = int(parsed_version[1])
-        self.v_patch = int(parsed_version[2])
-        self.v_pre_release = parsed_version[3] or ''
-        self.v_build = parsed_version[4] or ''
```
bhalevy, line 37: In addition to the commit message, this had a bug if "prerelease" is missing and "build" exists, then it would have been in position 3, not 4.

ccmlib/utils/version.py
```diff
@@ -81,6 +81,15 @@ def parse(version_string: str) -> re.Match:
             f"Cannot parse provided '{version_string}' scylla_version for the comparison. "
             f"Transformed scylla_version: {_scylla_version}")
 
+    def major(self) -> 'ComparableScyllaVersion':
```
bhalevy, line 84: I'm curious why the return type is quoted?

fruch: It can be removed, if using `from __future__ import annotations`

Which was supposed to the default in python 3.10, but it's not yet the default in any release

See for more details:
https://stackoverflow.com/questions/66734640/any-downsides-to-using-from-future-import-annotations-everywhere

ccmlib/utils/version.py
fruch: The original code has a set of regex and "fix" that align the version to a strict semver format, so probably so of the assumptions here counted on that.

Issue #582, opened 2024-06-18
add a helper to extract config option from packaged scylla.yaml


tchaikov:
we need this feature to tell, for instance, if tablets are enabled by default.

PR #583, opened 2024-06-18
ccmlib/common: add function to get the default scylla.yaml

fruch:
this function need to get the install_dir, and out of it it can extract the default scylla.yaml (i.e. before any modification that ccm would be doing to it)

Closes: #582

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -843,6 +843,10 @@ def get_version_from_build(install_dir=None, node_path=None):
                     return match.group(1)
     raise CCMError("Cannot find version")
 
+def get_default_scylla_yaml(install_dir):
+    scylla_yaml_path = Path(install_dir) / SCYLLA_CONF_DIR / SCYLLA_CONF
+    with scylla_yaml_path.open() as f:
+        return yaml.safe_load(f)
 
 def _get_scylla_version(install_dir):
     scylla_version_files = [
--- a/tests/test_scylla_relocatable_cluster.py
+++ b/tests/test_scylla_relocatable_cluster.py
@@ -3,7 +3,7 @@ import subprocess
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


ccmlib/common.py
```diff
@@ -843,6 +843,10 @@ def get_version_from_build(install_dir=None, node_path=None):
                     return match.group(1)
     raise CCMError("Cannot find version")
 
+def get_default_scylla_yaml(install_dir):
+    scylla_yaml_path = Path(install_dir) / SCYLLA_CONF_DIR / SCYLLA_CONF
+    with scylla_yaml_path.open(mode='r') as f:
```
tchaikov, line 848: nit, `open()` passes `mode='r'` by default.

tchaikov (approved):
lgtm

tests/test_scylla_relocatable_cluster.py
```diff
@@ -65,3 +65,12 @@ def test_cqlsh(self, relocatable_cluster):
         for s in ['(1 rows)', 'key', '1']:
             assert s in rv[0]
         assert rv[1] == ''
+
+
+    def test_get_default_scylla_yaml(self, relocatable_cluster):
+        install_dir = relocatable_cluster.get_install_dir()
+        scylla_yaml = get_default_scylla_yaml(install_dir)
+        assert 'enable_tablets' not in scylla_yaml
```
tchaikov, line 73: i'd use a more stable option in this file, as in new reloc packages, this does not hold anymore.

ccmlib/common.py
fruch: fixed

tests/test_scylla_relocatable_cluster.py
fruch: fixed

PR #584, opened 2024-07-01
ccmlib/cluster: fix automatic rack naming

enaydanov:
Make Cluster.populate() work correctly for option 3.b of the nodes argument (e.g., {"dc1": [2, 2]}): as described in the comment it'll be converted to something like {"dc1": {"RAC1": 2, "RAC2": 2}}

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -283,7 +283,7 @@ class Cluster(object):
                 if isinstance(x, int):
                     topology[dc] = OrderedDict([(None, x)])
                 elif isinstance(x, list):
-                    topology[dc] = OrderedDict([(None, n) for n in x])
+                    topology[dc] = OrderedDict([(f"RAC{i}", n) for i, n in enumerate(x, start=1)])
                 elif isinstance(x, dict):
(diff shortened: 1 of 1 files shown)


enaydanov:
@bhalevy @fruch ping

fruch (approved):
LGTM,

even that there way too many options with this parameter, and would benefit from having unittest for that part.

bhalevy (approved):
LGTM, thanks.

The title says "fix", is there a reproducer for the current implementation failure?
I don't oppose this change, by the reason I used `None` initially is to let the existing code select the default rack names.

enaydanov:
> The title says "fix", is there a reproducer for the current implementation failure? I don't oppose this change, by the reason I used `None` initially is to let the existing code select the default rack names.

I'm trying to use it here: https://github.com/scylladb/scylla-dtest/pull/4575 , but it looks like I'm a first user of this code :)

There are two problems with initial approach to "select the default rack names":

1. I can't find a code which generates these "default rack names" in scylla-ccm
2. It sets `None` as a key for all items in `OrderedDict()`:

```python
>>> from collections import OrderedDict
>>> OrderedDict([(None, x) for x in range(5)])
OrderedDict([(None, 4)])
>>>
```

PR #585, opened 2024-07-03
node: stress_process: specifiy -schema replication(strategy=NetworkTopologyStrategy) by default

bhalevy:
Currently stress creates the keyspace with SimpleReplication which is deprecated and inhibits tablets.

This change passes the strategy=NetworkTopologyStrategy replication option by default, if it is not otherwise specified.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1311,7 +1311,41 @@ class Node(object):
                 if not [opt for opt in stress_options if opt.startswith('jmx=')]:
                     stress_options.extend(['-port', 'jmx=' + self.jmx_port])
 
+        replication_re = re.compile(r"(?P<key>replication\s*)\((?P<opt>[^)]*)\)")
+        profile_re = re.compile(r"(?P<key>profile\s*=\s*)(?P<opt>[^ ]+)")
+        profile_match = None
+        replication_match = None
+        schema_idx = None
+        for i, arg in enumerate(stress_options):
+            profile_match = profile_re.search(arg)
+            if profile_match:
+                self.warning(f"No support for adjusting replication strategy in user profiles yet: {arg}")
+                break
+            if schema_idx is None and "-schema" in arg:
+                schema_idx = i
+            replication_match = replication_re.search(arg)
+            if replication_match:
+                opt = replication_match.group("opt")
+                if "strategy" in opt:
+                    if not "NetworkTopologyStrategy" in opt:
+                        self.warning(f"{arg} does not use NetworkTopologyStrategy for stress")
+                else:
+                    try:
+                        stress_options[i] = replication_re.sub(r"\g<key>(strategy=NetworkTopologyStrategy,\g<opt>)", arg)
+                    except Exception as e:
+                        self.warning(f"Error while trying to add 'strategy=NetworkTopologyStrategy' to stress replication option: {e}")
+                        pass
+                break
+        if not profile_match:
+            if not replication_match:
+                opt = "replication(strategy=NetworkTopologyStrategy,factor=1)"
+                if schema_idx is not None:
+                    stress_options.insert(schema_idx + 1, opt)
+                else:
+                    stress_options.extend(["-schema", opt])
+
         args = stress + stress_options
+        self.debug(f"Executing {args}")
         stdout_handle = kwargs.pop("stdout", subprocess.PIPE)
         stderr_handle = kwargs.pop("stderr", subprocess.PIPE)
         p = subprocess.Popen(args,


bhalevy:
For example, see the test_refresh_and_restart_after_compaction_strategy_change and test_reshard_after_compaction_strategy_and_smp_change dtests:
```python
        node1.stress(["write", "n=0", "no-warmup", "-schema", "replication(factor=1)", "-rate", "threads=1"])
```

ccmlib/node.py
```diff
@@ -1311,7 +1311,22 @@ def stress_process(self, stress_options, **kwargs):
                 if not [opt for opt in stress_options if opt.startswith('jmx=')]:
                     stress_options.extend(['-port', 'jmx=' + self.jmx_port])
 
+        if "-schema" in stress_options and "replication" in stress_options[i := stress_options.index("-schema") + 1]:
```
yarongilor, line 1314: would it work with this existing code:
```
node1.stress(stress_options=["write", "cl=ONE", "n=10000", "-schema replication(factor=1)", "-mode cql3 native", "-rate threads=10"], capture_output=True)
```
?

ccmlib/node.py
```diff
@@ -1311,7 +1311,22 @@ def stress_process(self, stress_options, **kwargs):
...
+            opt = stress_options[i]
+            if "strategy" in opt:
+                if not "NetworkTopologyStrategy":
```
yarongilor, line 1317: `"NetworkTopologyStrategy" in opt`?

ccmlib/node.py
bhalevy: hmm, it expects the replication option to be in the following element, but I guess we make it work this way as well

ccmlib/node.py
bhalevy: right, will fix, thanks

bhalevy:
@fruch what puthin version do we use ccm ci?
So we need backward compatibility?

fruch:
> @fruch what puthin version do we use ccm ci? So we need backward compatibility?

as the CI says, we support python3.8-3.12 (same compatibility of the python-driver that use ccm in it's tests)

so no usage of walrus operator

fruch:
@bhalevy 

few comments:
1) soon we'll need to extract the c-s usage out of ccm - i.e. once c-s won't be part of unified package - and it would be packaged on it's own
2) lets raise an issue in java-tools repo (until c-s would be on it's own repo) to do this on it's own by default

fruch:
@bhalevy 

also, why aren't we fixing the dtest to do what we expect ? wouldn't it be easier to reason about ?

bhalevy:
> @bhalevy
> 
> also, why aren't we fixing the dtest to do what we expect ? wouldn't it be easier to reason about ?

Not all dtest specify the schema for c-stress as they don't care.
We can do the wrapping in dtest rather than in ccm, but currently we still call `node.stress()`, that already manipulates the stress command line args, so it seems like the right place to do so.

bhalevy:
- fixed the match and replace algorithm
- and searching for `"NetworkTopologyStrategy" in opt`

bhalevy:
> @bhalevy
> 
> few comments:
> 
> 1. soon we'll need to extract the c-s usage out of ccm - i.e. once c-s won't be part of unified package - and it would be packaged on it's own
> 2. lets raise an issue in java-tools repo (until c-s would be on it's own repo) to do this on it's own by default

https://github.com/scylladb/scylla-tools-java/issues/400

bhalevy:
BTW, for dtest I'm not sure we need to use cassandra-stress at all.
We could other tools, or just provide a simple load-generator in python that would use the python driver.
I don't think there's anything in particular in scylla-stress that we need in dtest...

bhalevy:
Seen a problem in e.g. test_nodetool_snapshot_during_major_compaction.
Adding another `-schema` option for the replication strategy is not possible.
```
13:10:19,742 1838328 ccm                            DEBUG    cluster.py          :754  | test_nodetool_snapshot_during_major_compaction: node1: Executing ['/home/bhalevy/.ccm/scylla-repository/local_tarball/scylla-core-package/scylla-tools/tools/bin/cassandra-stress', 'write', 'n=1000000', '-rate', 'threads=10', '-schema', 'compaction(strategy=SizeTieredCompactionStrategy,enabled=false)', '-node', '127.0.86.1', '-port', 'jmx=7199', '-schema', 'replication(strategy=NetworkTopologyStrategy,factor=1)']
13:10:20,275 1838328 errors                         ERROR    conftest.py         :202  | test_nodetool_snapshot_during_major_compaction: test failed: 
self = <snapshot_test.TestSnapshot object at 0x7fac7c3df800>

    def test_nodetool_snapshot_during_major_compaction(self):
        def run_compaction(node):
            logger.info("Start compaction by command")
            node.compact()
            logger.info("Compaction done")
    
        cluster = self.cluster
        cluster.populate(1).start()
        node1 = cluster.nodelist()[0]
    
        ks_name = "keyspace1"
        table_name = "standard1"
    
        logger.info("Run stress command")
        num_keys = 1000000 if self.cluster.scylla_mode != "debug" else 10000
>       results = node1.stress(["write", f"n={num_keys}", "-rate", "threads=10", "-schema", "compaction(strategy=SizeTieredCompactionStrategy,enabled=false)"])

snapshot_test.py:431: 
../../.pyenv/versions/3.12.3/envs/dtest-3.12.3/lib/python3.12/site-packages/ccmlib/node.py:1342: in stress
    return handle_external_tool_process(p, ['stress'] + stress_options)

process = <Popen: returncode: 1 args: ['/home/bhalevy/.ccm/scylla-repository/local_tar...>
cmd_args = ['stress', 'write', 'n=1000000', '-rate', 'threads=10', '-schema', ...]

    def handle_external_tool_process(process, cmd_args):
        out, err = process.communicate()
        if (out is not None) and isinstance(out, bytes):
            out = out.decode()
        if (err is not None) and isinstance(err, bytes):
            err = err.decode()
        rc = process.returncode
    
        if rc != 0:
>           raise ToolError(cmd_args, rc, out, err)
E           ccmlib.node.ToolError: Subprocess ['stress', 'write', 'n=1000000', '-rate', 'threads=10', '-schema', 'compaction(strategy=SizeTieredCompactionStrategy,enabled=false)'] exited with non-zero status; exit status: 1; 
E           stdout: -schema is defined multiple times. Each option/command can be specified at most once.
```

bhalevy:
Closing now that it's the default in cassandra-stress

PR #586, opened 2024-07-09
node.py: Fix Node.clear(only_data=True) deleting data files

paszkow:
`Node.clear(only_data=True)` shall clear user defined keyspaces only leaving all system ones untouched. The patch skips all system* keyspace directories when clearing user data.

Keyspace directories contain table directories not files that shall be deleted. When clearing keyspace directories, all underlying table directories are removed.

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -1006,11 +1006,11 @@ class Node(object):
             if only_data:
                 for dir in os.listdir(full_dir):
                     keyspace_dir = os.path.join(full_dir, dir)
-                    if os.path.isdir(keyspace_dir) and dir != "system":
-                        for f in os.listdir(keyspace_dir):
-                            full_path = os.path.join(keyspace_dir, f)
-                            if os.path.isfile(full_path):
-                                os.remove(full_path)
+                    if os.path.isdir(keyspace_dir) and not dir.startswith("system"):
+                        for table in os.listdir(keyspace_dir):
+                            table_dir = os.path.join(keyspace_dir, table)
+                            shutil.rmtree(table_dir)
+                            os.mkdir(table_dir)
             else:
                 for root, dirs, files in os.walk(full_dir):
                     for f in files:


ccmlib/node.py
```diff
@@ -1006,11 +1006,13 @@ def clear(self, clear_all=False, only_data=False, saved_caches=False):
...
                 for dir in os.listdir(full_dir):
                     keyspace_dir = os.path.join(full_dir, dir)
-                    if os.path.isdir(keyspace_dir) and dir != "system":
+                    if os.path.isdir(keyspace_dir) and not dir.startswith("system"):
                         for f in os.listdir(keyspace_dir):
                             full_path = os.path.join(keyspace_dir, f)
                             if os.path.isfile(full_path):
                                 os.remove(full_path)
+                            elif d == "data":
+                                shutil.rmtree(full_path)
```
tchaikov, line 1015: why not just put something like
```py
for f in os.listdir(keyspace_dir):
    table_dir = os.path.join(keyspace_dir, f)
    shutil.rmtree(table_dir)
    os.mkdir(table_dir)
```

paszkow: You are right. For some reason I though we enter this path for commitlog and other mentioned directories but there is `if only_data:` So obviously we do not.

ccmlib/node.py
```diff
@@ -1006,11 +1006,13 @@ def clear(self, clear_all=False, only_data=False, saved_caches=False):
             if only_data:
                 for dir in os.listdir(full_dir):
                     keyspace_dir = os.path.join(full_dir, dir)
-                    if os.path.isdir(keyspace_dir) and dir != "system":
+                    if os.path.isdir(keyspace_dir) and not dir.startswith("system"):
                         for f in os.listdir(keyspace_dir):
```
fruch, line 1010: I think this is the core of the fix, but the description doesn't really explain it

Also reference to that dtest is broken cause of this would help with understand the reason for the chage

ccmlib/node.py
fruch: Maybe using pathlib.Path could be a bit more fit and no need for 3 different modules to do this

ccmlib/node.py
paszkow: Actually, for me the other change was more important as it did not remove sstable files. My tests were passing but repair was not triggered.

tchaikov (approved):
lgtm

ccmlib/node.py
```diff
@@ -1006,11 +1006,11 @@ def clear(self, clear_all=False, only_data=False, saved_caches=False):
             if only_data:
                 for dir in os.listdir(full_dir):
                     keyspace_dir = os.path.join(full_dir, dir)
-                    if os.path.isdir(keyspace_dir) and dir != "system":
-                        for f in os.listdir(keyspace_dir):
-                            full_path = os.path.join(keyspace_dir, f)
-                            if os.path.isfile(full_path):
-                                os.remove(full_path)
+                    if os.path.isdir(keyspace_dir) and not dir.startswith("system"):
```
tchaikov, line 1009: does this change apply to https://github.com/pcmanus/ccm as well, probably we can upstream this change?

fruch: I'm not exactly sure where the  upstream is

Since they are using personal forks

I don't think there is a real one for that

fruch (approved):

ccmlib/node.py
tchaikov: I think we forked from this project. And I am sure this is the upstream.

fruch: It's moved

And that the name I've seen mentioned in the upstream dtest repo
https://github.com/riptano/ccm

We could suggest it, but I don't know if they would care about it or not.

I still don't think it's needed, we aren't gonna sync from it, maybe only cherry pick specific things.

tchaikov: No it’s not needed. The reason why suggest this was not because it’s needed, but because it’s a community, we benefit from others‘ work, would be ideal if we can contribute back.

fruch: I would suggest opening an issue. and see if anyone is interested in this getting fixed.

My past experiences in other similar cases, it was a bit tedious, and took a long time to get things in.

I don't say we shouldn't do those, but just be mindful of people time (and not block thier PRs for those)

PR #587, opened 2024-07-09
ccmlib/scylla_repository: handle name change of the unified package

fruch:
seem like at some point (spotted on release 2024.2) the arch value in the unified package was move to the end of the filename which cause ccm to select the wrong arch file

ccm had a fallback to support older version (4.3 and below) unified package that doesn't mention arch at all (before we had ARM support)

this change is catering to both options, but makeing sure all of the possible options are considered, correctly.


ccmlib/scylla_repository.py
```diff
@@ -71,15 +71,21 @@ def release_packages(s3_url, version, arch='x86_64', scylla_product='scylla'):
...
     # 'scylla-x86_64-unified-package-5.1.2-0.20221225.4c0f7ea09893-5.1.2.0.20221225.4c0f7ea09893.tar.gz'
     # 'scylla-x86_64-unified-package-5.1.3-0.20230112.addc4666d502-5.1.3.0.20230112.addc4666d502.tar.gz'
+    # older versions unified packages might look like
+    # 'scylla-unified-package-4.3.0~rc0-0.20220810.86a6c1fb2b79.tar.gz'
+    # and newer versions (6.0/2024.2 and up) might look like:
+    # 'scylla-unified-package-6.0.0~rc0-0.20231219.c7cdb16538f2.x86_64.tar.gz'
     all_unified_packages = [name for name in files_in_bucket if f'{scylla_product}-{arch}-unified-package' in name or
-                            (f'{scylla_product}-unified' in name and arch in name) or
-                            f'{scylla_product}-unified' in name]
+                            (f'{scylla_product}-unified' in name and arch in name ) or
```
ilya-rarov, line 86: AFAICT, the only change in this line is the whitespace here `name )`.
Seems this is an unintended change that could be reverted.

fruch: fixed

ilya-rarov (approved):
LGTM

ccmlib/scylla_repository.py
```diff
@@ -71,15 +71,21 @@ def release_packages(s3_url, version, arch='x86_64', scylla_product='scylla'):
...
     # 'scylla-x86_64-unified-package-5.1.2-0.20221225.4c0f7ea09893-5.1.2.0.20221225.4c0f7ea09893.tar.gz'
     # 'scylla-x86_64-unified-package-5.1.3-0.20230112.addc4666d502-5.1.3.0.20230112.addc4666d502.tar.gz'
+    # older versions unified packages might look like
+    # 'scylla-unified-package-4.3.0~rc0-0.20220810.86a6c1fb2b79.tar.gz'
+    # and newer versions (6.0/2024.2 and up) might look like:
+    # 'scylla-unified-package-6.0.0~rc0-0.20231219.c7cdb16538f2.x86_64.tar.gz'
     all_unified_packages = [name for name in files_in_bucket if f'{scylla_product}-{arch}-unified-package' in name or
                             (f'{scylla_product}-unified' in name and arch in name) or
-                            f'{scylla_product}-unified' in name]
+                            (f'{scylla_product}-unified' in name and all([known_arch not in name for known_arch in known_archs]))
```
tchaikov, line 87: i found that i had a hard time matching the comment with the code. so created #588

PR #591, opened 2024-07-30
scylla_repository: setup: add support for debug mode

### LINKED ISSUE #590: scylla_repository setup should support scylla-debug in the version format
Currently the `version` format supported by scylla_repository.setup can be either:
`unstable/<branch>:<build-date>` (build-date can be `latest` as well)
or
`release:<release-number>`

And on top of it we use a couple of environment variables to specify some more attributes:
`SCYLLA_PRODUCT` can be `scylla` or `scylla-enterprise` (default is `scylla`)
`SCYLLA_ARCH` can `x86_64` or `aarch64` (default is `x86_64`)

But, there is no way to specify `scylla-debug` vs. `scylla` for the unified package.

I really don't like the use of the environment variables for the extra attributes.
Maybe we should extend the version syntax to allow passing them with the version string, as well as the `debug` flag.

bhalevy:
This series adds support for specifying debug mode for scylla_version.

```
            flags can be added to the version using [(flag(=value)?)+]
              - supported flags are:
                - product=(scylla|scylla-enterprise)
                - arch=(x86_64|aarch64)
                - debug
```

Fixes #590

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -244,6 +244,7 @@ def setup(version, verbose=True, skip_downloads=False):
     type_n_version = version.split(':', 1)
     scylla_product = os.environ.get('SCYLLA_PRODUCT', 'scylla')
     scylla_arch = os.environ.get('SCYLLA_ARCH', uname().machine)
+    scylla_debug = os.environ.get('SCYLLA_DEBUG', '')
 
     packages = None
 
@@ -252,6 +253,8 @@ def setup(version, verbose=True, skip_downloads=False):
     if type_n_version[0] == 'release':
         version = normalize_scylla_version(version)
         type_n_version = version.split(os.path.sep, 1)
+    if scylla_debug:
+        version = f"{version}-debug"
     version_dir = version_directory(version) if not skip_downloads else None
 
     # If the test version is unstable (not release, maybe private branch) and installation folder exists,
@@ -270,7 +273,10 @@ def setup(version, verbose=True, skip_downloads=False):
 
             s3_url = get_relocatable_s3_url('', major_version, RELEASE_RELOCATABLE_URLS_BASE)
 
-            packages, type_n_version[1] = release_packages(s3_url=s3_url, version=s3_version, arch=scylla_arch, scylla_product=scylla_product)
+            scylla_product_mode = scylla_product
+            if scylla_debug:
+                scylla_product_mode = f"{scylla_product_mode}-debug"
+            packages, type_n_version[1] = release_packages(s3_url=s3_url, version=s3_version, arch=scylla_arch, scylla_product=scylla_product_mode)
         else:
             _, branch = type_n_version[0].split("/")
             s3_url = get_relocatable_s3_url(branch, s3_version, RELOCATABLE_URLS_BASE)
@@ -278,7 +284,9 @@ def setup(version, verbose=True, skip_downloads=False):
             try:
                 build_manifest = read_build_manifest(s3_url)
                 url = build_manifest.get(f'unified-pack-url-{scylla_arch}')
-                assert url, "didn't found the url for unified package"
+                assert url, f"didn't found the url for unified package: build_manifest={build_manifest}"
+                if scylla_debug:
+                    url = re.sub("-unified", "-debug-unified", url)
                 if not url.startswith('s3'):
                     url = f's3.amazonaws.com/{url}'
                 url = f'http://{url}'
@@ -316,6 +324,8 @@ def setup(version, verbose=True, skip_downloads=False):
                 )
 
         version = os.path.join(*type_n_version)
+        if scylla_debug:
+            version = f"{version}-debug"
 
     if skip_downloads:
         return directory_name(version), packages
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -284,6 +284,8 @@ def setup(version, verbose=True, skip_downloads=False):
             try:
                 build_manifest = read_build_manifest(s3_url)
                 url = build_manifest.get(f'unified-pack-url-{scylla_arch}')
+                if not url:
+                    url = build_manifest.get("unified-pack-url")
                 assert url, f"didn't found the url for unified package: build_manifest={build_manifest}"
                 if scylla_debug:
                     url = re.sub("-unified", "-debug-unified", url)
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -235,16 +235,40 @@ def setup(version, verbose=True, skip_downloads=False):
               - release:4.2.1
               - release:2020.1
               - release:2020.1.5
+            flags can be added to the version using [(flag(=value)?)+]
+              - supported flags are:
+                - product=(scylla|scylla-enterprise)
+                - arch=(x86_64|aarch64)
+                - debug
 
     :param verbose: if True, print progress during download
     :param skip_downloads: if True, skips the actual download of files for testing purposes
 
     """
     s3_url = ''
-    type_n_version = version.split(':', 1)
-    scylla_product = os.environ.get('SCYLLA_PRODUCT', 'scylla')
-    scylla_arch = os.environ.get('SCYLLA_ARCH', uname().machine)
-    scylla_debug = os.environ.get('SCYLLA_DEBUG', '')
+    version_re = re.compile(r"""
+                            (?P<type>
+                                \w+(?:/(?:[\w\d.~-]+))?
+                            )
+                            :(?P<version>[\w\d:.~-]+)
+                            (?:
+                                \[
+                                    (?P<flags>[^\]]*)
+                                \]
+                            )?""", re.VERBOSE)
+    m = version_re.match(version)
+    assert m, f"Invalid version {version}: use (release:version|unstable/branch:(latest|timestamp))([(flag(=value)?)*])? where flags can be any of: (product|arch|debug)"
+    type_n_version = [m.group("type"), m.group("version")]
+    flags = dict()
+    if m.group("flags"):
+        for kv in m.group("flags").split(","):
+            t = kv.split("=", 1)
+            key = t[0].strip().lower()
+            value = t[1].strip() if len(t) > 1 else True
+            flags[key] = value
+    scylla_product = flags.get("product", os.environ.get('SCYLLA_PRODUCT', 'scylla'))
+    scylla_arch = flags.get("arch", os.environ.get('SCYLLA_ARCH', uname().machine))
+    scylla_debug = flags.get("debug", os.environ.get('SCYLLA_DEBUG', False))
 
     packages = None
 
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -8,6 +8,7 @@ import sys
 import glob
 import urllib
 import logging
+from platform import uname
 import random
 import time
 from pathlib import Path
@@ -242,7 +243,7 @@ def setup(version, verbose=True, skip_downloads=False):
     s3_url = ''
     type_n_version = version.split(':', 1)
     scylla_product = os.environ.get('SCYLLA_PRODUCT', 'scylla')
-    scylla_arch = os.environ.get('SCYLLA_ARCH', 'x86_64')
+    scylla_arch = os.environ.get('SCYLLA_ARCH', uname().machine)
 
     packages = None
 
--- a/tests/test_scylla_repository.py
+++ b/tests/test_scylla_repository.py
@@ -2,6 +2,7 @@ import time
 import typing
 from pathlib import Path
 import random
+import re
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


ccmlib/scylla_repository.py
```diff
@@ -234,15 +234,47 @@ def setup(version, verbose=True, skip_downloads=False):
...
+    assert m, f"Invalid version {version}: use (release:version|unstable/branch:(latest|timestamp))([(flag(=value)?)*])? where flags can be any of: (product|arch|debug)"
+    type_n_version = [m.group("type"), m.group("version")]
+    flags = dict()
+    if m.group("flags"):
+        for kv in m.group("flags").split(","):
+            t = kv.split("=", 1)
+            key = t[0].strip().lower()
+            if len(t) > 1:
+                value = t[1].strip()
+                if value.lower() in ["true", "yes", "1"]:
```
fruch, line 268: I don't understand this part, non of those flags has boolean content, why do we need this ?

ccmlib/scylla_repository.py
```diff
@@ -234,15 +234,47 @@ def setup(version, verbose=True, skip_downloads=False):
...
 
     :param verbose: if True, print progress during download
     :param skip_downloads: if True, skips the actual download of files for testing purposes
 
     """
     s3_url = ''
-    type_n_version = version.split(':', 1)
-    scylla_product = os.environ.get('SCYLLA_PRODUCT', 'scylla')
-    scylla_arch = os.environ.get('SCYLLA_ARCH', 'x86_64')
+    version_re = re.compile(r"""
```
fruch, line 248: for such elaborate regex, it would be nice to have a unittest demo the different options are work as expected

bhalevy: okay, I tested it manually, but can add a unit test

ccmlib/scylla_repository.py
bhalevy: It's a little semantic sugar if someone wants to specify `[debug=true]` instead of `[debug]`
since the syntax allows `[debug=value]`

bhalevy: it's not a must-have by any means

ccmlib/scylla_repository.py
bhalevy: updated tests (and actually had to fix the regex to accommodate for the exiting versions)

ccmlib/scylla_repository.py
bhalevy: removed in v2

bhalevy:
tests/test_scylla_repository.py::TestReinstallPackages::test_setup_no_unified_packages_reinstall fails for me locally also with current master scylladb/scylla-ccm@9ce5d753d213e7fc682bf0d121b8c59b597199a0

fruch:
> tests/test_scylla_repository.py::TestReinstallPackages::test_setup_no_unified_packages_reinstall fails for me locally also with current master [9ce5d75](https://github.com/scylladb/scylla-ccm/commit/9ce5d753d213e7fc682bf0d121b8c59b597199a0)

for now skipping the test:
https://github.com/scylladb/scylla-ccm/pull/592

we'll need to clear it with the rest of the code for supporting older release, that doesn't have unified package (scylla 4.2 and below)

ccmlib/scylla_repository.py
```diff
@@ -234,15 +234,40 @@ def setup(version, verbose=True, skip_downloads=False):
...
+    type_n_version = [m.group("type"), m.group("version")]
+    flags = dict()
+    if m.group("flags"):
+        for kv in m.group("flags").split(","):
+            t = kv.split("=", 1)
+            key = t[0].strip().lower()
+            value = t[1].strip() if len(t) > 1 else True
+            flags[key] = value
+    scylla_product = flags.get("product", os.environ.get('SCYLLA_PRODUCT', 'scylla'))
+    scylla_arch = flags.get("arch", os.environ.get('SCYLLA_ARCH', 'x86_64'))
```
cezarmoise, line 269:
```suggestion
    scylla_arch = flags.get("arch", os.environ.get('SCYLLA_ARCH', platform.uname().machine))
```
A better default that includes those that might have arm based systems.

bhalevy: Makes sense.  It's somewhat out of scope for this PR, but  an be incorporated into the series whilr making changes in this area.

fruch: In dtest we do that

So we can pick the right docker image:

https://github.com/scylladb/scylla-dtest/blob/f58d7b89534ad895ee7d1ec54c404414eedcceca/scripts/run_test.sh#L108

bhalevy: > In dtest we do that
> 
> So we can pick the right docker image:
> 
> [scylladb/scylla-dtest@`f58d7b8`/scripts/run_test.sh#L108](https://github.com/scylladb/scylla-dtest/blob/f58d7b89534ad895ee7d1ec54c404414eedcceca/scripts/run_test.sh#L108)

meaning?
This seems to support @cezarmoise's suggestion, right?

bhalevy: Added in v3

bhalevy:
In v3 ([267249b](https://github.com/scylladb/scylla-ccm/commit/267249be408fc66e55f121e618dfe7404afc32a3)):
- rebased
- added initialization to SCYLLA_ARCH as per https://github.com/scylladb/scylla-ccm/pull/591#discussion_r1697230144

ccmlib/scylla_repository.py
```diff
@@ -234,15 +235,40 @@ def setup(version, verbose=True, skip_downloads=False):
               - release:4.2.1
               - release:2020.1
               - release:2020.1.5
+            flags can be added to the version using [(flag(=value)?)+]
+              - supported flags are:
+                - product=(scylla|scylla-enterprise)
```
cezarmoise, line 240: Don't we already have enough information to determine scylla vs scylla-enterprise without having to specify it?

ccmlib/scylla_repository.py
```diff
@@ -234,15 +235,40 @@ def setup(version, verbose=True, skip_downloads=False):
...
+                - arch=(x86_64|aarch64)
```
cezarmoise, line 241: Is there an use case where one would want a different arch than that of their system's?

bhalevy: I think it could be good to support installing scylla from a foreign architecture package and launch it remotely on an appropriate machine.  I don't think it's done in practice now though.

ccmlib/scylla_repository.py
bhalevy: it's obvious for `unstable/enterprise`, but for `unstable/branch-2023.1` we'll have to second-guess the branch naming conventions. That's possible of course and we already do it elsewhere, not that I like it very much...

bhalevy: Some cases, like for `byo` I think we wouldn't be able to infer the product from the scylla_version.

bhalevy: Speaking of that, how can we represent a version for e.g. `https://downloads.scylladb.com/testing/unstable/scylla-enterprise/enterprise-2024.1/relocatable/byo/2024-07-23T20%3A47%3A40Z/`
Maybe we should support a url (`s3://` or `https://`) in addition to the current shorthand forms.

cezarmoise: Something like `testing/enterprise-2024.01:byo/2024-07-23T20:47:40Z`
Even though what is before the `/` is ignored:
https://github.com/scylladb/scylla-ccm/blob/9ce5d753d213e7fc682bf0d121b8c59b597199a0/ccmlib/scylla_repository.py#L274

We would just need to add another URL to this list:
https://github.com/scylladb/scylla-ccm/blob/9ce5d753d213e7fc682bf0d121b8c59b597199a0/ccmlib/scylla_repository.py#L32-L35

Something like `https://s3.amazonaws.com/downloads.scylladb.com/testing/unstable/scylla-enterprise/{0}/relocatable/{1}/}`

Of course, there are other cases in the testing subfolder that would need to be covered as well.

ccmlib/scylla_repository.py
fruch: In theory on can download the arm files with it in you machine while running the tests under docker with emulation of different arch then your machine.

In practice we call install.sh, and it can't work on the wrong arch.

So I don't think there is such a case for now

ccmlib/scylla_repository.py
fruch: We do support download from a full path of the unified package, it's just not via scylla_version parameter.

That would be enough, we don't need so many complications to this one.

As for the product, we do estimate it good enough base on release or branch name. I haven't used it for quite some time now, I think it can be removed, and not introduced as a new option.

bhalevy: Let's postpone further cleanups till after we merge this PR
I think it has enough value on its own already.

fruch (approved):

fruch:
@bhalevy 

had to yank this one out.

it was break CI

```
20:26:30  INTERNALERROR> Traceback (most recent call last):
20:26:30  INTERNALERROR>   File "/usr/local/lib/python3.12/site-packages/_pytest/main.py", line 268, in wrap_session
20:26:30  INTERNALERROR>     config.hook.pytest_sessionstart(session=session)
20:26:30  INTERNALERROR>   File "/usr/local/lib/python3.12/site-packages/pluggy/_hooks.py", line 513, in __call__
20:26:30  INTERNALERROR>     return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
20:26:30  INTERNALERROR>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
20:26:30  INTERNALERROR>   File "/usr/local/lib/python3.12/site-packages/pluggy/_manager.py", line 120, in _hookexec
20:26:30  INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
20:26:30  INTERNALERROR>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
20:26:30  INTERNALERROR>   File "/usr/local/lib/python3.12/site-packages/pluggy/_callers.py", line 182, in _multicall
20:26:30  INTERNALERROR>     return outcome.get_result()
20:26:30  INTERNALERROR>            ^^^^^^^^^^^^^^^^^^^^
20:26:30  INTERNALERROR>   File "/usr/local/lib/python3.12/site-packages/pluggy/_result.py", line 100, in get_result
20:26:30  INTERNALERROR>     raise exc.with_traceback(exc.__traceback__)
20:26:30  INTERNALERROR>   File "/usr/local/lib/python3.12/site-packages/pluggy/_callers.py", line 103, in _multicall
20:26:30  INTERNALERROR>     res = hook_impl.function(*args)
20:26:30  INTERNALERROR>           ^^^^^^^^^^^^^^^^^^^^^^^^^
20:26:30  INTERNALERROR>   File "/jenkins/workspace/scylla-staging/dtest-pytest-gating/scylla-dtest/conftest.py", line 370, in pytest_sessionstart
20:26:30  INTERNALERROR>     _scylla_mode = scylla_mode(cassandra_dir, scylla_version)
20:26:30  INTERNALERROR>                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
20:26:30  INTERNALERROR>   File "/jenkins/workspace/scylla-staging/dtest-pytest-gating/scylla-dtest/tools/marks.py", line 35, in scylla_mode
20:26:30  INTERNALERROR>     return dtest_config.get_scylla_mode()
20:26:30  INTERNALERROR>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
20:26:30  INTERNALERROR>   File "/jenkins/workspace/scylla-staging/dtest-pytest-gating/scylla-dtest/dtest_config.py", line 90, in get_scylla_mode
20:26:30  INTERNALERROR>     ccm_repo_cache_dir, _ = ccmlib.scylla_repository.setup(self.scylla_version)
20:26:30  INTERNALERROR>                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
20:26:30  INTERNALERROR>   File "/home/jenkins/.local/lib/python3.12/site-packages/ccmlib/scylla_repository.py", line 260, in setup
20:26:30  INTERNALERROR>     assert m, f"Invalid version {version}: use (release:version|unstable/branch:(latest|timestamp))([(flag(=value)?)*])? where flags can be any of: (product|arch|debug)"
20:26:30  INTERNALERROR>            ^
20:26:30  INTERNALERROR> AssertionError: Invalid version 9214aecbe71c1e58c418f143b8a5c4b4554c4a5a: use (release:version|unstable/branch:(latest|timestamp))([(flag(=value)?)*])? where flags can be any of: (product|arch|debug)
20:26:30  Docker exitcode: 3
```

`scylla_version` can be any string once you specify the unified package specifically, like dtest CI is doing (and is the manual way as well for using specific unified package url)

PR #593, opened 2024-08-01
ccmlib/scylla_node: stop hardcoding `api_port`

fruch:
servel place in the code were assuming that `api_port` would be 10000.

some tests in the java-driver we setting `api_port` to a random port hence breaking some of the ccm functionality, like nodetool and wait_for_other_notice logic

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -115,6 +115,13 @@ class ScyllaNode(Node):
 
         return self._has_jmx
 
+    @property
+    def scylla_yaml(self) -> Dict[str, Any]:
+        return yaml.load(Path(self.get_conf_dir()) / common.SCYLLA_CONF)
+    @property
+    def api_port(self) -> int:
+        return self.scylla_yaml.get('api_port', 10000)
+
     def scylla_mode(self):
         return self.cluster.get_scylla_mode()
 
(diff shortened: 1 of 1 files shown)


ccmlib/scylla_node.py
```diff
@@ -115,6 +115,13 @@ def has_jmx(self):
 
         return self._has_jmx
 
+    @property
+    def scylla_yaml(self) -> Dict[str, Any]:
+        return yaml.load(Path(self.get_conf_dir()) / common.SCYLLA_CONF)
+    @property
```
tchaikov, line 121: could you please add an empty line here?

tchaikov (commented):
lgtm modulo a nit.

ccmlib/scylla_node.py
tchaikov: FWIW, we are encoding `10000` in dtest as well. i will fix them as well in piecemeal.

fruch: > FWIW, we are encoding `10000` in dtest as well. i will fix them as well piecemeal.

And also in SCT, and other places.

I just noticed it in java-driver, that probably does it for no actual reason, historical reason let's call it

bhalevy (approved):

Issue #589, opened 2024-07-17
ScyllaNode: override get_highest_supported_sstable_version using scylla api


bhalevy:
Based on https://github.com/scylladb/scylladb/issues/19772, let's override `Node.get_highest_supported_sstable_version` in the `ScyllaNode` class, and invoke the new system api to `get_highest_supported_sstable_version`.

This would allow upgrade dtest to work with the latest versiosn (that have ME_SSTABLE_FORMAT enabled by default, without a corresponding cluster feature, that the existing implementation of `get_highest_supported_sstable_version` depends on)

bhalevy:
Note that the implementation needs to be backward compatible so it'd work also with old versions that don't support the new api yet.

bhalevy:
https://github.com/scylladb/scylladb/issues/19772 is fixed so the new api is available.
We can start implementing this libccm function that depends on it.

bhalevy:
@Lakshmipathi what's the verdict for this PR, can it be closed?

Cc @pehala

Lakshmipathi:
@bhalevy I see your latest comment here https://github.com/scylladb/scylla-dtest/pull/4809  after fixing it, it can be merged.

PR #596, opened 2024-08-14
ccmlib/scylla_node: Get highest supported sstable version using the new api

Lakshmipathi:
Expose newly added api endpoint /system/highest_supported_sstable_version, so that dtest test-scripts can get highest supported sstable version using this api.
More details:  https://github.com/scylladb/scylla-ccm/issues/589

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -76,6 +76,7 @@ class ScyllaNode(Node):
         self.upgraded = False
         self.upgrader = NodeUpgrader(node=self)
         self.node_hostid = None
+        self.highest_supported_sstable_version = None
         self._create_directory()
 
     @property
@@ -1397,6 +1398,18 @@ class ScyllaNode(Node):
                 return self.node_hostid
         except Exception as e:
             self.error(f"Failed to get hostid using {url}: {e}")
+        return None
+
+    def get_highest_supported_sstable_version(self, timeout=60):
+        try:
+            node_address = self.address()
+            url = f"http://{node_address}:{self.api_port}/system/highest_supported_sstable_version"
+            response = requests.get(url=url, timeout=timeout)
+            if response.status_code == requests.codes.ok:
+                self.highest_supported_sstable_version = response.json()
+                return self.highest_supported_sstable_version
+        except Exception as e:
+            self.warning(f"Failed to get highest_supported_sstable_version using {url}: {e}")
         return None
 
     def watch_rest_for_alive(self, nodes, timeout=120, wait_normal_token_owner=True):


mykaul:
nit, and not particularly related to this PR: I'm not sure how we do not have a wrapper to the REST API calls we make - there should be timeout, retries, handling of the different error codes and default return values, and we should just have a function asking for and endpoint. 
In this specific implementation - we do have timeout (which is very long - 60 secs), we do not have retries, we do not have logging or handling of anything other than  requests.codes.ok 

Again, not for this specific PR, but general observation.

Lakshmipathi:
Yes, it makes sense to have such a wrapper. may be should we create a new GH issue?

mykaul:
> Yes, it makes sense to have such a wrapper. may be should we create a new GH issue?

Yes, of course. It's unrelated to this PR, I've just noted it here as I saw the code.

fruch:
> nit, and not particularly related to this PR: I'm not sure how we do not have a wrapper to the REST API calls we make - there should be timeout, retries, handling of the different error codes and default return values, and we should just have a function asking for and endpoint. In this specific implementation - we do have timeout (which is very long - 60 secs), we do not have retries, we do not have logging or handling of anything other than requests.codes.ok
> 
> Again, not for this specific PR, but general observation.

most of the api calls, are not in ccm, but in dtest.
anyhow ccm usage of api are for specific usage, and in mostly places does have retry logic

in the context of this change, I don't understand why it's needed to be in ccm, and not in dtest ?

ccmlib/scylla_node.py
```diff
@@ -1399,6 +1400,18 @@ def hostid(self, timeout=60, force_refresh=False):
             self.error(f"Failed to get hostid using {url}: {e}")
         return None
 
+    def get_highest_supported_sstable_version(self, timeout=60):
```
fruch, line 1403: this is not the function @bhalevy was asking to override, i.e. `get_node_supported_sstable_versions` is what exist in `node.py`

I think we should do this implementation in dtest, this is used for upgrades, and only called from dtest.
this logic and api calls are not related to any of the ccm responsibilities

mykaul: Maybe added to https://github.com/scylladb/scylla-dtest/blob/next/tools/rest_clients.py ?

fruch: yeah that one of the place we have usage of scylla APIs.
and we can extend is with default timeout and such, as needed.

even that I think that each call on it's own, would need to decide if retries are needed, or what's the default timeout.
but yes it's a good think to have them consolidated in one place (I just don't think ccm is that place, I it's beyond it's scope, like it doesn't have any function to do cql connection with a driver)

Lakshmipathi: > this is not the function @bhalevy was asking to override, i.e. `get_node_supported_sstable_versions` is what exist in `node.py`
> 
> I think we should do this implementation in dtest, this is used for upgrades, and only called from dtest. this logic and api calls are not related to any of the ccm responsibilities

ok. Should I move the existing `get_node_supported_sstable_versions()`  from ccmlib/node.py to dtest's tools/rest.clients.py ? Its used by only by scylla-dtest/rolling_upgrade_test.py

fruch: I think it would be better

Not all utility functions fits in ccm

Lakshmipathi: Opened new PR on dtest: https://github.com/scylladb/scylla-dtest/pull/4809 will close this PR.

Lakshmipathi:
Moved to https://github.com/scylladb/scylla-dtest/pull/4809

PR #597, opened 2024-08-14
treewide: make sure we use ruamel.yaml in a thread safe manner

fruch:
we have multiple tests failing with the following:
```
>       self.stream.write(data)
E       ValueError: I/O operation on closed file.

../scylla/.local/lib/python3.12/site-packages/ruamel/yaml/emitter.py:1249: ValueError
```

seems like it start when we have calls from multiple thread that might be
updating the same files, or sharing the same state.

this change make sure we create `YAML` instances as needed,
so they won't be shared between threads

Ref: https://sourceforge.net/p/ruamel-yaml/tickets/367/

#### Testing
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/358/ (Raft - OSS)
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/359/ (Gossip - OSS)
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/361/ (Raft - OSS)
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/362/ (Gossip - OSS)

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -27,8 +27,6 @@ from boto3.session import Session
 from botocore import UNSIGNED
 from botocore.client import Config
 
-yaml = YAML()
-yaml.default_flow_style = False
 
 BIN_DIR = "bin"
 CASSANDRA_CONF_DIR = "conf"
@@ -817,6 +815,7 @@ def normalize_interface(itf):
 
 def parse_settings(args):
     settings = {}
+    yaml = YAML()
     for s in args:
         splitted = s.split(':', 1)
         if len(splitted) != 2:
@@ -888,7 +887,7 @@ def get_version_from_build(install_dir=None, node_path=None):
 def get_default_scylla_yaml(install_dir):
     scylla_yaml_path = Path(install_dir) / SCYLLA_CONF_DIR / SCYLLA_CONF
     with scylla_yaml_path.open() as f:
-        return yaml.load(f)
+        return YAML().load(f)
 
 def _get_scylla_version(install_dir):
     scylla_version_files = [
@@ -970,7 +969,7 @@ def is_dse_cluster(path):
             cluster_path = os.path.join(path, name)
             filename = os.path.join(cluster_path, 'cluster.conf')
             with open(filename, 'r') as f:
-                data = yaml.load(f)
+                data = YAML().load(f)
             if 'dse_dir' in data:
                 return True
     except IOError:
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -24,8 +24,6 @@ from ccmlib.cli_session import CliSession
 from ccmlib.repository import setup
 from ccmlib.utils.version import parse_version
 
-yaml = YAML()
-yaml.default_flow_style = False
 
 class Status():
     UNINITIALIZED = "UNINITIALIZED"
@@ -138,7 +136,7 @@ class Node(object):
         node_path = os.path.join(path, name)
         filename = os.path.join(node_path, 'node.conf')
         with open(filename, 'r') as f:
-            data = yaml.load(f)
+            data = YAML().load(f)
         try:
             itf = data['interfaces']
             initial_token = None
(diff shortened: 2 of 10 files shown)


fruch:
seeing 
https://sourceforge.net/p/ruamel-yaml/tickets/367/

maybe this fix isn't enough, and we should be sharing the `YAML` object at all, espacily not across threads

mykaul:
> seeing https://sourceforge.net/p/ruamel-yaml/tickets/367/
> 
> maybe this fix isn't enough, and we should be sharing the `YAML` object at all, espacily not across threads

should or shouldn't?

fruch:
v2:

- switch to stop sharing of `YAML` instances, since they aren't thread safe

ccmlib/common.py
```diff
@@ -1036,3 +1035,8 @@ def get_installed_scylla_package_hash(source_file: Path):
         if line.startswith("hash="):
             current_hash = line.replace("hash=", "").strip()
     return current_hash
+
+def get_yaml_instance():
+    yaml = YAML()
+    yaml.default_flow_style = False
```
tchaikov, line 1041: nit, it's not "getting" an instance. it creates one. so i'd suggest just use
```py
yaml = YAML()
```
because `default_flow_style` is `False` if `typ` is `"rt"` by default. and in that case, 
```py
self.default_flow_style = False
```

see https://github.com/commx/ruamel-yaml/blob/master/main.py#L66 , https://github.com/commx/ruamel-yaml/blob/27f71ab3c9bbf604de36b1ab6e170065f37cc2cb/main.py#L100 and https://github.com/commx/ruamel-yaml/blob/27f71ab3c9bbf604de36b1ab6e170065f37cc2cb/main.py#L152

fruch: changed it.

and I'll give it a run

ccmlib/common.py
```diff
@@ -1036,3 +1035,6 @@ def get_installed_scylla_package_hash(source_file: Path):
         if line.startswith("hash="):
             current_hash = line.replace("hash=", "").strip()
     return current_hash
+
+def get_yaml_instance():
+    return YAML()
```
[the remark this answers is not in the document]
fruch: removed it

ccmlib/scylla_node.py
```diff
@@ -19,10 +19,10 @@
 import logging
 
 import psutil
-from ruamel.yaml import YAML
```
tchaikov, line 22: ditto.

ccmlib/node.py
```diff
@@ -17,15 +17,11 @@
 import locale
 from collections import namedtuple
 
-from ruamel.yaml import YAML
```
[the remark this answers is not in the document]
fruch: the price of refactoring it, yet again...

tchaikov (approved):
lgtm. thanks!

fruch:
> > we have multiple tests failing with the following:
> > 
> > ```
> > >       self.stream.write(data)
> > E       ValueError: I/O operation on closed file.
> > 
> > ../scylla/.local/lib/python3.12/site-packages/ruamel/yaml/emitter.py:1249: ValueError
> > ```
> > 
> > seems like it start when we have calls from multiple thread that might be updating the same files, or sharing the same state.
> > 
> > this change make sure we create `YAML` instances as needed, so they won't be shared between threads
> > 
> > Ref: https://sourceforge.net/p/ruamel-yaml/tickets/367/
> > #### Testing
> > 
> >     * [x]  🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/358/ (Raft - OSS)
> > 
> >     * [x]  🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/359/ (Gossip - OSS)
> > 
> >     * [ ]  🕐 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/361/ (Raft - OSS)
> > 
> >     * [ ]  🕐 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/362/ (Gossip - OSS)
> 
> the link of https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/new-dtest-pytest-parallel/362/ led me to a 404 page
> 

That job is configured by default to not let multiple jobs run in parallel

So the 2nd is queued, and the link would be valid when it would start.

I have a script that triggers those, and it can predict the future 😉

tchaikov:
wow, that's your crystal ball!

PR #598, opened 2024-08-18
scylla_repository: setup: allow debug in version string

cezarmoise:
Add support for optional [:debug] flag following the version string to specify debug mode.

Closes: #590
Follow-up on: #591, simple version of that. Just check for the debug flag, and if it exists, add "-debug" where appropiate.

Tested locally
```
➜ pytest -s --scylla-version='unstable/master:latest:debug' --tablets  materialized_views_test.py::TestMaterializedViews::test_do_not_finish_view_building_with_hints
S3 download: scylla-debug-unified-6.2.0~dev-0.20240816.afee3924b3dc.x86_64.tar.gz100%|██████████| 245M/245M [00:11<00:00, 21.7MB/s]
Extracting /tmp/ccm-nfwvr5ro.tar.gz (http://s3.amazonaws.com/downloads.scylladb.com/unstable/scylla/master/relocatable/2024-08-16T01:33:28Z/scylla-debug-unified-6.2.0~dev-0.20240816.afee3924b3dc.x86_64.tar.gz, /tmp/tmprl4krzl2) as version unstable/master/latest-debug ...
Relocatable package format version 3.0 detected.
Completed to install Scylla in the folder '/home/cezar/.ccm/scylla-repository/unstable/master/latest-debug'
```

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -238,12 +238,21 @@ def setup(version, verbose=True, skip_downloads=False):
               - release:4.2.1
               - release:2020.1
               - release:2020.1.5
+            3. Debug versions of the above
+              - unstable/master:latest:debug
+              - unstable/enterprise:2020-08-18T14:49:18Z:debug
+              - release:4.3:debug
+              - release:4.2.1:debug
 
     :param verbose: if True, print progress during download
     :param skip_downloads: if True, skips the actual download of files for testing purposes
 
     """
     s3_url = ''
+    scylla_debug = version.endswith(":debug")
+    if scylla_debug:
+        version = version[:-6]
+    scylla_mode_suffix = '-debug' if scylla_debug else ''
     type_n_version = version.split(':', 1)
     scylla_product = os.environ.get('SCYLLA_PRODUCT', 'scylla')
     scylla_arch = os.environ.get('SCYLLA_ARCH', 'x86_64')
@@ -255,7 +264,7 @@ def setup(version, verbose=True, skip_downloads=False):
     if type_n_version[0] == 'release':
         version = normalize_scylla_version(version)
         type_n_version = version.split(os.path.sep, 1)
-    version_dir = version_directory(version) if not skip_downloads else None
+    version_dir = version_directory(version + scylla_mode_suffix) if not skip_downloads else None
 
     # If the test version is unstable (not release, maybe private branch) and installation folder exists,
     # we want to check if this version was downloaded nd installed already in the past and was changed.
@@ -273,21 +282,23 @@ def setup(version, verbose=True, skip_downloads=False):
 
             s3_url = get_relocatable_s3_url('', major_version, RELEASE_RELOCATABLE_URLS_BASE)
 
-            packages, type_n_version[1] = release_packages(s3_url=s3_url, version=s3_version, arch=scylla_arch, scylla_product=scylla_product)
+            packages, type_n_version[1] = release_packages(s3_url=s3_url, version=s3_version, arch=scylla_arch, scylla_product=scylla_product + scylla_mode_suffix)
         else:
             _, branch = type_n_version[0].split("/")
             s3_url = get_relocatable_s3_url(branch, s3_version, RELOCATABLE_URLS_BASE)
 
             try:
                 build_manifest = read_build_manifest(s3_url)
-                url = build_manifest.get(f'unified-pack-url-{scylla_arch}')
-                assert url, "didn't found the url for unified package"
+                url = build_manifest.get(f'unified-pack-url-{scylla_arch}') or build_manifest.get('unified-pack-url')
+                assert url, f"didn't find the url for unified package, {build_manifest=}"
+                if scylla_debug:
+                    url = re.sub("-unified", "-debug-unified", url)
                 if not url.startswith('s3'):
                     url = f's3.amazonaws.com/{url}'
                 url = f'http://{url}'
                 packages = RelocatablePackages(scylla_unified_package=url)
             except Exception as ex:
-                logging.exception("could download relocatable")
+                logging.exception(f"could not download relocatable, {ex=}")
 
                 scylla_package_path = f'{scylla_product}-package.tar.gz'
                 scylla_arch_package_path = f'{scylla_product}-{scylla_arch}-package.tar.gz'
@@ -318,7 +329,7 @@ def setup(version, verbose=True, skip_downloads=False):
                     scylla_package=os.path.join(s3_url, scylla_package_path)
                 )
 
-        version = os.path.join(*type_n_version)
+        version = os.path.join(*type_n_version) + scylla_mode_suffix
 
     if skip_downloads:
         return directory_name(version), packages
--- a/tests/test_scylla_repository.py
+++ b/tests/test_scylla_repository.py
@@ -2,6 +2,7 @@ import time
 import typing
 from pathlib import Path
 import random
+import re
 
 import pytest
 
@@ -23,6 +24,7 @@ class TestScyllaRepository:
         ("release:2022.2", '2022.2'),
         ("release:2023.1", '2023.1'),
         ("release:2024.2", '2024.2'),
+        ("release:2024.2:debug", '2024.2'),
         ("release:5.1", '5.1'),
         ("release:5.0", '5.0'),
         ("release:4.6", '4.6'),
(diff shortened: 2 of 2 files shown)


ccmlib/scylla_repository.py
```diff
@@ -255,7 +263,7 @@ def setup(version, verbose=True, skip_downloads=False):
     if type_n_version[0] == 'release':
         version = normalize_scylla_version(version)
         type_n_version = version.split(os.path.sep, 1)
-    version_dir = version_directory(version) if not skip_downloads else None
+    version_dir = version_directory(version + "-debug" * scylla_debug) if not skip_downloads else None
```
cezarmoise, line 266: this uses the fact that bool is just an 1 or 0, and a string times 0 is the empty string

ccmlib/scylla_repository.py
```diff
@@ -238,12 +238,20 @@ def setup(version, verbose=True, skip_downloads=False):
...
+              - release:4.2.1:debug
 
     :param verbose: if True, print progress during download
     :param skip_downloads: if True, skips the actual download of files for testing purposes
 
     """
     s3_url = ''
+    scylla_debug = version.split(':')[-1] == "debug"
+    if scylla_debug:
+        version = version[:-6]
```
cezarmoise, line 254: If the debug flag is found, remove if from the version string to the rest of the logic in the function is not impacted

ccmlib/scylla_repository.py
```diff
@@ -273,21 +281,23 @@ def setup(version, verbose=True, skip_downloads=False):
...
+            packages, type_n_version[1] = release_packages(s3_url=s3_url, version=s3_version, arch=scylla_arch, scylla_product=scylla_product + "-debug" * scylla_debug)
         else:
             _, branch = type_n_version[0].split("/")
             s3_url = get_relocatable_s3_url(branch, s3_version, RELOCATABLE_URLS_BASE)
 
             try:
                 build_manifest = read_build_manifest(s3_url)
-                url = build_manifest.get(f'unified-pack-url-{scylla_arch}')
-                assert url, "didn't found the url for unified package"
+                url = build_manifest.get(f'unified-pack-url-{scylla_arch}') or build_manifest.get('unified-pack-url')
```
cezarmoise, line 291: Reused logic from #591, this add compatibility with older manifests

pehala, line 291: Can something simlar be used for debug as well? I would like that better than replacing it in the final URL

ccmlib/scylla_repository.py
pehala: That is very clever, but the fact that you felt that it deserved a comment means it might be not verbose enough to other users, especially in the future. Would simplifying it so that it is more readable keep it as compact as it is now or would it be much longer?

fruch:
One thing we should take for a spin at least with gating tests (what we missed with the last PR)

ccmlib/scylla_repository.py
bhalevy: I agree.  Something like `f"{version}{'-debug' if scylla_debug else ''}" would even be more readable IMO

ccmlib/scylla_repository.py
```diff
@@ -238,12 +238,20 @@ def setup(version, verbose=True, skip_downloads=False):
...
+              - unstable/enterprise:2020-08-18T14:49:18Z:debug
+              - release:4.3:debug
+              - release:4.2.1:debug
 
     :param verbose: if True, print progress during download
     :param skip_downloads: if True, skips the actual download of files for testing purposes
 
     """
     s3_url = ''
+    scylla_debug = version.split(':')[-1] == "debug"
```
bhalevy, line 252: how about `version.endswith(":debug")`?
It seems clearer to me, especially when it is truncated using `version[:-6]`

ccmlib/scylla_repository.py
bhalevy: cc @fruch :point_up:

ccmlib/scylla_repository.py
```diff
@@ -273,21 +281,23 @@ def setup(version, verbose=True, skip_downloads=False):
...
         else:
             _, branch = type_n_version[0].split("/")
             s3_url = get_relocatable_s3_url(branch, s3_version, RELOCATABLE_URLS_BASE)
 
             try:
                 build_manifest = read_build_manifest(s3_url)
-                url = build_manifest.get(f'unified-pack-url-{scylla_arch}')
-                assert url, "didn't found the url for unified package"
+                url = build_manifest.get(f'unified-pack-url-{scylla_arch}') or build_manifest.get('unified-pack-url')
+                assert url, f"didn't found the url for unified package, {build_manifest=}"
```
bhalevy, line 292: nit: "didn't find"

bhalevy: and a stray `=` after `build_manifest`?

ccmlib/scylla_repository.py
```diff
@@ -273,21 +281,23 @@ def setup(version, verbose=True, skip_downloads=False):
...
+                if scylla_debug:
+                    url = re.sub("-unified", "-debug-unified", url)
                 if not url.startswith('s3'):
                     url = f's3.amazonaws.com/{url}'
                 url = f'http://{url}'
                 packages = RelocatablePackages(scylla_unified_package=url)
             except Exception as ex:
-                logging.exception("could download relocatable")
+                logging.exception(f"could download relocatable, {ex=}")
```
bhalevy, line 300: nit: could not?

ccmlib/scylla_repository.py
cezarmoise: > and a stray `=` after `build_manifest`?

This is a fstring feature

```
>>> x = "test"
>>> f"{x=}"
"x='test'"
```

ccmlib/scylla_repository.py
```diff
@@ -318,7 +328,7 @@ def setup(version, verbose=True, skip_downloads=False):
                     scylla_package=os.path.join(s3_url, scylla_package_path)
                 )
 
-        version = os.path.join(*type_n_version)
+        version = os.path.join(*type_n_version) + "-debug" * scylla_debug
```
bhalevy, line 331: here too, `* scylla_debug` is too sophisticated :)
`f"{'-debug' if scylla_debug else ''}"` is longer, but clearer.
Maybe we can store it in a variable and reuse it multiple times.
In theory, it could even be the value of `scylla_debug` instead of a boolean, but it's a hack...

bhalevy: LGTM otherwise

bhalevy (approved):

cezarmoise:
> One thing we should take for a spin at least with gating tests (what we missed with the last PR)

https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/550/ 
https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/551/ 

looks ok at first glance

bhalevy:
@fruch is on PTO, @tchaikov can you please review too?

tests/test_scylla_repository.py
```diff
@@ -111,14 +127,14 @@ def test_setup_release_enterprise_no_unified_package(self, version, expected_cdi
...
-        assert '2021-01-18T15_48_13Z' in cdir
-        assert packages.scylla_unified_package is None
-        assert packages.scylla_package == 'https://s3.amazonaws.com/downloads.scylladb.com/unstable/scylla/master/relocatable/2021-01-18T15:48:13Z/scylla-package.tar.gz'
-        assert packages.scylla_tools_package == 'https://s3.amazonaws.com/downloads.scylladb.com/unstable/scylla/master/relocatable/2021-01-18T15:48:13Z/scylla-tools-package.tar.gz'
-        assert packages.scylla_jmx_package == 'https://s3.amazonaws.com/downloads.scylladb.com/unstable/scylla/master/relocatable/2021-01-18T15:48:13Z/scylla-jmx-package.tar.gz'
-
+    @pytest.mark.parametrize(argnames=['version', 'expected_cdir', 'scylla_debug'], argvalues=[
+        ("unstable/master:2021-01-18T15:48:13Z", "2021-01-18T15:48:13Z", False),
+        ("unstable/master:2021-01-18T15:48:13Z:debug", "2021-01-18T15:48:13Z", True),
+    ])
```
bhalevy, line 133: Can we test this also with enterprise url please?

cezarmoise: https://github.com/scylladb/scylla-ccm/pull/598/commits/aa386e60413c4ca2b93b03655801b44bb4cd7715

bhalevy:
> > One thing we should take for a spin at least with gating tests (what we missed with the last PR)
> 
> @cezarmoise did you run verify this PR with next_gating? Link to [dtest-byo](https://jenkins.scylladb.com/job/scylla-master/job/byo/job/dtest-byo/) job?

I see https://github.com/scylladb/scylla-ccm/pull/598#issuecomment-2298321286

tests/test_scylla_repository.py
```diff
@@ -32,18 +34,28 @@ class TestScyllaRepository:
...
     def test_setup_release_oss(self, version, expected_version_string):
         cdir, version = scylla_setup(version=version, verbose=True)
         assert expected_version_string in version
 
-    def test_setup_unstable_master_new_url(self):
-        cdir, version = scylla_setup(version="unstable/master:2023-09-05T13:59:07Z", verbose=True)
-        assert version == '5.4.0-dev'
+    @pytest.mark.parametrize(argnames=['version', 'expected_version_string'], argvalues=[
+        ("unstable/master:2023-09-05T13:59:07Z", '5.4.0-dev'),
+        ("unstable/master:2023-09-05T13:59:07Z:debug", '5.4.0-dev'),
```
roydahan, line 46: so how do you know if it actually took a debug version?

pehala: You dont, there is an another test that tests if the actual debug version was taken

cezarmoise:
@fruch @tchaikov ping

PR #601, opened 2024-09-05
ccmlib/scylla_node: use functools.cached_property to define has_jmx

tchaikov:
for better readability

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -3,6 +3,7 @@
 
 from datetime import datetime
 import errno
+import functools
 import json
 import os
 import signal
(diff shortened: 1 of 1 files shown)


fruch (approved):
Nit: the title sounds like you are doing a big refactor to all of those opportunities in that module, while you are handling only the specific jmx property.

tchaikov:
> Nit: the title sounds like you are doing a big refactor to all of those opportunities in that module, while you are handling only the specific jmx property.

thanks for your approval, it turns out `has_jmx()` is the only place i can switch to the new decorator.

tchaikov:
v2:

- update the title of the commit to be more specific.

PR #603, opened 2024-09-07
ccmlib/common.py: make java lookup logic more comprehensive

dkropachev:
Before this commit `make_cassandra_env` would only find java at
`/usr/lib/jvm/*` directory and could have misconstrue
`java-18-openjdk-amd64` as `java-8`.
This commit makes `make_cassandra_env` to look at any directory under `/usr/lib/jvm` and
to read version from java output

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -389,23 +389,42 @@ def make_cassandra_env(install_dir, node_path, update_conf=True, hardcode_java_v
 
     if hardcode_java_version:
         known_jvm_names = {
-            '8': ('8-openjdk', '8-jdk', '1.8.0-openjdk'),
-            '11': ('11-openjdk', '11-jdk', '1.11.0-openjdk'),
+            '8': ['1.8', '8'],
+            '11': ['11']
         }
-        assert hardcode_java_version in known_jvm_names.keys(), \
+        jvm_root_path = "/usr/lib/jvm/"
+        assert hardcode_java_version in known_jvm_names, \
             f"hardcode_java_version={hardcode_java_version} not supported in:\n{known_jvm_names}"
 
-        for java_path in pathlib.Path('/usr/lib/jvm/').iterdir():
-            if (java_path.is_dir
-                    and any(name in java_path.name for name in known_jvm_names[hardcode_java_version])
-                    and list(java_path.rglob('java'))):
-                env['JAVA_HOME'] = str(java_path)
-                break
-        else:
-            raise ArgumentError(f"java-8 wasn't found in /usr/lib/jvm/\n {list(pathlib.Path('/usr/lib/jvm/').iterdir())}")
+        java_home_path = get_java_home_path(pathlib.Path(jvm_root_path), known_jvm_names[hardcode_java_version])
+        if java_home_path is None:
+            raise ArgumentError(f"java-{hardcode_java_version} wasn't found in {jvm_root_path}")
 
+        env['JAVA_HOME'] = str(java_home_path.as_posix())
     return env
 
+def get_java_home_path(parent_path: pathlib.Path, hardcode_java_version: List[str]) -> [pathlib.Path]:
+    for java in pathlib.Path(parent_path).rglob('*/bin/java'):
+        if get_jvm_spec_version(java) in hardcode_java_version:
+            return java.parent.parent
+    return None
+
+java_version_property_regexp = re.compile(r'\s*java\.specification\.version\s*=\s*([0-9.]+)\s*')
+
+def get_jvm_spec_version(java_bin_path: pathlib.Path) -> Optional[str]:
+    if not java_bin_path.is_file() or not os.access(java_bin_path, os.X_OK):
+        return None
+    try:
+        properties = subprocess.check_output([java_bin_path.as_posix(), '-XshowSettings:properties', '-version'], stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError as e:
+        return None
+
+    match = java_version_property_regexp.search(properties.decode('utf-8'))
+    if match is None:
+        return None
+    for version in match.groups():
+        return version
+    return None
 
 def make_dse_env(install_dir, node_path):
     env = os.environ.copy()
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ tests/test_results
 .direnv/
 
 .envrc
+ccmlib/tests/
--- a/ccmlib/test_common.py
+++ b/ccmlib/test_common.py
@@ -1,0 +1,93 @@
+import os
+import subprocess
+import unittest
+
+from ccmlib.common import java_version_property_regexp, get_java_home_path
+
+
+class Test(unittest.TestCase):
+    java_19_output = """Property settings:
+        file.encoding = UTF-8
+        file.separator = /
+        java.class.path = 
+        java.class.version = 63.0
+        java.home = /usr/lib/jvm/java-19-openjdk-amd64
+        java.io.tmpdir = /tmp
+        java.library.path = /usr/lib/x86_64-linux-gnu/jni/
+
+            /usr/java/packages/lib
+            /usr/lib/x86_64-linux-gnu/jni
+            /lib/x86_64-linux-gnu
+            /usr/lib/x86_64-linux-gnu
+            /usr/lib/jni
+            /lib
+            /usr/lib
+        java.runtime.name = OpenJDK Runtime Environment
+        java.runtime.version = 19.0.2+7-Ubuntu-0ubuntu322.04
+        java.specification.name = Java Platform API Specification
+        java.specification.vendor = Oracle Corporation
+        java.specification.version = 19
+        java.vendor = Private Build
+        java.vendor.url = Unknown
+        java.vendor.url.bug = Unknown
+        java.version = 19.0.2
+        java.version.date = 2023-01-17
+        java.vm.compressedOopsMode = Zero based
+        java.vm.info = mixed mode, sharing
+        java.vm.name = OpenJDK 64-Bit Server VM
+        java.vm.specification.name = Java Virtual Machine Specification
+        java.vm.specification.vendor = Oracle Corporation
+        java.vm.specification.version = 19
+        java.vm.vendor = Private Build
+        java.vm.version = 19.0.2+7-Ubuntu-0ubuntu322.04
+        jdk.debug = release
+        line.separator = \n 
+        native.encoding = UTF-8
+        os.arch = amd64
+        os.name = Linux
+        os.version = 6.8.0-40-generic
+        path.separator = :
+        stderr.encoding = UTF-8
+        stdout.encoding = UTF-8
+        sun.arch.data.model = 64
+        sun.boot.library.path = /usr/lib/jvm/java-19-openjdk-amd64/lib
+        sun.cpu.endian = little
+        sun.io.unicode.encoding = UnicodeLittle
+        sun.java.launcher = SUN_STANDARD
+        sun.jnu.encoding = UTF-8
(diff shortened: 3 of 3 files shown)


dkropachev:
@fruch , @tchaikov , can you please take a look at it

ccmlib/common.py
```diff
@@ -407,6 +402,26 @@ def make_cassandra_env(install_dir, node_path, update_conf=True, hardcode_java_v
     return env
 
 
+_version_regexp = re.compile('[0-9.]+')
+
+def is_jvm_dir_of_version_id(java_version, path):
+    if not path.is_dir():
+        return False
+    if 'jdk' not in path.name:
```
tchaikov, line 410: this is also fragile, and difficult to comprehend, so instead of parsing the version number encoded in the path, i'd suggest using something like
```py
def get_jvm_spec_version(java: pathlib.Path) -> Optional[str]:
    properties = subprocess.check_output([java.as_posix(), '-XshowSettings:properties', '-version'], stderr=subprocess.STDOUT)
    for line in properties.splitlines():
        parts = line.strip().decode().split('=', 1)
        if len(parts) < 2:
            continue
        name, value = [part.strip() for part in parts]
        if name == 'java.specification.version':
            return value
    return None


known_jvm_names = {
    '8': ['1.8', '8'],
    '11': ['11']
}
for java_path in pathlib.Path('/usr/lib/jvm/').iterdir():
    if not java_path.is_dir():
        continue
    java = java_path / 'bin' / 'java'
    if not java.exists():
        continue
    jvm_version = get_jvm_spec_version(java)
    if jvm_version in known_jvm_names[hardcode_java_version]:
        env['JAVA_HOME'] = str(java_path)
        break
```

also, instead of using "fix" in the title of the commit message, would be better to be more specific. in this case, could be

> ccmlib/common: prevent mistaking JVM 18 as JVM 8
>
> // and explain that we use "in" to check the versions, and "jre-18-openjdk" includes "8-openjdk", so the former is considered as "8-openjdk" by mistake

dkropachev: Great idea, done.
Commit message also fixed to be more detailed.

ccmlib/common.py
```diff
@@ -389,23 +389,46 @@ def make_cassandra_env(install_dir, node_path, update_conf=True, hardcode_java_v
...
 
+        env['JAVA_HOME'] = str(java_home_path.as_posix())
     return env
 
+def get_java_home_path(parent_path: pathlib.Path, hardcode_java_version: List[str]) -> [pathlib.Path]:
+    for path in parent_path.iterdir():
+        if not path.is_dir():
+            if path.name == "java" and get_jvm_spec_version(path) in hardcode_java_version:
+                return path.parent.parent
+            continue
```
tchaikov, line 411: why? this hurts the readability. `java` is always a symlink, which points to one of the directories we are checking here. so i'd suggest drop this recursive call.

ccmlib/common.py
```diff
@@ -389,23 +389,46 @@ def make_cassandra_env(install_dir, node_path, update_conf=True, hardcode_java_v
...
+        properties = subprocess.check_output([java_bin_path.as_posix(), '-XshowSettings:properties', '-version'], stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError as e:
+        return None
+    for line in properties.splitlines():
+        parts = line.strip().decode().split('=', 1)
+        if len(parts) < 2:
+            continue
+        name, value = [part.strip() for part in parts]
+        if name == 'java.specification.version':
+            return value
```
fruch, line 430: might be clearer just to use regex here

ccmlib/common.py
fruch: java executable is known inside each directory, I agree no point iterating over all of the files and directories

dkropachev:
```
➜  ~ find /usr/lib/jvm -name java 
/usr/lib/jvm/jdk-20/bin/java
/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
```

I clearly remember that even on older ubuntu distros java could land almost anywhere in this folder, depending on the package.

fruch: `/usr/lib/jvm/java-8-openjdk-amd64/bin/java` should exist as well...

anyhow, you are trying to fix a specific issue in this PR, and finding the java executable isn't part of it

dkropachev: Take a look again, now it looks way better with no recursion.

ccmlib/common.py
dkropachev: Done, please take a look.

ccmlib/common.py
```diff
@@ -389,23 +389,46 @@ def make_cassandra_env(install_dir, node_path, update_conf=True, hardcode_java_v
...
 
+def get_java_home_path(parent_path: pathlib.Path, hardcode_java_version: List[str]) -> [pathlib.Path]:
+    for java in pathlib.Path(parent_path).rglob('*/bin/java'):
+        if not java.is_file():
+            continue
+        if get_jvm_spec_version(java) in hardcode_java_version:
+            return java.parent.parent
+    return None
+
+java_version_property_regexp = re.compile('\s*(?:java\.specification\.version|java\.vm\.specification\.version)\s*=\s*([0-9.]+)\s*')
```
tchaikov, line 414: i don't really think the regular expression is more readable. i cannot parse this in 1 second.

ccmlib/common.py
tchaikov: IMHO, it's not clearer.

ccmlib/common.py
```diff
@@ -389,23 +389,46 @@ def make_cassandra_env(install_dir, node_path, update_conf=True, hardcode_java_v
...
 
+def get_java_home_path(parent_path: pathlib.Path, hardcode_java_version: List[str]) -> [pathlib.Path]:
+    for java in pathlib.Path(parent_path).rglob('*/bin/java'):
+        if not java.is_file():
+            continue
+        if get_jvm_spec_version(java) in hardcode_java_version:
+            return java.parent.parent
+    return None
+
+java_version_property_regexp = re.compile(r'\s*(?:java\.specification\.version|java\.vm\.specification\.version)\s*=\s*([0-9.]+)\s*')
```
tchaikov, line 414: this hurts the readability. i still think the split version is better.

ccmlib/common.py
```diff
@@ -389,23 +389,46 @@ def make_cassandra_env(install_dir, node_path, update_conf=True, hardcode_java_v
...
+        if java_home_path is None:
+            raise ArgumentError(f"java-{hardcode_java_version} wasn't found in {jvm_root_path}")
 
+        env['JAVA_HOME'] = str(java_home_path.as_posix())
     return env
 
+def get_java_home_path(parent_path: pathlib.Path, hardcode_java_version: List[str]) -> [pathlib.Path]:
+    for java in pathlib.Path(parent_path).rglob('*/bin/java'):
+        if not java.is_file():
+            continue
```
tchaikov, line 409: i don't think this happens in real world. but , meh..

ccmlib/common.py
```diff
@@ -389,23 +389,46 @@ def make_cassandra_env(install_dir, node_path, update_conf=True, hardcode_java_v
...
+        if get_jvm_spec_version(java) in hardcode_java_version:
+            return java.parent.parent
+    return None
+
+java_version_property_regexp = re.compile(r'\s*(?:java\.specification\.version|java\.vm\.specification\.version)\s*=\s*([0-9.]+)\s*')
+
+def get_jvm_spec_version(java_bin_path: pathlib.Path) -> Optional[str]:
+    if not java_bin_path.is_file():
```
tchaikov, line 417: so you check this again?

ccmlib/common.py
tchaikov: but .. i don't dare to assume as a reviewer.

fruch: in previous version you were looking for one specific key, now you are look for two different ones, why ?

ccmlib/common.py
dkropachev: 100%, but to be 100% sure, why not.

ccmlib/common.py
```diff
@@ -389,23 +389,46 @@ def make_cassandra_env(install_dir, node_path, update_conf=True, hardcode_java_v
...
+        return None
+
+    match = java_version_property_regexp.search(properties.decode('utf-8'))
+    if match is None:
+        return None
+    for version in match.groups():
+        return version
+    return None
+
+print(get_java_home_path(pathlib.Path("/usr/lib/jvm"), ["8", "1.8"]))
```
fruch, line 431: please that this test code out. (or put it in in tests folder)

ccmlib/common.py
dkropachev: i don't have strong opponion on that, too me regexp looks a bit better, but have downside of having regexp ).

@fruch, WDYT?

ccmlib/common.py
tchaikov: then you might want to check if it is executable or not.

ccmlib/common.py
dkropachev: removed from the cycle.

ccmlib/common.py
dkropachev: done.

ccmlib/common.py
dkropachev: you are 100% correct here, readability of the regexp somewhere around 0 for people who not deal with it on regular bases.

dkropachev: > in previous version you were looking for one specific key, now you are look for two different ones, why ?

It has changed in recent java versions.

ccmlib/common.py
fruch: go right ahead and add test code.

but this part if used in 5 cases where we need to run cassandra.

and we don't use java in master anymore, i.e. we won't run jmx or the old nodetool

I'm not sure in what context you are using this part, and chasing it the issue in it ? 
in using cassandra for the driver tests ?

dkropachev: > go right ahead and add test code.
> 
In progress.
> but this part if used in 5 cases where we need to run cassandra.
> 
> and we don't use java in master anymore, i.e. we won't run jmx or the old nodetool
> 
> I'm not sure in what context you are using this part, and chasing it the issue in it ? in using cassandra for the driver tests ?

while runing `java-driver` and `cpp-rust-driver` tests localy on my PC I have faced this issue with ccm picking wrong java, as result whole suit either fails or getting stuck.

fruch: running it with which scylla version ?

fruch: sounds like you are running cassandra and not scylla, cause running scylla you should be getting into that code path....

dkropachev: > go right ahead and add test code.

Done

ccmlib/common.py
dkropachev: added

ccmlib/common.py
dkropachev: > sounds like you are running cassandra and not scylla, cause running scylla you should be getting into that code path....

Correct, running on cassandra.

ccmlib/tests/test_results/ccm.log
fruch, line 1: please remove this log from the commit...

ccmlib/common.py
fruch: it's now clear enough, and with less logic to follow.

ccmlib/common.py
fruch: now it's better, look for one specific value, is straightforward, and can be understood in less then 1sec

fruch (approved):
LGTM

just take the log from the test out of the commit
maybe it need to be add to .gitignore

ccmlib/tests/test_results/ccm.log
dkropachev: done.

PR #604, opened 2024-09-07
ccmlib/common.py: remove cygwin code from `get_stress_bin`

dkropachev:
Removes cygwin support from `get_stress_bin`

(It used access unassigned variable `path` in certain case)

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -548,23 +548,11 @@ def get_stress_bin(install_dir):
     else:
         raise Exception("Cannot find stress binary (maybe it isn't compiled)")
 
-    # make sure it's executable -> win32 doesn't care
-    if sys.platform == "cygwin":
-        # Yes, we're unwinding the path join from above.
-        path = parse_path(stress)
-        short_bin = parse_bin(stress)
-        add_exec_permission(path, short_bin)
-    elif not os.access(stress, os.X_OK):
+    if not os.access(stress, os.X_OK):
         try:
-            # try to add user execute permissions
-            # os.chmod doesn't work on Windows and isn't necessary unless in cygwin...
-            if sys.platform == "cygwin":
-                add_exec_permission(path, stress)
-            else:
-                os.chmod(stress, os.stat(stress).st_mode | stat.S_IXUSR)
+            os.chmod(stress, os.stat(stress).st_mode | stat.S_IXUSR)
         except:
             raise Exception(f"stress binary is not executable: {stress}")
-
     return stress
 
 


dkropachev:
@fruch , @tchaikov , can you please take a look at it

ccmlib/common.py
```diff
@@ -556,12 +556,7 @@ def get_stress_bin(install_dir):
         add_exec_permission(path, short_bin)
     elif not os.access(stress, os.X_OK):
         try:
-            # try to add user execute permissions
```
tchaikov, line 559: @dkropachev hi Dmitry, ideally "fix get_stress_bin" should summarize the change you made in this change. but "fix" is not specific enough. also, in the commit message, could you please add more details on

- the symptom of the problem, it's impact. and
- why shall we make this change, 
- how do you address the problem

dkropachev: done

tchaikov: thanks . much better now. so one can understand the issue better just by skimming the title. but could you please go further?

> common: always ensure that stress is executable
>
> If stress binary does not have executable permissions on non cygwin
> platform `get_stress_bin` fails with:
> ```
> UnboundLocalError: local variable 'path' referenced before assignment
> ```
> 
> This PR removed rebase leftovers and make it call `os.chmod` in such
> case

also, on non-cygwin platform the code block of 

https://github.com/scylladb/scylla-ccm/blob/81076bce792a0fb3f2050e4c209a93e4a62ab55f/ccmlib/common.py#L558-L566

is executed. in which, https://github.com/scylladb/scylla-ccm/blob/81076bce792a0fb3f2050e4c209a93e4a62ab55f/ccmlib/common.py#L564 is executed. but in this branch, `path` is not referenced at all.

i think your change is a definite improvement. just cannot match it with the commit message though.

dkropachev: @tchaikov , i am sorry, where furthere should I go? This PR literaly removes 3 line of a code that is a leftover from the rebase mistake done years ago.

tchaikov: i explained that we cannot run into the error message based my understanding. not that i was asking for "more", but was not sure i understood the error message you quoted.

fruch: I don't mind really mind removing some windows support, we don't really support it (since scylla doesn't run on iwndows)

but @tchaikov question is valid, when do we to situations the stress command can't be executed ? and why ccm need to support it ?

also on master dtest for example, we don't use this code anymore, we switch to run c-s from a docker image, and not via CCM.

dkropachev: @tchaikov , @fruch  got it, please check if new version of the commit message and commit it self are ok with you.

fruch (approved):
LGTM

PR #605, opened 2024-09-08
tests/ccmcluster.py: make parse_cluster_status ignore lines that does not match

dkropachev:
It could fail with following error:
```
    def parse_cluster_status(self, stdout):
        """
            Output:

            Cluster: 'CCMCluster-reloc'
            ---------------------------
            node1: UP

            return:
            [(node1, UP)]
        """
        nodes_status = []
        for line in stdout.split("\n")[2:]:
>           node, status = line.split(":")
E           ValueError: not enough values to unpack (expected 2, got 1)
```

### DIFF
--- a/tests/ccmcluster.py
+++ b/tests/ccmcluster.py
@@ -113,7 +113,10 @@ class CCMCluster:
         """
         nodes_status = []
         for line in stdout.split("\n")[2:]:
-            node, status = line.split(":")
+            chunks = line.split(":")
+            if len(chunks) != 2:
+                continue
+            node, status = chunks
             nodes_status.append((node.strip(), status.strip()))
 
         return nodes_status


dkropachev:
@fruch , @tchaikov , one more please.

tchaikov (commented):
could improve the title a little bit:

> ccmlib/ccmcluster.py: ignore empty lines in parse_cluster_status()

tests/ccmcluster.py
```diff
@@ -113,7 +113,10 @@ def parse_cluster_status(self, stdout):
         """
         nodes_status = []
         for line in stdout.split("\n")[2:]:
-            node, status = line.split(":")
+            chunks = line.split(":")
+            if len(chunks) != 2:
```
tchaikov, line 117: presumably it was an empty which broke the test. an alternative is:
if not line:
    # ignore empty lines
    continue
node, status = line.split(":", 1)
```

tchaikov (approved):
just two nits. lgtm.

fruch (approved):
LGTM

fruch:
> could improve the title a little bit:
> 
> > ccmlib/ccmcluster.py: ignore empty lines in parse_cluster_status()

also it's tests/ccmcluster.py, which is kind of important context

tests/ccmcluster.py
dkropachev: > presumably it was an empty line which broke the test. an alternative is:

tbh, I have no idea which line broke it, but it is better to make it as resiliant as possible.

> ```python
> if not line:
>     # ignore empty lines
>     continue
> node, status = line.split(":", 1)
> ```

This code will still fail if line does not contain ':'.

dkropachev:
> > could improve the title a little bit:
> > > ccmlib/ccmcluster.py: ignore empty lines in parse_cluster_status()
> 
> also it's tests/ccmcluster.py, which is kind of important context

Updated.

tests/ccmcluster.py
tchaikov: yes, by looking at the error message, i think the offending line was empty. 

> as resiliant as possible.

i disagree. we program with a contract. "as resilient as possible" without a contract is error-prune, because we don't have a standard or contract for "as resilient as possible".

anyway, this is not that important in this context.

PR #609, opened 2024-09-12
node,scylla_node,cluster,scylla_cluster: Add type hints for nodes and clusters

cezarmoise:
- Add type hints for nodes and clusters to help debugging in dtest
- Fix linting / formatting issues

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -3,16 +3,19 @@
 import os
 import random
 import shutil
+import subprocess
 import threading
 import time
 from collections import OrderedDict, defaultdict
 from concurrent.futures import ThreadPoolExecutor
+from typing import List, Tuple
 
 from ruamel.yaml import YAML
 
 from ccmlib import common, repository
 from ccmlib.node import Node, NodeError
 from ccmlib.common import logger
+from ccmlib.scylla_node import ScyllaNode
 from ccmlib.utils.version import parse_version
 
 
@@ -23,8 +26,8 @@ class Cluster(object):
         self.id = 0
         self.ipprefix = None
         self.ipformat = None
-        self.nodes = OrderedDict()
-        self.seeds = []
+        self.nodes: OrderedDict[str, ScyllaNode] = OrderedDict()
+        self.seeds: List[ScyllaNode] = []
         self.partitioner = partitioner
         self.snitch = snitch
         self._config_options = {}
@@ -140,7 +143,7 @@ class Cluster(object):
             def __init__(self, cluster):
                 self.executor = ThreadPoolExecutor(max_workers=1)
                 self.thread = None
-                self.cluster = cluster
+                self.cluster: Cluster = cluster
                 self.req_stop_event = threading.Event()
                 self.done_event = threading.Event()
                 self.log_positions = defaultdict(int)
@@ -213,7 +216,7 @@ class Cluster(object):
     def hasOpscenter(self):
         return False
 
-    def nodelist(self):
+    def nodelist(self) -> List[ScyllaNode]:
         return [self.nodes[name] for name in list(self.nodes.keys())]
 
     def version(self):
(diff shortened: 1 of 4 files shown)


cezarmoise:
I will do a jenkins run to check for issues

cezarmoise:
> I would recommend add some linter that can identify those, and configure it specific to check those files in CI.
> 
> I think ruff might be use to force adding type notation
> 
> or `mypy` --disallow-untyped-functions

Those tools check all function parameters. For example, with ruff I get that there are about 3000 untyped parameters or function returns. 
I just added hints where nodes and clusters were involved, so it's easier for the editor to navigate trough their methods.

cezarmoise:
> I will do a jenkins run to check for issues

https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/564/
Ran without issues

cezarmoise:
https://github.com/scylladb/scylla-ccm/commit/df0a17a748a30e5fbd478290a734159b42f42535 rebased to solve conflicts

fruch:
> My only issue is that `typing.*` are deprecated since 3.9, because now you can use inbuilt types for typing instead, e.g. `test: list[str] = []`
> 
> See https://docs.python.org/3/library/typing.html#aliases-to-types-in-collections
> 
> Not sure whaat is our minimal working version, but 3.9 seems old & stable enough to do it the new way. @fruch WDYT?

The minimum is 3.8, we need to support the same versions as the python-driver does, cause ccm is used in it's testing

The typing.* isn't yet deprecated AFAIK, linters are suggesting not to use it.

ccmlib/node.py
```diff
@@ -1125,7 +1127,7 @@ def get_sstablespath(self, output_file=None, datafiles=None, keyspace=None, tabl
         return sstablefiles
 
     def run_sstablerepairedset(self, set_repaired=True, datafiles=None, keyspace=None, column_families=None):
-        cdir = self.get_install_dir()
+        cdir = self.get_install_dir()  # noqa: F841
```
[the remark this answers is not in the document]
fruch: Cause we never linted this code.

This code path also probably isn't in use...

ccmlib/cluster.py
```diff
@@ -246,14 +249,14 @@ def add(self, node, is_seed, data_center=None, rack=None):
 
     # nodes can be provided in multiple notations, determining the cluster topology:
     # 1. int - specifying the number of nodes in a single DC, single RACK cluster
-    # 2. list[int] - specifying the number of nodes in a multi DC, single RACK per DC cluster
+    # 2. List[int] - specifying the number of nodes in a multi DC, single RACK per DC cluster
```
fruch, line 252: No real need to change those comments

ccmlib/cluster.py
```diff
@@ -328,8 +331,8 @@ def populate(self, nodes, debug=False, tokens=None, use_vnodes=False, ipprefix=N
             self._update_config()
         return self
 
-    def new_node(self, i, auto_bootstrap=False, debug=False, initial_token=None, add_node=True, is_seed=True, data_center=None, rack=None):
-        ipformat = self.get_ipformat()
+    def new_node(self, i, auto_bootstrap=False, debug=False, initial_token=None, add_node=True, is_seed=True, data_center=None, rack=None) -> ScyllaNode:
+        ipformat = self.get_ipformat()  # noqa: F841
```
fruch, line 335: What is the point of introducing noqa ? If we don't introduce the linter as well ?

ccmlib/cluster.py
```diff
@@ -365,7 +368,7 @@ def get_storage_interface(self, nodeid):
         return (self.get_node_ip(nodeid), 7000)
 
     def get_node_jmx_port(self, nodeid):
-        return 7000 + nodeid * 100 + self.id;
```
fruch, line 368: Shame to get rid of ANSI C, isn't it ? 😜

ccmlib/cluster.py
```diff
@@ -421,7 +424,7 @@ def remove_dir_with_retry(self, path):
                 try:
                     common.rmdirs(path)
                     removed = True
-                except:
+                except Exception:
```
fruch, line 427: Linter should be angry about this, isn't it ? too broad exception ?

I think in this case we can set a specific set of possible exceptions

fruch (approved):
LGTM

Issue #610, opened 2024-09-16
Fix Node._parse_pending_tasks and Node.wait_for_compactions to account for tasks in progress


cezarmoise:
The current code doesn't take into account tasks in progress, so tests that depend on `wait_for_compactions` will continue while compactions are not finished and fail.

It expects output like:
```
 pending tasks: 42
 - ks1.cf1: 13
 - ks1.cf2: 19
 - ks2.cf1: 10
  ```
but the actual output is like:
```
 pending tasks: 42
 - ks1.cf1: 13
 - ks1.cf2: 19
 - ks2.cf1: 10

id                                   compaction type keyspace table completed total unit progress
55eaee80-7445-11ef-9197-2931a44dadc4 COMPACTION      ks3      cf3   32116     55680 keys 57.68%  
55e8f2b0-7445-11ef-b438-2930a44dadc4 COMPACTION      ks4      cf4   46789     55936 keys 83.65%  
Active compaction remaining time :        n/a
```

See https://github.com/scylladb/scylladb/commit/a7abaa457b892c8e22552dabb67d8be35013c715

Found during https://github.com/scylladb/scylla-dtest/pull/4938, which relates to https://github.com/scylladb/scylla-dtest/issues/4702.

PR #616, opened 2024-09-24
[Backport 2024.1] ccmlib/node: check both pending and active tasks when waiting for compactions

cezarmoise:
Currently, when using `wait_for_compactions`, only pending tasks are considered. If there are active compaction tasks, the function will return anyway and this causes flakyness in tests, as sometimes the active task finishes before the next operation, but sometimes it doesn't.

- Rename Node._parse_pending_tasks to Node._parse_tasks
- Count all tasks, pending and active
- Allow searching for tasks based on only keyspace
- Update and refactor test to allow more varied cases

closes: #610 
refs: https://github.com/scylladb/scylla-dtest/issues/4702

cherry-picked ee7ebefe86b18ee2de91d700cd54f145e8dde8e1^..4b1ac161a6212d6f6a5bc0adfe755482e3a55d33

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -77,7 +77,7 @@ jobs:
 
     - name: Copy logs/results
       if: contains(github.event.pull_request.labels.*.name, 'PR-upload-log')
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v4
       with:
         name: ccm-tests-log
         path: tests/test_results/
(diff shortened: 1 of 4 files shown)


cezarmoise:
PR checks fail with 
> Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v2`. Learn more: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

fruch:
branch-2024.1 was missing a fix to the CI, adding it here.

fruch:
now it's one broken test, that try to run the latest scylla, and expect it to have JMX

it's a series of change not going to backport, on next backports we should disable that test

PR #621, opened 2024-10-17
scylla_node: add support to run `scylla types` tool

lkshminarayanan:
This PR implements the `run_scylla_types` method in the ScyllaNode class
to run the `types` tool. Also added a new class, ScyllaKeyType to allow
callers to define the key type that is being sent to the types tool.

This is required by scylladb/scylla-dtest#5093 to deduce shards of the
partition keys.

Refs scylladb/scylladb#18011

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -15,7 +15,8 @@ import time
 import threading
 from pathlib import Path
 from collections import OrderedDict
-from typing import Any, Dict, List, Optional, Tuple, Union, TYPE_CHECKING
+from typing import Any, Dict, Iterable, List, Optional, Tuple, Union, TYPE_CHECKING
+from enum import Enum
 
 import logging
 
@@ -37,6 +38,55 @@ from ccmlib.utils.version import parse_version
 
 if TYPE_CHECKING:
     from ccmlib.scylla_cluster import ScyllaCluster
+
+class ScyllaType:
+    """
+    Helper class for defining Scylla type definitions using type strings.
+
+    Refer to the Cassandra class names found at
+    https://github.com/scylladb/scylladb/blob/master/docs/dev/cql3-type-mapping.md for valid type strings.
+
+    Examples :
+        type1 = ScyllaType.make_partition_key("Int32Type")
+        type2 = ScyllaType.make_clustering_key("Int32Type", "FloatType")
+    """
+    class TypeKind(Enum):
+        REGULAR = 1
+        CLUSTERING_KEY = 2
+        PARTITION_KEY = 3
+
+    kind: TypeKind
+    types: Iterable[str]
+
+    def __init__(self, kind: TypeKind, types: Iterable[str]):
+        if not types:
+            raise common.ArgumentError("Please pass at least one type to create ScyllaType")
+        self.kind = kind
+        self.types = types
+
+    @classmethod
+    def make_regular(cls, keytype: str):
+        return cls(ScyllaType.TypeKind.REGULAR, (keytype,))
+
+    @classmethod
+    def make_clustering_key(cls, *keytypes):
+        return cls(ScyllaType.TypeKind.CLUSTERING_KEY, keytypes)
+
+    @classmethod
+    def make_partition_key(cls, *keytypes):
+        return cls(ScyllaType.TypeKind.PARTITION_KEY, keytypes)
+
+    def as_types_args(self) -> List[str]:
+        """Return the type definition as arguments to scylla types command"""
+        args = []
+        if self.kind == ScyllaType.TypeKind.CLUSTERING_KEY:
+            args.append("--prefix-compound")
+        elif self.kind == ScyllaType.TypeKind.PARTITION_KEY:
+            args.append("--full-compound")
+        for keytype in self.types:
+            args.extend(["-t", keytype])
+        return args
+
 
 class ScyllaNode(Node):
 
@@ -1784,6 +1834,33 @@ class ScyllaNode(Node):
         super().wait_for_compactions(keyspace=keyspace,
                                      column_family=column_family,
                                      idle_timeout=idle_timeout)
+
+    def run_scylla_types(self, action: str, scylla_type: ScyllaType, *values: Tuple[Any], extra_args: Optional[List[Any]]=None):
+        """Invoke scylla-types command, with the specified action and extra_args.
+
+        For more information about scylla-types, see https://opensource.docs.scylladb.com/stable/operating-scylla/admin-tools/scylla-types.html
+
+        Params:
+        * action - The scylla-types operation to run.
+        * scylla_type - The key type of the values being passed to the tool.
+        * value - The values on which the action is to be applied.
+        * extra_args - Additional command-line arguments to pass to scylla-types action.
+
+        Returns: stdout
+
+        Raises: ToolError if scylla-types returns a non-zero exit code.
+        """
+        if extra_args is None:
+            extra_args = []
+        extra_args.extend(scylla_type.as_types_args())
+
+        scylla_path = common.join_bin(self.get_path(), BIN_DIR, 'scylla')
+        cmd_and_args = [scylla_path, "types", action, *extra_args, "--", *values]
+        cmd_and_args = [str(value) for value in cmd_and_args]
+        res = subprocess.run(cmd_and_args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, check=False, env=self._get_environ())
+        if res.returncode:
+            raise ToolError(command=' '.join(cmd_and_args), exit_status=res.returncode, stdout=res.stdout, stderr=res.stderr)
+        return res.stdout.strip()
 
 
 class NodeUpgrader:


ccmlib/scylla_node.py
```diff
@@ -1785,6 +1785,78 @@ def wait_for_compactions(self,
...
+        * types : cassandra type class name of the values' types (example : ShortType, Int32Type, TimeUUIDType, etc.)
+        * prefixable_compounds: if true, values are prefixable compounds (e.g. clustering key)
+                                else, the values are full compounds (e.g. partition keys)
+
+        Returns: serialized value
+
+        Raises: subprocess.CalledProcessError if scylla-types serialize returns a non-zero exit code.
+
+        """
+        if len(values) > len(types):
```
pehala, line 1829: What about using a dict? That way it cannot get mismatched

lkshminarayanan: @pehala  We cannot use a dict because we need to allow duplicate keys.

I tried using List[Tuple[str, str]], but if prefixable_compounds is set to True, the length of values can be less than the length of types and the existing code seemed more user friendly for such cases. So, I have left it as it is.

ccmlib/scylla_node.py
```diff
@@ -1785,6 +1785,78 @@ def wait_for_compactions(self,
...
+        return (res.stdout, res.stderr)
+
+    def serialize_values(self, values: List[str], types: List[str], prefixable_compounds=False) -> str:
+        """Serialize the given values using scylla-types serialize action
+
+        Params:
+        * values : the values to be serialized
+        * types : cassandra type class name of the values' types (example : ShortType, Int32Type, TimeUUIDType, etc.)
+        * prefixable_compounds: if true, values are prefixable compounds (e.g. clustering key)
+                                else, the values are full compounds (e.g. partition keys)
```
denesb, line 1822: scylla-types also works with simple types, so this should be a tri-state, a bool cannot express all values.

ccmlib/scylla_node.py
```diff
@@ -1785,6 +1785,78 @@ def wait_for_compactions(self,
...
+            additional_args.extend(["-t", value_type])
+
+        scylla_path = common.join_bin(self.get_path(), BIN_DIR, 'scylla')
+        scylla_types_cmd_with_args = [scylla_path, "types", action] + additional_args + ["--"] + values
+        res = subprocess.run(scylla_types_cmd_with_args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=text, check=False, env=self._get_environ())
+        if res.returncode:
+            raise ToolError(command=' '.join(scylla_types_cmd_with_args), exit_status=res.returncode, stdout=res.stdout, stderr=res.stderr)
+        return (res.stdout, res.stderr)
+
+    def serialize_values(self, values: List[str], types: List[str], prefixable_compounds=False) -> str:
```
denesb, line 1815: I'd prefer if we had just `run_scylla_types()`, instead of exposing a separate methods for the different actions, this will cause an explosion of methods.

ccmlib/scylla_node.py
```diff
@@ -1785,6 +1785,78 @@ def wait_for_compactions(self,
                                      column_family=column_family,
                                      idle_timeout=idle_timeout)
 
+    def run_scylla_types(self, action: str, values: List[str], types: List[str], additional_args: Optional[List[str]]=None, text=True):
```
denesb, line 1788: We can use a convenience type to help people with the `types`:
```python

class TypeKind(Enum):
    Simple = 1
    ClusteringKey = 2
    PartitionKey = 3

class scylla_types:
    kind: TypeKind
    types: Tuple[str]

    def __init__(self, kind, types):
        self.kind = kind
        self.types = types

    @staticmethod
    def make_simple(type: str):
        return Type(TypeKind.Simple, (type,))

    @staticmethod
    def make_clustering_key(*types):
        return Type(TypeKind.ClusteringKey, types)

    @staticmethod
    def make_partition_key(*types):
        return Type(TypeKind.PartitionKey, types)
```

Maybe we can add an num with all the different types too, to save people a lookup.
See https://github.com/scylladb/scylladb/blob/master/docs/dev/cql3-type-mapping.md for a table of all the types.

denesb (changes requested):

ccmlib/scylla_node.py
pehala: I would not add the enum with all the different type, this PR would explode with content. I would like to keep this simple. I like the idea of ScyllaType class with kind, only I would use class_method instead. I think this class could solve `values `and `list` handling

denesb: Yes, I later realized that adding an enum with all types is problematic also because some types are compound, e.g. lists and that is hard to express with an enum. So we can leave type as a free string, but let's add a link to the above document in the comments so people can quickly find the proper type names.

ccmlib/scylla_node.py
lkshminarayanan: Done.

ccmlib/scylla_node.py
lkshminarayanan: Done.

ccmlib/scylla_node.py
lkshminarayanan: I have updated the code based on your feedback. Let me know your comments on that.

lkshminarayanan:
Changelog V2 :
 - Removed the separate methods for different actions
 - Added the ScyllaKeyType class to allow define a key type

ccmlib/scylla_node.py
```diff
@@ -1785,6 +1832,32 @@ def wait_for_compactions(self,
...
+        Returns: stdout
+
+        Raises: ToolError if scylla-types returns a non-zero exit code.
+        """
+        if extra_args is None:
+            extra_args = []
+        extra_args.extend(key_type.as_types_args())
+
+        scylla_path = common.join_bin(self.get_path(), BIN_DIR, 'scylla')
+        scylla_types_cmd_with_args = [scylla_path, "types", action] + [str(arg) for arg in extra_args] + ["--"] + [str(value) for value in values]
```
pehala, line 1855: `[scylla_path, "types", action, *(str(arg) for arg in extra_args), "--", *(str(value) for value in values)]` or
```
args = [scylla_path, "types", action, *extra_args, "--", *values]
args = [str(value) for value in args]
```
Less lists is always better

ccmlib/scylla_node.py
```diff
@@ -38,6 +39,52 @@
...
+        type1 = ScyllaKeyType.make_partition_key("Int32Type")
+        type2 = ScyllaKeyType.make_clustering_key("Int32Type", "FloatType")
+    """
+    class TypeKind(Enum):
+        REGULAR_KEY = 1
+        CLUSTERING_KEY = 2
+        PARTITION_KEY = 3
+
+    kind: TypeKind
+    types: Tuple[str]
```
pehala, line 56: Does this need to be tuple, from the usages it seems you can use `Iterable[str]` instead?

ccmlib/scylla_node.py
```diff
@@ -38,6 +39,52 @@
 if TYPE_CHECKING:
     from ccmlib.scylla_cluster import ScyllaCluster
 
+class ScyllaKeyType:
```
denesb, line 42: Note that `scylla types` works with regular types -- i.e. values belonging to regular columns -- not just keys. So this name is misleading.

ccmlib/scylla_node.py
```diff
@@ -38,6 +39,52 @@
...
+    """
+    Helper class to define Scylla Key definition with Type strings.
+    Use the cassandra class name from https://github.com/scylladb/scylladb/blob/master/docs/dev/cql3-type-mapping.md as type strings.
+    Examples :
+        type1 = ScyllaKeyType.make_partition_key("Int32Type")
+        type2 = ScyllaKeyType.make_clustering_key("Int32Type", "FloatType")
+    """
+    class TypeKind(Enum):
+        REGULAR_KEY = 1
```
denesb, line 51: Just `REGULAR`, see above.
Also apply the change to other places where `regular` is used.

denesb (changes requested):

ccmlib/scylla_node.py
lkshminarayanan: Ack. I misunderstood the usage. I have the updated the code now.

ccmlib/scylla_node.py
lkshminarayanan: Done.

ccmlib/scylla_node.py
lkshminarayanan: Ack. I have replaced the code with the second variant. Thanks!

ccmlib/scylla_node.py
lkshminarayanan: Yes, I have replaced it now.

lkshminarayanan: Neither the iterable from typing nor the Iterable from collections.abc work in 3.8. So, I have changed it back to Tuple[str]. Is that okay?

ccmlib/scylla_node.py
```diff
@@ -38,6 +39,55 @@
...
+        REGULAR = 1
+        CLUSTERING_KEY = 2
+        PARTITION_KEY = 3
+
+    kind: TypeKind
+    types: Tuple[str]
+
+    def __init__(self, kind: TypeKind, types: Tuple[str]):
+        if not types:
+            raise common.ArgumentError("Please pass atleast one type to create ScyllaType")
```
tchaikov, line 63: s/atleast/at least/

ccmlib/scylla_node.py
pehala: I think it should work https://docs.python.org/3.8/library/typing.html#typing.Iterable, but I am fine with leaving Tuple in as well

lkshminarayanan: It worked! Thank you, I have replaced it.

pehala (approved):

ccmlib/scylla_node.py
```diff
@@ -1785,6 +1835,33 @@ def wait_for_compactions(self,
                                      column_family=column_family,
                                      idle_timeout=idle_timeout)
 
+    def run_scylla_types(self, action: str, key_type: ScyllaType, *values: Tuple[Any], extra_args: Optional[List[Any]]=None):
```
denesb, line 1838: nit: s/key_type/scylla_type/

denesb (approved):
LGTM, one nit remaining.

ccmlib/scylla_node.py
lkshminarayanan: Fixed it.

denesb (approved):

denesb:
@fruch please review/merge.

PR #619, opened 2024-09-29
scylla_node: ignore scylla-tools config files if scylla-tools is missing

tchaikov:
this change is created in the same spirit of 3a886404, so that we don't assume the existence of scylla-tools. otherwise we could have following failure:

```
07:51:05  Traceback (most recent call last):
07:51:05    File "/jenkins/workspace/scylla-master/gating-dtest-release/scylla/.local/bin/ccm", line 74, in <module>
07:51:05      cmd.run()
07:51:05    File "/jenkins/workspace/scylla-master/gating-dtest-release/scylla/.local/lib/python3.12/site-packages/ccmlib/cmds/cluster_cmds.py", line 269, in run
07:51:05      cluster.populate(self.nodes, self.options.debug, use_vnodes=self.options.vnodes, ipprefix=self.options.ipprefix, ipformat=self.options.ipformat)
07:51:05    File "/jenkins/workspace/scylla-master/gating-dtest-release/scylla/.local/lib/python3.12/site-packages/ccmlib/cluster.py", line 330, in populate
07:51:05      self.new_node(i, debug=debug, initial_token=tk, data_center=dc, rack=rack)
07:51:05    File "/jenkins/workspace/scylla-master/gating-dtest-release/scylla/.local/lib/python3.12/site-packages/ccmlib/cluster.py", line 337, in new_node
07:51:05      node = self.create_node(name=f'node{i}',
07:51:05             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
07:51:05    File "/jenkins/workspace/scylla-master/gating-dtest-release/scylla/.local/lib/python3.12/site-packages/ccmlib/scylla_cluster.py", line 94, in create_node
07:51:05      return ScyllaNode(name, self, auto_bootstrap, None,
07:51:05             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
07:51:05    File "/jenkins/workspace/scylla-master/gating-dtest-release/scylla/.local/lib/python3.12/site-packages/ccmlib/scylla_node.py", line 53, in __init__
07:51:05      super().__init__(name, cluster, auto_bootstrap,
07:51:05    File "/jenkins/workspace/scylla-master/gating-dtest-release/scylla/.local/lib/python3.12/site-packages/ccmlib/node.py", line 128, in __init__
07:51:05      self.import_config_files()
[...]

PR #623, opened 2024-11-06
scylla_node: ignore scylla-tools files if scylla-tools is missing

denesb:
This is a follow-up to https://github.com/scylladb/scylla-ccm/pull/619. Patch scylla_node.py so every access to java tools dir is conditional on the dir existing in the first place. With this PR, tests are passing with the java tools dropped form the unified package (https://github.com/scylladb/scylladb/pull/20740).

Refs: https://github.com/scylladb/scylladb/issues/14856

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -214,6 +214,8 @@ class ScyllaNode(Node):
         raise NotImplementedError('ScyllaNode.get_tool_args')
 
     def get_env(self):
+        if not self.get_tools_java_dir():
+            return os.environ.copy()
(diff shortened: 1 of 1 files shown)


fruch:
@denesb are you sure you tested ontop of #619  ?

since it was merged 4 hours ago, and you have a conflict with it.

denesb:
> @denesb are you sure you tested ontop of #619 ?
> 
> since it was merged 4 hours ago, and you have a conflict with it.

Ah, seems to be the same thing. I guess my PR is not needed after all. Closing.

PR #624, opened 2024-11-07
ccmlib/scylla_node.py: fix get_env() for non-cassandra nodes

denesb:
0ef455c544d8ea4da7c8d2d72ad1575fe5a325f4 changes get_env() to return empty env for scylla nodes, as the native tools do not need any environment variables, unlike the previous java ones which we are about to drop. This however broke cqlsh, which needs the current environment to work correctly, but returning empty {} as currently happens resets the environment vars.
Fix this by returning a clone of the current env, instead of reseting it.


fruch (approved):
LGTM

PR #626, opened 2024-11-14
ccmlib/common.py: update `pid_listening_on` to return multiple pids

fruch:
in some situations, like when running cassandra 4.1, we can endup with some of the commands in this function return multiple pids and not just one process

to handle that we just return all of the pids back. in the enad it's just used for debug printing about who's holding the ports.

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -730,15 +730,17 @@ def validate_install_dir(install_dir):
             raise ArgumentError(f'{install_dir} does not appear to be a cassandra or dse installation directory')
 
 
-def pid_listening_on(addr: str, port: int):
-    '''get pid of the process listening on the given port'''
-    pid = -1
+def pids_listening_on(addr: str, port: int) -> List[int]:
+    """get pids of the processes listening on the given address:port"""
+
+    pids = []
     if lsof := shutil.which('lsof'):
         result = subprocess.run([lsof, '-t', '-i', f'tcp@{addr}:{port}'],
                                 stdout=subprocess.PIPE,
+                                text=True,
                                 check=True)
         if result.stdout:
-            pid = int(result.stdout)
+            pids = [int(p) for p in result.stdout.split()]
     elif ss := shutil.which('ss'):
         result = subprocess.run([ss,
                                  '--no-header',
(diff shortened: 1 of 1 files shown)


ccmlib/common.py
```diff
@@ -748,12 +748,12 @@ def pid_listening_on(addr: str, port: int):
                                  f'src = inet:{addr}:{port}'],
                                 stdout=subprocess.PIPE,
                                 check=True)
-        matched = re.search(r'pid=(\d+)', result.stdout)
+        matched = re.search(r'pid=(\d+)', str(result.stdout))
```
tchaikov, line 751: probably we can instead pass `text=True` to `subprocess.run()`, so it can decode the output using the default encoding for us? like
```py
        result = subprocess.run([ss,
                                 '--no-header',
                                 '--listening',
                                 '--processes',
                                 '--tcp',
                                 f'src = inet:{addr}:{port}'],
                                stdout=subprocess.PIPE,
                                check=True,
                                text=True)
```

ccmlib/common.py
```diff
@@ -748,12 +748,12 @@ def pid_listening_on(addr: str, port: int):
...
         if matched is not None:
-            pid = int(matched.group(1))
+            pids = [int(matched.group(1))]
```
tchaikov, line 753: if we are interested in find all matches, probably we can use 
```py
matches = re.findall(r'pid=(\d+)', result.stdout)
pids = [int(match) for match in matches]
```

because, `re.search()` only returns the first match. FWIW, there is no need to pass `re.MULTILINE` as the `flags` passed to `re.findall()`.

ccmlib/common.py
```diff
@@ -730,15 +730,15 @@ def validate_install_dir(install_dir):
             raise ArgumentError(f'{install_dir} does not appear to be a cassandra or dse installation directory')
 
 
-def pid_listening_on(addr: str, port: int):
+def pids_listening_on(addr: str, port: int) -> List[int]:
     '''get pid of the process listening on the given port'''
-    pid = -1
+    pids = []
```
tchaikov, line 735: nit, could update the docstring accordingly.

ccmlib/common.py
fruch: yeah, I thought about it

but in the situation I go, this command didn't really returned multiple processes, while `lsof` did
so I gave it up.

but can change that one as well

ccmlib/common.py
fruch: change the docstring

ccmlib/common.py
fruch: added the `text=True`

ccmlib/common.py
fruch: replaced search with findall

tchaikov (approved):
thanks. lgtm

PR #630, opened 2025-01-21
Introduce UV as a dependency management

pehala:
### Overview
* Alternative to #620 
* Migrate to using `pyproject.toml`
  * Keep setuptools as a build backend 
* Introduces UV for python & dependency management
* First step to solving #555  
* Add `main()` method to `ccm` script
   * Easier to refer to said method
* Update workflows action versions
   * Solves deprecated warnings about node12 
   * Unnecessary for this PR, I can extract it to a separate one if needed


.github/workflows/ci-tests.yml
```diff
@@ -15,9 +15,9 @@ jobs:
     name: CCM Integration tests
     uses: ./.github/workflows/integration-tests.yml
     with:
-      reloc_version: "2023-04-13T13:31:00Z"
+      reloc_version: "2025-01-19T09:39:05Z"
```
fruch, line 18: it won't be enough, one would need to update the tests as well

also I think maybe pinning down the runner ubuntu version would help for now

pehala: I managed to fix it by using java11 and newer version of the build. I however had to change the version in the tests as well and fixed some test bugs I encountered as well. I will extract the changes into a separate PR

pehala:
@fruch https://jenkins.scylladb.com/view/master/job/scylla-master/job/byo/job/dtest-byo/902/ passed ok

fruch (approved):
LGTM

pehala:
@fruch One thing I apparently did here, is to change version from 2.0.5 to 0.1.0? Does it matter?

dkropachev:
We need to revert it, it broken ccm:
```
Download Scylla (2025.1.1) image        0s

Run rm -rf /tmp/download.ccm || true 
  rm -rf /tmp/download.ccm || true 
  mkdir /tmp/download.ccm || true
  ccm create ccm_1 -i 127.0.1. -n 3:0 -v "release:2025.1.1" --scylla --config-dir=/tmp/download.ccm
  rm -rf /tmp/download.ccm
  shell: /usr/bin/bash -e {0}
  env:
    JAVA_HOME: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/8.0.442-6/x64
    JAVA_HOME_8_X64: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/8.0.442-6/x64
    pythonLocation: /opt/hostedtoolcache/Python/3.11.12/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.11.12/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.12/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.12/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.12/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.11.12/x64/lib
  File "/opt/hostedtoolcache/Python/3.11.12/x64/bin/ccm", line 5, in <module>
    from ccm import main
ModuleNotFoundError: No module named 'ccm'
Error: Process completed with exit code 1.
```

pehala:
Could it be caused by rhe changed version, or do you believe there is a bigger problem here?

fruch:
I can confirm it work, like the following
```
uv python install 3.11

uv venv -p 3.11 ~/ttt
source ~/ttt/bin/activate
python -m pip install https://github.com/pehala/scylla-ccm/archive/refs/heads/change-version.zip
ccm --help
```

with master version the ccm call was broken

PR #631, opened 2025-01-21
Update and fix workflows and tests

pehala:
* Updates Scylla version to `2025-01-19T09:39:05Z`
  * Affects both workflows and tests 
  * Update `test_get_default_scylla_yaml` to fetch different variable, `consistent_cluster_management` is no longer in default config
* Update to Java 11
   * `2025-01-19T09:39:05Z` does not work with Java 8
* Fix `test_config` getting SCYLLA_VERSION from env
   * Tests have hardcoded version
* Fix version specification for 2024.2
   * We cannot expect `2024.2.0` as newer releases are already out

PR #632, opened 2025-01-27
Remove old_tests

pehala:
They are not used anymore, I do not see value in keeping them. If we want to keep them for posterity, I think separate branch might do the trick as well


fruch (approved):
LGTM

pehala:
Rebased on top of next to include #631

PR #634, opened 2025-02-13
Setup the launch env for running scylla nodetool

bhalevy:
When using a locally-built scylla binary, we must set `LD_LIBRARY_PATH=dbuild_so_dir` so that the binary will find its proper dynamic libraries.

Note that we run patchelf to setup the loader for the scylla binary to `os.path.join(dbuild_so_dir, 'ld-linux-x86-64.so.*')`, but that's not enough for it to locate other dynamic libs it might need.


ccmlib/scylla_node.py
```diff
@@ -216,6 +216,10 @@ def get_tool_args(self, toolname):
         raise NotImplementedError('ScyllaNode.get_tool_args')
 
     def get_env(self):
+        try:
+            return self._launch_env
+        except AttributeError:
```
pehala, line 221: Could we fix the assignment of this variable so it is null instead of missing instead? Catching AttributeError is very ugly

fruch (approved):
LGTM

bhalevy:
In v2 ([bb1e442](https://github.com/scylladb/scylla-ccm/commit/bb1e442b1654d5fe4338696557a03f985c17bb0f)):
- initialize ScyllaNode._launch_env to None

ccmlib/scylla_node.py
bhalevy: Done in v2.
@pehala Please re-review

pehala (approved):

bhalevy:
@scylladb/dtest-maintainers please merge

PR #635, opened 2025-02-19
scylla_node: __init__: initialize _launch_env early

bhalevy:
With f3cbb1d, `_get_environ` accesses self._launch_env assuming it is always initialized.

However, `_get_environ` is called indirectly from the superclass: None.__init__ function, via `import_bin_files()`.


pehala (approved):

fruch (approved):
LGTM

Issue #638, opened 2025-03-05
CCM fails to fetch scylla 2025.1 relocatables from web


muzarski:
Currently, ccm is unable to fetch scylla 2025.1 RCs from web.

```bash
$ ./ccm create --scylla -n 2:0 -i 127.0.0. --version="release:2025.1" -b test_cluster



Traceback (most recent call last):
  File "scylla-ccm/./ccm", line 74, in <module>
    cmd.run()
    ~~~~~~~^^
  File "scylla-ccm/ccmlib/cmds/cluster_cmds.py", line 220, in run
    cluster = ScyllaCluster(self.path, self.name, install_dir=self.options.install_dir,
                            version=self.options.version, verbose=True, manager=manager_install_dir)
  File "scylla-ccm/ccmlib/scylla_cluster.py", line 53, in __init__
    super(ScyllaCluster, self).__init__(path, name, partitioner,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
                                        install_dir, create_directory,
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                        version, verbose,
                                        ^^^^^^^^^^^^^^^^^
                                        snitch=SNITCH, cassandra_version=cassandra_version,
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                        docker_image=docker_image)
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "scylla-ccm/ccmlib/cluster.py", line 72, in __init__
    dir, v = self.load_from_repository(version, verbose)
             ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "scylla-ccm/ccmlib/scylla_cluster.py", line 74, in load_from_repository
    install_dir, version = scylla_repository.setup(version, verbose)
                           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "scylla-ccm/ccmlib/scylla_repository.py", line 282, in setup
    packages, type_n_version[1] = release_packages(s3_url=s3_url, version=s3_version, arch=scylla_arch, scylla_product=scylla_product + scylla_mode_suffix)
[...]

PR #639, opened 2025-03-05
Fix fetching scylla 2025.x relocatables

muzarski:
Fixes: https://github.com/scylladb/scylla-ccm/issues/638

The problem is well described in the linked issue. TLDR, it looks like we dropped "scylla-enterprise-" prefix for 2025.x relocatables. New prefix is "scylla-" (as for previous open source versions).


fruch (approved):
LGTM

PR #640, opened 2025-03-09
sni_proxy.py: add option to select the temp directory used

fruch:
since when we are working inside podman we can't mount /tmp and we need the exact same direcorty mount as host for sni_proxy to work as expected

### DIFF
--- a/ccmlib/utils/sni_proxy.py
+++ b/ccmlib/utils/sni_proxy.py
@@ -194,8 +194,8 @@ def get_cluster_info(cluster, port=9142):
     return nodes_info
 
 
-def refresh_certs(cluster, nodes_info):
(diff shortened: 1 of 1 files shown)


mykaul:
nit: SNI proxy was deprecated. Not sure it won't be back again in the future, but right now, deprecated.

fruch:
> nit: SNI proxy was deprecated. Not sure it won't be back again in the future, but right now, deprecated.

testing of it is still in dtest, and in all drivers.

mykaul:
> > nit: SNI proxy was deprecated. Not sure it won't be back again in the future, but right now, deprecated.
> 
> testing of it is still in dtest, and in all drivers.

We shouldn't.

Issue #642, opened 2025-03-18
Initial token calculation for multiple DC is broken


dkropachev:
Provision multi DC cluster and list `initial_token`:
```
rm -rf ~/.ccm/test
ccm create test -i 127.0.0. -n 4:0 -v release:6.2 --scylla
cat ~/.ccm/test/*/node.conf | grep init
initial_token: -9223372036854775808
initial_token: -9223372036854775708
```

As you can see only two nodes got `initial_token`

PR #643, opened 2025-03-18
Fix multi DC token calculation

dkropachev:
Provision multi DC cluster and list `initial_token`:
```
rm -rf ~/.ccm/test
ccm create test -i 127.0.0. -n 4:0 -v release:6.2 --scylla
cat ~/.ccm/test/*/node.conf | grep init
initial_token: -9223372036854775808
initial_token: -9223372036854775708
```

As you can see only two nodes got `initial_token`
To fix that we need to pass `node_locations` to `balanced_tokens_across_dcs` And have ti fixed a bit.

After the fix:
```
rm -rf ~/.ccm/test
ccm create test -i 127.0.0. -n 4:0 -v release:6.2 --scylla
cat ~/.ccm/test/*/node.conf | grep init
Current cluster is now: test
initial_token: -9223372036854775808
initial_token: -4611686018427387904
initial_token: 0
initial_token: 4611686018427387904
```

Fixes: https://github.com/scylladb/scylla-ccm/issues/642

### DIFF
--- a/ccmlib/cluster.py
+++ b/ccmlib/cluster.py
@@ -319,7 +319,7 @@ class Cluster(object):
             if dcs is None or len(dcs) <= 1:
                 tokens = self.balanced_tokens(node_count)
             else:
-                tokens = self.balanced_tokens_across_dcs(dcs)
+                tokens = self.balanced_tokens_across_dcs(node_locations)
 
         assert node_count == len(node_locations)
         for i in range(1, node_count + 1):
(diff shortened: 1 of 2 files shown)


dkropachev:
@fruch , @Bouncheck

fruch:
1) not sure how one of the nodes has 0, I don't understand the logic here
2) we should run dtest gating to validate it's not breaking (not that I think any dtest is using this option)

fruch:
@dkropachev please open PRs from forks

dkropachev:
> 1. not sure how one of the nodes has 0, I don't understand the logic here
> 2. we should run dtest gating to validate it's not breaking (not that I think any dtest is using this option)
> 
> **EDIT**: there a few test that do use it: https://github.com/search?q=repo%3Ascylladb%2Fscylla-dtest%20initial_token&type=code

Running: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dkropachev/job/dtest-gating/2

fruch (approved):
LGTM

PR #645, opened 2025-03-24
Prepare for ScyllaDB move to host ids instead of ips

gleb-cloudius:
Moving ScyllaDB to host ids changes some output to use ids instead of ips. Prepare for that by checking for both.

### Testing
- [x] 🟡 https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/894/ (Gossip - OSS)
- [x] :green_circle: https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/895/ (Raft - OSS)
- [x] :green_circle: https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/896/ (Tablets - OSS)

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -1449,6 +1449,10 @@ class ScyllaNode(Node):
     def hostid(self, timeout=60, force_refresh=False):
         if self.node_hostid and not force_refresh:
             return self.node_hostid
+        m = self.grep_log("init - Setting local host id to (.+?)$")
+        if m:
+            self.node_hostid = m[-1][1].group(1)
+            return self.node_hostid
         try:
             node_address = self.address()
             url = f"http://{node_address}:{self.api_port}/storage_service/hostid/local"
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -522,7 +522,7 @@ class Node(object):
         the log is watched from the beginning.
         """
         tofind = nodes if isinstance(nodes, list) else [nodes]
-        tofind = [f"{node.address()}.* now (dead|DOWN)" for node in tofind]
+        tofind = [f"({node.address()}|{node.hostid()}).* now (dead|DOWN)" for node in tofind]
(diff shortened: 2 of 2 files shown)


fruch:
@gleb-cloudius FYI after this would be review/merged, the sub-module in dtest would need to be updated.

also I've trigger all of the gating tests with it, to validate it's not breaking anything.

ccmlib/scylla_node.py
```diff
@@ -1449,6 +1449,10 @@ def rollback(self, upgrade_to_version):
     def hostid(self, timeout=60, force_refresh=False):
         if self.node_hostid and not force_refresh:
             return self.node_hostid
+        m = self.grep_log("init - Setting local host id to (.+?)$")
```
fruch, line 1452: this is for stages we don't have the API available yet ?

gleb-cloudius: Not only. Sometimes we want to get host id of a host that is already down.

gleb-cloudius:
Restarted the first one: https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/897/

fruch (approved):
LGTM

fruch:
> Restarted the first one: https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/897/

It's was a test known to be unstable, and the rerun passed

fruch:
> @gleb-cloudius FYI after this would be review/merged, the sub-module in dtest would need to be updated.
> 
> also I've trigger all of the gating tests with it, to validate it's not breaking anything.

Now we can do the dtest part

mykaul:
This needs to be backported to 2025.1, as it seems to be failing on it as well, I suspect.

fruch:
> This needs to be backported to 2025.1, as it seems to be failing on it as well, I suspect.

the dtest PR can be backported, once someone would open it up

PR #647, opened 2025-04-02
Modify Node::repair method for nodetool cluster repair

Deexie:
Since https://github.com/scylladb/scylladb/pull/22905 nodetool repair
repairs only vnode keyspaces. To repair tablet keyspaces, you need
to use nodetool cluster repair.

Add ScyllaNode::repair that uses the appropriate nodetool command
to repair specified tables. ScyllaNode::repair method gets repair
parameters as arguments. It creates objects of newly added
{Tablet,Vnode}RepairOptions classes to serialize the params.

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -281,6 +281,19 @@ class ScyllaCluster(Cluster):
 
         self._update_config(install_dir=self.nodelist()[0].node_install_dir)
 
+    def repair(self):
+        """
+        Repairs the whole cluster, both vnode (with nodetool repair) and tablet (with nodetool cluster repair) keyspaces
+        """
+        self.nodetool("repair")
+
+        stdout, _ = self.nodetool('help')
+        if "  cluster  " in stdout:
+            for node in list(self.nodes.values()):
+                if node.is_running():
+                    node.nodetool("cluster repair")
+                    break
+
 
 class ScyllaManager:
     def __init__(self, scylla_cluster, install_dir=None):
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -39,6 +39,7 @@ from ccmlib.utils.version import parse_version
 if TYPE_CHECKING:
     from ccmlib.scylla_cluster import ScyllaCluster
 
+
 class ScyllaType:
     """
     Helper class for defining Scylla type definitions using type strings.
@@ -1879,6 +1880,100 @@ class ScyllaNode(Node):
         if res.returncode:
             raise ToolError(command=' '.join(cmd_and_args), exit_status=res.returncode, stdout=res.stdout, stderr=res.stderr)
         return res.stdout.strip()
+
+    def _get_common_repair_options(self, keyspace: Optional[str] = None, tables: Optional[List[str]] = None, hosts: Optional[List[str]] = None, dcs: Optional[List[str]] = None):
+        """
+        Serializes repair options common for both vnode and tablet repair
+        """
+        options = []
+        if keyspace:
+            options.append(keyspace)
+            if tables:
+                options.append(" ".join(tables))
+
+        if hosts:
+            options.append("--in-hosts")
+            options.append(",".join(hosts))
+
+        if dcs:
+            options.append("--in-dc")
+            options.append(",".join(dcs))
+
+        return options
+
+    def vnode_repair(self, keyspace: Optional[str] = None, tables: Optional[List[str]] = None, hosts: Optional[List[str]] = None, dcs: Optional[List[str]] = None,
+                   end_token: Optional[int] = None, local: bool = False, partitioner_range: bool = False, start_token: Optional[int] = None, timeout=None):
+        """
+        Serializes repair options and runs nodetool repair
+        """
+        options = self._get_common_repair_options(keyspace, tables, hosts, dcs)
+
+        if end_token:
+            options.append("--end-token")
+            options.append(f"{end_token}")
+
+        if local:
+            options.append("--in-local-dc")
+
+        if partitioner_range:
+            options.append("--partitioner-range")
+
+        if start_token:
+            options.append("--start-token")
+            options.append(f"{start_token}")
+
+        return self.nodetool("repair " + " ".join(options), timeout=timeout)
+
+    def tablet_repair(self, keyspace: Optional[str] = None, tables: Optional[List[str]] = None, hosts: Optional[List[str]] = None, dcs: Optional[List[str]] = None,
+                           tablet_tokens: Optional[List[str]] = None, timeout=None):
+        """
+        Serializes repair options and runs nodetool cluster repair
+        """
+        options = self._get_common_repair_options(keyspace, tables, hosts, dcs)
+
+        if tablet_tokens:
+            options.append("--tablet-tokens")
+            options.append(",".join(tablet_tokens))
+
+        return self.nodetool("cluster repair " + " ".join(options), timeout=timeout)
+
+    def repair(self, keyspace: Optional[str] = None, tables: Optional[List[str]] = None, hosts: Optional[List[str]] = None, dcs: Optional[List[str]] = None,
+               end_token: Optional[int] = None, local: bool = False, partitioner_range: bool = False, start_token: Optional[int] = None,
+               tablet_tokens: Optional[List[str]] = None, timeout: Optional[int] = None):
+        """
+        Runs both vnode (nodetool repair) and tablet (nodetool cluster repair) repair with specified params
+        :param keyspace: the keyspace to repair
+        :param tables: tables to repair in the given keyspace
+        :param hosts: hosts participating in the repair
+        :param dcs: dcs participating in repair
+        :param end_token: token on which to end repair (for vnode repair only)
+        :param local: if true repair is run only on local dc (for vnode repair only)
+        :param partitioner_range: if true only the primary replicas are repaired (for vnode repair only)
+        :param start_token: token on which to begin repair (for vnode repair only)
+        :param tablet_tokens: tokens owned by the tablets to repair (for tablet repair only)
+        :param timeout: timeout of nodetool operations
+        """
+        outs_and_errs = []
+        if keyspace:
+            url = f"http://{self.address()}:{self.api_port}/storage_service/keyspaces"
+            resp = requests.get(url=url, params={"replication": "vnodes"})
+            resp.raise_for_status()
+
+            if keyspace in resp.json():
+                res = self.vnode_repair(keyspace, tables, hosts, dcs, end_token, local, partitioner_range, start_token, timeout=timeout)
+                outs_and_errs.append(res)
+            else:
+                res = self.tablet_repair(keyspace, tables, hosts, dcs, tablet_tokens, timeout=timeout)
+                outs_and_errs.append(res)
+        else:
+            res = self.vnode_repair(keyspace, tables, hosts, dcs, end_token, local, partitioner_range, start_token, timeout=timeout)
+            outs_and_errs.append(res)
+
+            res = self.tablet_repair(keyspace, tables, hosts, dcs, tablet_tokens, timeout=timeout)
+            outs_and_errs.append(res)
+
+        outs, errs = zip(*outs_and_errs)
+        return outs, errs
 
 
 class NodeUpgrader:


ccmlib/node.py
```diff
@@ -1418,11 +1539,69 @@ def drain(self, block_on_log=False):
...
+
+    def run_repair(self, options, **kwargs):
+        cmd = "repair " + options
+        return self.nodetool(cmd, **kwargs)
+
+    def run_cluster_repair(self, options, **kwargs):
+        cmd = "cluster repair " + options
         return self.nodetool(cmd, **kwargs)
 
+    def repair(self, options=[], replication: RepairOptions.KeyspaceReplication = RepairOptions.KeyspaceReplication.ALL, keyspace: Optional[str] = None,
```
fruch, line 1562: this is a bit weird, that means that if one uses `options` it won't do the cluster repair for tablets ?

ccmlib/cluster.py
```diff
@@ -616,6 +616,13 @@ def drain(self):
     def repair(self):
         self.nodetool("repair")
 
+        stdout, _ = self.nodetool('help')
+        if " cluster:" in stdout:
+            for node in list(self.nodes.values()):
```
fruch, line 621: isn't cluster repair, means I need todo it only once ?

i.e. what's the point of running it on all nodes ?

fruch: also this code now does repair node by node, and then cluster repair node by node, 

I'm not sure it's the intention

Deexie: There is `break` after the first `cluster repair` on a running node

ccmlib/node.py
Deexie: Yes, with options param we cannot correctly decide which keyspaces should be repaired

ccmlib/cluster.py
fruch: but again, the code does regular reapir on all nodes, and then cluser repair....

Deexie: Yes, nodetool repair will only repair vnode keyspaces; nodetool cluster repair - only tablet keyspaces

ccmlib/node.py
fruch: also what happens what there a mix case, that have both tablets and vnodes keyspaces ?
(not that I know of such tests in dtest, at least for now)

Deexie: Before the changes, `Node::repair` could have repaired each type of keyspaces. 
`node1.repair()` repaired both vnode and tablet keyspaces.
`node1.repair(["ks"])` repaired ks keyspace regardless of whether it was tablet- or vnode-based.

It would be the best to maintain this after the change, so that user does not need to check whether the cluster has vnode or tablet, or both keyspaces. Moreover, since we have a new command that wasn't there before, we should check if it is available in the version we are testing and if not then fallback to the old procedure. That's quite a lot to leave for a user.

The function looks complicated, but it just performs all the checks to ensure the right command is used. I agree that the function takes a lot of params, but in practice, most of them would get a default value and users have to specify only the params they are going to use. So instead of writing:
`node1.repair(["ks", "cf1", "cf2", "-dc", "dc1,dc3"])` (or `cluster_repair` whenever needed)
we may write:
`node1.repair(keyspace="ks", tables=["cf1", "cf2"], dcs=["dc1", "dc2"])`

The keyspace will be repaired with the right command regardless of its type and the version of Scylla. And what's best, is that I don't need to check what's the dc selector flag name, whether I should use the comma or the space or whatever as the separator.

The options param is still there to maintain the compatibility with the older dtest (I'd like to merge the changes in order: ccm -> dtest -> scylla), but I think it would be best to get rid of it as a follow up.

asias, line 1562: I think it is better to drop the raw option=[] support completely so that we support only the proper options list here.

ccmlib/node.py
```diff
@@ -1418,11 +1539,69 @@ def drain(self, block_on_log=False):
...
+                else:
+                    res = self.run_repair(repair_options.get_vnode_repair_options(), **kwargs)
+                    outs_and_errs.append(res)
+            else:
+                if self.cluster_command_available():
+                    if repair_options.repair_vnodes():
+                        res = self.run_repair(repair_options.get_vnode_repair_options(), **kwargs)
+                        outs_and_errs.append(res)
+
+                    if repair_options.repair_tablets() and self.cluster_command_available():
```
asias, line 1591: I think we do not need to support the replication parameter to limit the scope of the repair to tablet or vnode. This is not useful in a test.

ccmlib/node.py
```diff
@@ -1418,11 +1539,69 @@ def drain(self, block_on_log=False):
...
+               job_threads: Optional[int] = None, partitioner_range: bool = False, pull: bool = False, sequential: bool = False, start_token: Optional[int] = None,
+               trace: bool = False, tablet_tokens: Optional[list[str]] = None, **kwargs):
+        if not options:
+            repair_options = RepairOptions(replication, keyspace, tables, hosts, dcs, dc_parallel, end_token, full, local, ignore_unreplicated_keyspaces,
+                                           job_threads, partitioner_range, pull, sequential, start_token, trace, tablet_tokens)
+            outs_and_errs = []
+            if repair_options.keyspace:
+                url = f"http://{self.address()}:{self.api_port}/storage_service/keyspaces"
+                resp = requests.get(url=url, params={"replication": "vnodes"})
+
```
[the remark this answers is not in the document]
asias: If the table is a tablet table and the nodetool  cluster repair cmd is not supported, we'd better fail the test too.

ccmlib/node.py
```diff
@@ -1418,11 +1539,69 @@ def drain(self, block_on_log=False):
...
+    def run_cluster_repair(self, options, **kwargs):
+        cmd = "cluster repair " + options
         return self.nodetool(cmd, **kwargs)
 
+    def repair(self, options=[], replication: RepairOptions.KeyspaceReplication = RepairOptions.KeyspaceReplication.ALL, keyspace: Optional[str] = None,
+               tables: Optional[list[str]] = None, hosts: Optional[list[str]] = None, dcs: Optional[list[str]] = None, dc_parallel: bool = False,
+               end_token: Optional[int] = None, full: bool = False, local: bool = False, ignore_unreplicated_keyspaces: bool = False,
+               job_threads: Optional[int] = None, partitioner_range: bool = False, pull: bool = False, sequential: bool = False, start_token: Optional[int] = None,
+               trace: bool = False, tablet_tokens: Optional[list[str]] = None, **kwargs):
+        if not options:
```
asias, line 1567: Let's drop the extra options that might be passed by kwargs as raw repair options.

ccmlib/node.py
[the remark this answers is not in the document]
asias: It's like:

```
if keyspace is provieded
    if ks is tablet:
          use nodetool repair
    else
           use nodetool cluster repair
else if keyspace is not provided
       for all tablet ks
            use nodetool cluster repair
        for all vnode ks
               use nodetool repair
```

ccmlib/node.py
```diff
@@ -85,6 +88,124 @@ def __decode(self, value):
...
+            self.partitioner_range = partitioner_range
+            self.pull = pull
+            self.sequential = sequential
+            self.start_token = start_token
+            self.trace = trace
+
+        def get_options(self):
+            options = []
+            if self.dc_parallel:
+                options.append("--dc-parallel")
```
asias, line 128: Can be dropped.

ccmlib/node.py
```diff
@@ -85,6 +88,124 @@ def __decode(self, value):
...
+
+            if self.end_token:
+                options.append("--end-token")
+                options.append(f"{self.end_token}")
+
+            if self.full:
```
asias, line 134: Can be dropped.

ccmlib/node.py
```diff
@@ -85,6 +88,124 @@ def __decode(self, value):
...
+                options.append("--full")
+
+            if self.local:
+                options.append("--in-local-dc")
+
+            if self.ignore_unreplicated_keyspaces:
```
asias, line 140: Can be dropped.

ccmlib/node.py
```diff
@@ -85,6 +88,124 @@ def __decode(self, value):
...
+                options.append("--ignore-unreplicated-keyspaces")
+
+            if self.job_threads:
+                options.append("--job-threads")
+                options.append(f"{self.job_threads}")
```
asias, line 145: Can be dropped.

ccmlib/node.py
```diff
@@ -85,6 +88,124 @@ def __decode(self, value):
...
+
+            if self.partitioner_range:
+                options.append("--partitioner-range")
+
+            if self.pull:
+                options.append("--pull")
```
asias, line 151: Can be dropped.

ccmlib/node.py
```diff
@@ -85,6 +88,124 @@ def __decode(self, value):
...
+
+            if self.sequential:
+                options.append("--sequential")
```
asias, line 154: Can be dropped.

ccmlib/node.py
```diff
@@ -85,6 +88,124 @@ def __decode(self, value):
...
+
+            if self.start_token:
+                options.append("--start-token")
+                options.append(f"{self.start_token}")
+
+            if self.trace:
```
asias, line 160: Can be dropped.

ccmlib/node.py
```diff
@@ -85,6 +88,124 @@ def __decode(self, value):
 _active_tasks_pattern = re.compile(r'\s*([\w-]+)\s+\w+\s+(?P<ks>\w+)\s+(?P<cf>\w+)\s+\d+\s+\d+\s+\w+\s+\d+\.\d+%')
 
 
+class RepairOptions:
+    class KeyspaceReplication(Enum):
```
asias, line 92: Le'ts drop the KeyspaceReplication limiter. It is not very useful for the test but introduces a lot of confusing.

ccmlib/node.py
Deexie: I'll do it as a follow-up so that dtest does not fail after this change (and before its)

Deexie:
- drop dc_parallel, full, ignore_unreplicated_keyspaces, job_threads, pull, sequential, trace params
- drop KeyspaceReplication
- fail repair if storage_service/keyspaces fails
- fail if nodetool cluster repair isn't available

ccmlib/node.py
```diff
@@ -1418,11 +1494,53 @@ def drain(self, block_on_log=False):
         if block_on_log:
             self.watch_log_for("DRAINED", from_mark=mark)
 
-    def repair(self, options=[], **kwargs):
-        args = ["repair"] + options
-        cmd = ' '.join(args)
+    @property
+    def scylla_yaml(self) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
pehala: so now we have a duplicate code in node.py and scylla_node.py?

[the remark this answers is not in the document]
Deexie: Right, I'll drop it from scylla_node.py

ccmlib/node.py
```diff
@@ -1418,11 +1494,53 @@ def drain(self, block_on_log=False):
...
         return self.nodetool(cmd, **kwargs)
 
+    def repair(self, options=[], keyspace: Optional[str] = None,
+               tables: Optional[list[str]] = None, hosts: Optional[list[str]] = None, dcs: Optional[list[str]] = None, end_token: Optional[int] = None,
+               local: bool = False, partitioner_range: bool = False, start_token: Optional[int] = None, tablet_tokens: Optional[list[str]] = None, **kwargs):
+        if not options:
+            repair_options = RepairOptions(keyspace, tables, hosts, dcs, end_token, local, partitioner_range, start_token, tablet_tokens)
+            outs_and_errs = []
+            if repair_options.keyspace:
+                url = f"http://{self.address()}:{self.api_port}/storage_service/keyspaces"
```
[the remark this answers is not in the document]
Deexie:
```
build/lib/ccmlib/scylla_node.py:        bootstrap_message = r'storage_service .* Starting to bootstrap'
build/lib/ccmlib/scylla_node.py:            url = f"http://{node_address}:{self.api_port}/storage_service/hostid/local"
build/lib/ccmlib/scylla_node.py:            url_joining = f"http://{self.address()}:{self.api_port}/storage_service/nodes/joining"
build/lib/ccmlib/scylla_node.py:            url_tokens = f"http://{self.address()}:{self.api_port}/storage_service/tokens/"
build/lib/ccmlib/scylla_node.py:            url_host_ids = f"http://{self.address()}:{self.api_port}/storage_service/host_id"
ccmlib/scylla_node.py:        bootstrap_message = r'storage_service .* Starting to bootstrap'
ccmlib/scylla_node.py:            url = f"http://{node_address}:{self.api_port}/storage_service/hostid/local"
ccmlib/scylla_node.py:            url_joining = f"http://{self.address()}:{self.api_port}/storage_service/nodes/joining"
ccmlib/scylla_node.py:            url_tokens = f"http://{self.address()}:{self.api_port}/storage_service/tokens/"
ccmlib/scylla_node.py:            url_host_ids = f"http://{self.address()}:{self.api_port}/storage_service/host_id"
ccmlib/node.py:                url = f"http://{self.address()}:{self.api_port}/storage_service/keyspaces"
```
Looks like we don't

Deexie:
- get rid of cmd variable from run_repair and run_cluster_repair
- split VnodeRepairOptions and TabletRepairOptions
- s/list/List
- drop scylla_yaml and api_port from scylla_node
- drop options and kwargs params from Node::repair

ccmlib/scylla_node.py
```diff
@@ -162,14 +162,6 @@ def has_jmx(self):
         else:
             return os.path.isdir(os.path.join(install_dir, "tools", "jmx"))
 
-    @property
-    def scylla_yaml(self) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
fruch: why ? Cassandra doesn't have scylla.yaml and doesn't have api port... 

not sure why we want those properties there ?

ccmlib/node.py
```diff
@@ -1418,10 +1490,48 @@ def drain(self, block_on_log=False):
...
+    def run_repair(self, options, timeout=None):
+        return self.nodetool("repair " + options, timeout=timeout)
+
+    def run_cluster_repair(self, options, timeout=None):
+        return self.nodetool("cluster repair " + options, timeout=timeout)
+
+    def repair(self, keyspace: Optional[str] = None, tables: Optional[List[str]] = None, hosts: Optional[List[str]] = None, dcs: Optional[List[str]] = None,
+               end_token: Optional[int] = None, local: bool = False, partitioner_range: bool = False, start_token: Optional[int] = None,
+               tablet_tokens: Optional[List[str]] = None, timeout: Optional[int] = None):
+        outs_and_errs = []
```
fruch, line 1510: I think, all of this implementation should move to scylla_node.py

it doesn't have any sense being in this file/class.

ccmlib/cluster.py
```diff
@@ -616,6 +616,13 @@ def drain(self):
     def repair(self):
         self.nodetool("repair")
 
+        stdout, _ = self.nodetool('help')
```
fruch, line 619: also this whole logic should go into `ScyllaCluster.repair()`, it's doesn't have any relevancy for this file...

Deexie:
- move new repair methods to scylla_{node,cluster}.py

Deexie:
- modift commit message

Deexie:
@asias @fruch @pehala I addressed your comments. Please rereview.

All tests pass with current version, both with and without https://github.com/scylladb/scylladb/pull/23660.

ccmlib/scylla_node.py
```diff
@@ -1876,6 +1945,41 @@ def run_scylla_types(self, action: str, scylla_type: ScyllaType, *values: Tuple[
             raise ToolError(command=' '.join(cmd_and_args), exit_status=res.returncode, stdout=res.stdout, stderr=res.stderr)
         return res.stdout.strip()
 
+    def run_repair(self, options, timeout=None):
+        return self.nodetool("repair " + options, timeout=timeout)
+
+    def run_cluster_repair(self, options, timeout=None):
+        return self.nodetool("cluster repair " + options, timeout=timeout)
+
+    def repair(self, keyspace: Optional[str] = None, tables: Optional[List[str]] = None, hosts: Optional[List[str]] = None, dcs: Optional[List[str]] = None,
```
pehala, line 1954: I do not like naming of this, difference between `run_repair` and `repair` is not easily gueassable and it would be very confusing to me if I read without deeper knowledge of the system. This is not helped by the fact that there are no docstrings.

I would expect `repair` to do a nodetool repair directly, but it is not that and it incorporates both `repair` and `cluster_repair`. I would name it something else to signify it is not a direct call to the server but involves a bit of logic as well. `repair_all` might work?

pehala (changes requested):
Please provide docstrings to the new methods.

ccmlib/scylla_node.py
```diff
@@ -39,6 +39,75 @@
 if TYPE_CHECKING:
     from ccmlib.scylla_cluster import ScyllaCluster
 
+class RepairOptions:
```
pehala, line 42: I am not sold on these classes, they are internal only and are used only once in one method. Do we really need them? Cannot we design the method better, or maybe split it in two so we dont need them at all?

pehala: Another idea would be to have `run_repair` and `run_cluster_repair` use these classes as well, which would simplify the `repair` method as well

Deexie:
- add comments
- delete RepairOptions classes and serialize params directly in run_repair/run_cluster_repair

Deexie:
@pehala please rereview

ccmlib/scylla_node.py
pehala: Is it also how the customers are gonna use it? I.e. are they expected to run both repairs?

Deexie: Yes, they do

Deexie:
- drop options param from _run_cluster_repair and _run_repair

Deexie:
- rename _run_repair to vnode_repair and _run_cluster_repair to tablet_repair

pehala (approved):

Deexie:
@scylladb/ccm-maint The changes are approved and dtests pass with this changes. Please merge.

ccmlib/scylla_cluster.py
```diff
@@ -281,6 +281,19 @@ def upgrade_cluster(self, upgrade_version):
...
+        Repairs the whole cluster, both vnode (with nodetool repair) and tablet (with nodetool cluster repair) keyspaces
+        """
+        self.nodetool("repair")
+
+        stdout, _ = self.nodetool('help')
+        if "  cluster  " in stdout:
+            for node in list(self.nodes.values()):
+                if node.is_running():
+                    node.nodetool("cluster repair")
+                    break
```
dkropachev, line 295: Are you sure that we want it to silently skip if nodetool does not support it ?

dkropachev (changes requested):

dkropachev (approved):

Deexie:

roydahan:
@fruch have you approved after your review comments were addressed?

fruch (approved):
LGTM

PR #648, opened 2025-04-14
fix(scyll_node.start): remove unsupported parameter for old version

aleksbykov:
New parameter --rf-rack-valid-keyspaces cause scylla start failing for node with scylla versions 6.2, 2024.x. This cause upgrade tests from these version to failed at the begining.
The parameter --rf-rack-valid-keyspaces mostly passed via env parameter SCYLLA_EXT_OPTS for dtest runs.
Remove such parameter from the env variable before node start if scylla version on node is <2025.1


aleksbykov:
@scylladb/dtest-maintainers , review please

fruch (approved):
LGTM

fruch:
@aleksbykov 

you'll need to update the ccm submodule on dtest now...

Issue #650, opened 2025-04-17
Adding node without dc leads to broken `cassandra-rackdc.properties`


dkropachev:
Steps:
```
ccm create --scylla -n 1:0 -i 127.0.0. -v release:2025.1 -b test-ccm-bug
ccm start --wait-other-notice --wait-for-binary-proto --jvm_arg=--skip-wait-for-gossip-to-settle=0
ccm add -b -i 127.0.0.2 -j 7200 -r 2200 --scylla node2
ccm node2 start --wait-other-notice --wait-for-binary-proto --jvm_arg=--skip-wait-for-gossip-to-settle=0
```
Ends up at error:
```
Traceback (most recent call last):
  File "~/.venv/bin/ccm", line 74, in <module>
    cmd.run()
  File "~/.venv/lib/python3.11/site-packages/ccmlib/cmds/node_cmds.py", line 196, in run
    self.node.start(not self.options.no_join_ring,
  File "~/scylla-ccm/.venv/lib/python3.11/site-packages/ccmlib/scylla_node.py", line 762, in start
    scylla_process = self._start_scylla(args=args, marks=marks, update_pid=update_pid,
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/.venv/lib/python3.11/site-packages/ccmlib/scylla_node.py", line 405, in _start_scylla
    self.wait_for_binary_interface(from_mark=from_mark, process=self._process_scylla, timeout=t)
  File "~/.venv/lib/python3.11/site-packages/ccmlib/node.py", line 546, in wait_for_binary_interface
    self.watch_log_for("Starting listening for CQL clients", **kwargs)
  File "~/.venv/lib/python3.11/site-packages/ccmlib/node.py", line 512, in watch_log_for
    raise RuntimeError(f"The process is dead, returncode={process.returncode}")
RuntimeError: The process is dead, returncode=1
```

Node2 start fails with:
```
ERROR 2025-04-17 08:59:28,804 [shard 0:main] snitch_logger - Property file ~/.ccm/cpp-driver_2025-1-0_1-0/node2/conf/cassandra-rackdc.properties is incomplete. Some obligatory fields are missing.
[...]

PR #653, opened 2025-04-21
Fix multi dc provisioning

dkropachev:
Multi DC/rack provisioning is broken in multiple cases:
1. Rack for node is not stored
2. Manual provisioning is broken (create, add, add, ..., start)
3. Start, add, node start is broken
Fixes: https://github.com/scylladb/scylla-ccm/issues/650


dkropachev:
@fruch , there are two debatable changes:
1. Changing default dc pattern to `datacenter{id}`. It can lead to problems in test because dc is going to change, but makes dc naming consistent.
2. Adding `is_single_dc_cluster` flag to cluster. It is needed for cluster to know snitch for nodes they were provisioned with, in `multi dc` cluster it is required to use `GossipingPropertyFileSnitch`, while in `single dc`  `SimpleSnitch` is used. All nodes in cluster should have same snitch, otherwise node joining to the cluster will fail. Another way to solve it by defaulting to `GossipingPropertyFileSnitch`.

ccmlib/common.py
```diff
@@ -52,6 +52,9 @@
 
 DOWNLOAD_IN_PROGRESS_FILE = "download_in_progress"
 
+DEFAULT_DATACENTER = 'datacenter1'
+DEFAULT_RACK = 'rack1'
```
fruch, line 56: this is breaking checks for dtest....

you would need to run all test, and prove there's no dependency

fruch:
> @fruch , there are two debatable changes:
> 
> 1. Changing default dc pattern to `datacenter{id}`. It can lead to problems in test because dc is going to change, but makes dc naming consistent.

this is a breaking change...  to do it, one would need to prove it's not break any user of CCM (i.e. dtest, all drivers)

> 2. Adding `is_single_dc_cluster` flag to cluster. It is needed for cluster to know snitch for nodes they were provisioned with, in `multi dc` cluster it is required to use `GossipingPropertyFileSnitch`, while in `single dc`  `SimpleSnitch` is used. All nodes in cluster should have same snitch, otherwise node joining to the cluster will fail. Another way to solve it by defaulting to `GossipingPropertyFileSnitch`.

I don't mind changing the default to GossipingPropertyFileSnitch, but it should be synced with Core, i.e. they might want to make it default as well ?

PR #655, opened 2025-04-22
fix: Use correct script, revert version

pehala:
* Version was needlessly changed introduced in 39f4522ccdce95f18b6ef7004bd2b68442f65201, revert back to 2.0.5
* CCM script path was broken, setuptools does not like scripts outside the packages.
   * Move logic into ccmlib.bin and leave ccm script to only point to that package.


fruch (approved):
LGTM

fruch:
@pehala can you add as a followup a test to run the entry-point, to validate it's working ?

PR #656, opened 2025-04-22
Store node rack in node.conf

dkropachev:
Rack is not preserved in node.conf.
As result, next call of `__update_topology_files` overwrites rack information and all nodes dropped to rack1


fruch (approved):
LGTM

PR #657, opened 2025-04-22
Start old cassandra and scylla nodes one by one

dkropachev:
Cassandra 3.0 and Scylla older than 6.x does not support parallel node initialization. 
Tune CCM to start nodes one by one when server does not support parallel node initialization.


fruch (changes requested):
this logic should be in whom use CCM
those parameters are not really related to parallel execution, they are they for testing purposes to error paths. 

I don't think this logic should be in CCM

dkropachev:
> this logic should be in whom use CCM those parameters are not really related to parallel execution, they are they for testing purposes to error paths.
> 
> I don't think this logic should be in CCM

Added `--parallel-start` option, if not set, default is used, which is calculated at `Cluster.parallel_start_supported`

ccmlib/cluster.py
```diff
@@ -80,7 +80,18 @@ def __init__(self, path, name, partitioner=None, install_dir=None, create_direct
             if create_directory:
                 common.rmdirs(self.get_path())
             raise
-        self.debug(f"Started cluster '{self.name}' version {self.__version} installed in {self.__install_dir}")
+        self.debug(f"Started cluster '{self.name}' version {self._version} installed in {self.__install_dir}")
+
+    @property
+    def parallel_start_supported(self):
+        return self.database_version.startswith('3.')
```
fruch, line 87: 4.x doesn't support it ?

ccmlib/cluster.py
```diff
@@ -470,23 +481,27 @@ def show(self, verbose):
             else:
                 node.show(only_status=True)
 
-    def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False, wait_other_notice=False, jvm_args=None, profile_options=None, quiet_start=False):
+    def start(self, no_wait=False, verbose=False, wait_for_binary_proto=False, wait_other_notice=False, jvm_args=None,
+              profile_options=None, quiet_start=False, parallel_start=None):
         if jvm_args is None:
             jvm_args = []
+        if parallel_start is None:
+            parallel_start = self.parallel_start_supported
```
fruch, line 489: this should be explicit, not like that.

this gonna break all of dtest

fruch:
@dkropachev what is the context of this one, why are you trying to change the default behavior, for what end exactly ? 

just cause you don't like the `wait_for_*` parameter, of for a specific test (or test configuration) you can't do without this change ?

ccmlib/cluster.py
dkropachev: why ? we already start scylla nodes in parallel

dkropachev:
> @dkropachev what is the context of this one, why are you trying to change the default behavior, for what end exactly ?
> 
> just cause you don't like the `wait_for_*` parameter, of for a specific test (or test configuration) you can't do without this change ?

Context is that I have tried to start cassandra cluster with many nodes and couple of times it failed with token ring collission, due the fact that we are starting them in parallel.

fruch:
> > @dkropachev what is the context of this one, why are you trying to change the default behavior, for what end exactly ?
> > 
> > just cause you don't like the `wait_for_*` parameter, of for a specific test (or test configuration) you can't do without this change ?
> 
> Context is that I have tried to start cassandra cluster with many nodes and couple of times it failed with token ring collission, due the fact that we are starting them in parallel.

So pass the wait_for parameters and you are fine. Why do you need automatically do that decision?

Issue #646, opened 2025-03-31
Stopping and then starting the node fails for Cassandra


muzarski:
Branch: master

I believe this might be due to https://github.com/scylladb/scylla-ccm/pull/645 - when C* node is down, and we call `ccm node start`, the `hostid` method is called on a node that is down. I saw that `ScyllaNode::hostid` logic was adjusted, so it's possible to obtain a host id from Scylla node that is down.

Reproducer from cpp-rust-driver CI:
```
ccm> ccm create -n 3:0 -i 127.0.0. -v 3.11.19 -b cpp-driver_3-11-19_3-0
ccm> ccm updateconf --rt=10000 read_request_timeout_in_ms:10000 write_request_timeout_in_ms:10000 request_timeout_in_ms:10000 phi_convict_threshold:16 hinted_handoff_enabled:false enable_materialized_views:true dynamic_snitch_update_interval_in_ms:1000 native_transport_max_threads:1 concurrent_reads:2 concurrent_writes:2 concurrent_compactors:1 compaction_throughput_mb_per_sec:0 key_cache_size_in_mb:0 key_cache_save_period:0 memtable_flush_writers:1 max_hints_delivery_threads:1 cas_contention_timeout_in_ms:10000 file_cache_size_in_mb:0 rpc_min_threads:1 rpc_max_threads:1
ccm> ccm start --wait-other-notice --wait-for-binary-proto
ccm> ccm status
ccm> ccm node2 stop
ccm> ccm node3 stop
ccm> ccm node2 start --wait-other-notice --wait-for-binary-proto
unknown file: Failure
C++ exception with description "Traceback (most recent call last):
  File "/home/runner/.local/bin/ccm", line 74, in <module>
    cmd.run()
  File "/home/runner/.local/lib/python3.11/site-packages/ccmlib/cmds/node_cmds.py", line 196, in run
    self.node.start(not self.options.no_join_ring,
  File "/home/runner/.local/lib/python3.11/site-packages/ccmlib/node.py", line 675, in start
    node.watch_log_for_alive(self, from_mark=mark)
  File "/home/runner/.local/lib/python3.11/site-packages/ccmlib/node.py", line 534, in watch_log_for_alive
    tofind = [f"({node.address()}|{node.hostid()}).* now UP" for node in tofind]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #658, opened 2025-04-22
Do not read host-id on watch_log_for_alive

dkropachev:
`host-id` requires API to be up and running, which is not always the case right away.
Ignoring it there will help to reduce chance of having host-id reading error while node is starting.

Fixes: https://github.com/scylladb/scylla-ccm/issues/646

### DIFF
--- a/ccmlib/node.py
+++ b/ccmlib/node.py
@@ -106,6 +106,7 @@ class Node(object):
             is almost always the right choice.
         """
         self.name = name
+        self.node_hostid = None
         self.cluster: 'ScyllaCluster' = cluster
         self.status = Status.UNINITIALIZED
         self.auto_bootstrap = auto_bootstrap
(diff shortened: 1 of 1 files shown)


tests/tests/test_results/ccm.log
fruch, line 1: please remove this log

ccmlib/node.py
```diff
@@ -530,9 +530,19 @@ def watch_log_for_alive(self, nodes, from_mark=None, timeout=120, filename='syst
...
+        patterns = []
+        for node in nodes if isinstance(nodes, list) else [nodes]:
+            host_id = None
+            try:
+                host_id = node.hostid()
+            except Exception:
+                pass
+            if not host_id:
+                # If failed to get host_id, replate it with a pattern,
+                # Address is enough for node identification
```
gleb-cloudius, line 542: What does it mean? There is no longer address reported in the "now UP" message so as far as I see this will simply match UP message for any node.

dkropachev: Since when ? I see the following on 2025.1:
```
INFO  2025-04-24 01:02:07,908 [shard 0: gms] gossip - InetAddress 62671905-d3b4-4189-bedf-cd2f02926bea/127.0.0.2 is now UP, status = UNKNOWN
```

I have modified the code to look for ip-address and node-id when node-id is available, otherwise only ip-address.

dkropachev (commented):
I have added more changes to takle the problem, now it should be fast and stable.

tests/tests/test_results/ccm.log
dkropachev: done

gleb-cloudius (commented):
Is the merge commit intentional?

ccmlib/node.py
gleb-cloudius: Ah, it was dropped for DOWN message, but not for UP. Strange.

dkropachev:
> Is the merge commit intentional?

Yes, it just a way github UI does rebase on master, let me remove it.

gleb-cloudius (approved):

fruch (approved):
LGTM

even that I think thing host_id should be referenced only on scylla cluster and nodes

PR #661, opened 2025-05-08
ccmlib/scylla_node.py: run_scylla_sstable(): dump sstables from a snapshot

denesb:
When running scylla-sstable, sstables passed to the tool are prone to disappearing as a result of a concurrent compaction. To prevent this, callers are required to disable compaction on the table from which sstables are dumped. With tablets, new ways for sstables to disappear were introduced: migration, split and merge. Instead of piling on more responsibilities on the caller, ensure that sstables are stable during the dump internally: create a snapshot and dump the sstables from the snapshot. The snapshot is cleared after the dump, or left in place in case of error (for post-mortem analysis).

If the caller doesn't provide a keyspace, or they provide a specific list of datafiles to be dumped, they still have to ensure that they are stable.

After this change, the caller is still expected to keep schema tables non-compacting. This will be addressed in a future change.

Fixes: scylladb/scylla-dtest#5866

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -17,6 +17,7 @@ from pathlib import Path
 from collections import OrderedDict
 from typing import Any, Dict, Iterable, List, Optional, Tuple, Union, TYPE_CHECKING
 from enum import Enum
+import uuid
 
 import logging
 
@@ -1564,6 +1565,14 @@ class ScyllaNode(Node):
             Batch-mode can be only used if column_families contains a single item.
         * text - If True, output of the command is treated as text, if not as bytes
 
+        If datafiles is provided, the caller is responsible for making sure these
+        files are not removed by ScyllaDB while the tool is running.
+        If datafiles = None, this function will create a snapshot and dump
+        sstables from the snapshot to ensure that the sstables are not removed
+        while the tools is running.
+        The snapshot is removed after the dump completed, but it is left there in
+        case of error, for post-mortem analysis.
+
         Returns: map: {sstable: (stdout, stderr)} of all invokations. When batch == True, a single entry will be present, with empty key.
 
         Raises: subprocess.CalledProcessError if scylla-sstable returns a non-zero exit code.
@@ -1572,8 +1581,21 @@ class ScyllaNode(Node):
             additional_args = []
 
         scylla_path = common.join_bin(self.get_path(), BIN_DIR, 'scylla')
-        sstables = self._Node__gather_sstables(datafiles, keyspace, column_families)
         ret = {}
+
+        if datafiles is None and keyspace is not None:
+            tag = "sstable-dump-{}".format(uuid.uuid1())
+            kts = ",".join(f"{keyspace}.{column_family}" for column_family in column_families)
+            self.debug(f"run_scylla_sstable(): creating snapshot with tag {tag} to be used for sstable dumping")
+            self.nodetool(f"snapshot -t {tag} {kts}")
+            sstables = []
+            for column_family in column_families:
+                sstables.extend(glob.glob(os.path.join(self.get_path(), 'data', keyspace, f"{column_family}-*/snapshots/{tag}/*-Data.db")))
+        else:
+            sstables = self._Node__gather_sstables(datafiles, keyspace, column_families)
+            tag = None
+
+        self.debug(f"run_scylla_sstable(): preparing to dump sstables {sstables}")
 
         def do_invoke(sstables):
             # there are chances that the table is not replicated on this node,
@@ -1602,6 +1624,13 @@ class ScyllaNode(Node):
         else:
             for sst in sstables:
                 ret[sst] = do_invoke([sst])
+
+        # Deliberately not putting this in a `finally` block, if the command
+        # above failed, leave the snapshot with the sstables around for
+        # post-mortem analysis.
+        if tag is not None:
+            self.nodetool(f"clearsnapshot -t {tag} {keyspace}")
+
         return ret
 
     def dump_sstables(self,


mykaul:
@denesb - thanks for the quick fix. Do you have any idea why (1) we see it just now and (2) consistently failing with it?

denesb:
> @denesb - thanks for the quick fix. Do you have any idea why (1) we see it just now and (2) consistently failing with it?

I don't know, must be related to some change in timing. FYI it doesn't fail for me locally.

regevran (approved):
I looked it, LGTM.

ccmlib/scylla_node.py
```diff
@@ -1571,9 +1580,22 @@ def run_scylla_sstable(self, command, additional_args=None, keyspace=None, dataf
...
         ret = {}
 
+        if datafiles is None and keyspace is not None:
+            tag = "sstable-dump-{}".format(uuid.uuid1())
+            kts = ",".join([f"{keyspace}.{column_family}" for column_family in column_families])
+            self.debug(f"run_scylla_sstable(): creating snapshot with tag {tag} to be used for sstable dumping")
+            self.nodetool(f"snapshot -t {tag} {kts}")
+            sstables = []
+            for column_family in column_families:
+                sstables.extend(glob.glob(os.path.join(self.get_path(), 'data', keyspace, f"{column_family}-*/snapshots/{tag}/*-Data.db")))
```
pehala, line 1592: You could use Pathlib to do this, it has some differences compared to glob (https://docs.python.org/3/library/pathlib.html#comparison-to-the-glob-module) though, not sure if they are relevant here
```
path = self.get_path() / "data" / "keyspace
path.glob(f"{column_family}-*/snapshots/{tag}/*-Data.db")
```

ccmlib/scylla_node.py
```diff
@@ -1571,9 +1580,22 @@ def run_scylla_sstable(self, command, additional_args=None, keyspace=None, dataf
...
+        else:
+            sstables = self._Node__gather_sstables(datafiles, keyspace, column_families)
```
pehala, line 1594: Out of scope for this PR, possibly another enhancement, but if we expect this function to be called somewhere else, it should not be private.

ccmlib/scylla_node.py
```diff
@@ -1571,9 +1580,22 @@ def run_scylla_sstable(self, command, additional_args=None, keyspace=None, dataf
             additional_args = []
 
         scylla_path = common.join_bin(self.get_path(), BIN_DIR, 'scylla')
-        sstables = self._Node__gather_sstables(datafiles, keyspace, column_families)
         ret = {}
 
+        if datafiles is None and keyspace is not None:
+            tag = "sstable-dump-{}".format(uuid.uuid1())
+            kts = ",".join([f"{keyspace}.{column_family}" for column_family in column_families])
```
pehala, line 1587: No need to materialize list
```suggestion
            kts = ",".join(f"{keyspace}.{column_family}" for column_family in column_families)
```

pehala:
> I don't know, must be related to some change in timing. FYI it doesn't fail for me locally.

Do we have some run which would demonstrate it fixes the issue, or are we running blind and hoping this will fix it?

ccmlib/scylla_node.py
denesb: Is there an advantage of using `Pathlib` instead of `glob` for this?

ccmlib/scylla_node.py
denesb: I want to retire this `else:` branch entirely (remove it). But for that I need to survey the usages of `run_scylla_sstable()` and I didn't want to delay this fix. I will do it in a follow-up.

denesb:
> > I don't know, must be related to some change in timing. FYI it doesn't fail for me locally.
> 
> Do we have some run which would demonstrate it fixes the issue, or are we running blind and hoping this will fix it?

The failure doesn't reproduces locally. That said we have a good understanding of what the problem is and this patch should fix it. But yeah, it is a blind fix.

denesb:
New in v2: don't materialize the list

ccmlib/scylla_node.py
denesb: Done in v2.

ccmlib/scylla_node.py
pehala: Functionally no, I just consider it to be a better API as it is more concise and object oriented, so it is just a personal preference. Feel free to keep it if you like this better.

ccmlib/scylla_node.py
denesb: To elaborate that a bit: this interface was inherited from the legacy `node.run_sstable2json()`, we wanted a drop-in replacement.
We now know that this is an unsafe interface, because even after this PR, one can still write flaky dtests using this `run_scylla_sstable()`, with certain parameter combinations.
I want to change the API such that it will not be possible to write tests that will fail with disappearing sstables anymore.

ccmlib/scylla_node.py
denesb: I'd prefer to keep `glob`, it is the one I'm familiar with.

pehala (approved):

denesb:
@scylladb/qa-maintainers ping for merge.

denesb:
Who can merge this? @fruch?

fruch (approved):
LGTM

fruch:
@denesb 

we are trying to take latest ccm into dtest, and this is breaking some of the gating tests:

https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/959/

I'm gonna revert this one from CCM, and you'll need to re-introduce it and test gating with it.

PR #663, opened 2025-05-12
node_cmds.py: Remove cli command from the list

Lorak-mmk:
This command implementation was previosuly removed in commit [06a2adf303c292239b9a84b6a674824e0204797c](https://github.com/scylladb/scylla-ccm/commit/06a2adf303c292239b9a84b6a674824e0204797c)
Unfortunately this commit did not remove the command from the command list, which caused "--help" to error out, and not print some other commands.

I also added a test to prevent such errors in the future.

### DIFF
--- a/ccmlib/cmds/node_cmds.py
+++ b/ccmlib/cmds/node_cmds.py
@@ -33,7 +33,6 @@ def node_cmds():
         "updateconf",
         "updatelog4j",
         "stress",
-        "cli",
         "cqlsh",
         "scrub",
         "verify",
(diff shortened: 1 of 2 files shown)


dkropachev (approved):

Lorak-mmk:
This failure doesn't look like something that this change could cause. Is it a known flaky test?

dkropachev:
> This failure doesn't look like something that this change could cause. Is it a known flaky test?

To answer this question we need to start collecting cluster logs from CICD

PR #666, opened 2025-05-26
scylla_cluster: access a node through ScyllaCluster::nodelist

Deexie:
Currently, ScyllaCluster::repair accesses the first of nodes in a cluster
by calling nodes[0]. However nodes is a dict, keys of which aren't number.

Replace nodes with nodelist() that returns the list of nodes.


Deexie:
> _ No description provided. _



> @Deexie can you add something to the description to explain, where its coming from ? how it working or didn't work so far...

Done @fruch

PR #667, opened 2025-05-27
scylla_repository: cache setups

fruch:
so we won't hit s3 urls on each start of a test (cause of multiple calls to `scylla_repository.setup()`)

we gonna cache this call for the time of the test session

on a new test session, those would be called only once

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


mykaul (approved):
Thanks, didn't know '@cache' is the same as '@lru_cache'

dkropachev:
> > Thanks, didn't know '@cache' is the same as '@lru_cache'
> 
> It's been so only since python 3.9
> 
> And the CI test is still also for 3.8 :(

If you change it to `@lru_cache` it should work on older python

PR #670, opened 2025-06-10
Revert "ccmlib/scylla_node.py: run_scylla_sstable(): dump sstables from a snapshot"

fruch:
Reverts scylladb/scylla-ccm#661

this change seems to be breaking some of the dtest gating

https://jenkins.scylladb.com/job/scylla-staging/job/dtest-pytest-gating/959/

we failed to test it before merging it, so now it's need to be reverted

### DIFF
--- a/ccmlib/scylla_node.py
+++ b/ccmlib/scylla_node.py
@@ -17,7 +17,6 @@ from pathlib import Path
 from collections import OrderedDict
 from typing import Any, Dict, Iterable, List, Optional, Tuple, Union, TYPE_CHECKING
 from enum import Enum
-import uuid
 
 import logging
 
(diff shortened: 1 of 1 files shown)


denesb:
I though we run gating tests when merging CCM PRs.
Also, how come these were not failing during regular dtest gating test runs during CI and promotion? Was this commit never promoted to dtest's master?

fruch:
> I though we run gating tests when merging CCM PRs. Also, how come these were not failing during regular dtest gating test runs during CI and promotion? Was this commit never promoted to dtest's master?

This commit never reached dtest, i.e. someone needed to update the submodule 
ccm is not part of next anymore.

The gating runs might have wrong run with a different ccm branch
We'll need to cross check

PR #673, opened 2025-06-24
ccmlib/scylla_node.py: Enable RF-rack-valid keyspaces by default

dawmd:
Since all dtests have been adjusted to work with the configuration option `rf_rack_valid_keyspaces` enabled, let's turn on the option by default.

Refs scylladb/scylladb#23071
Refs scylladb/scylla-dtest#5633


fruch:
> Right. I clearly haven't fully woken up yet. :-) I'll close this PR for now. Thanks for the fast feedback!

One more reason not todo it in CCM 

There are other users of it in driver tests, which aren't ready for doing this as default

PR #676, opened 2025-07-17
ccmlib/scylla_node.py: run_scylla_sstable(): dump sstables from a snapshot

denesb:
When running scylla-sstable, sstables passed to the tool are prone to disappearing as a result of a concurrent compaction. To prevent this, callers are required to disable compaction on the table from which sstables are dumped. With tablets, new ways for sstables to disappear were introduced: migration, split and merge. Instead of piling on more responsibilities on the caller, ensure that sstables are stable during the dump internally: create a snapshot and dump the sstables from the snapshot. The snapshot is cleared after the dump, or left in place in case of error (for post-mortem analysis).

If the caller doesn't provide a keyspace, or they provide a specific list of datafiles to be dumped, they still have to ensure that they are stable.

After this change, the caller is still expected to keep schema tables non-compacting. This will be addressed in a future change.

Fixes: scylladb/scylla-dtest#5866


fruch (approved):
LGTM

PR #677, opened 2025-08-19
fix(scylla_cluster): improve version str extraction for Manager

mikliapko:
stdout returned by `sctool version` contains two versions and looks like `Client version: 3.6.0\nServer version: 3.6.0\n`. 

In Manager, server and client version are always the same, thus, stdout can be splitlined and the first part taken and passed to `ComparableScyllaVersion()`.

In current implemented the input like `3.6.0\nServer version: 3.6.0\n` is set, which `ComparableScyllaVersion()` somehow manages to handle but which is not technically correct.

**Testing:**
- [ ]

### DIFF
--- a/ccmlib/scylla_cluster.py
+++ b/ccmlib/scylla_cluster.py
@@ -312,7 +312,9 @@ class ScyllaManager:
     @property
     def version(self):
         stdout, _ = self.sctool(["version"], ignore_exit_status=True)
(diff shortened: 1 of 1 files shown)


mikliapko:
> @mikliapko this is still needed ?

Yes, in general, it doesn't make any tests to fail (it was discovered during debugging) but can be improved to avoid potential issues.

Issue #681, opened 2025-11-12
Starting scylla with SSL fails


adespawn:
Starting a scyllaDB cluster, with the following set of commands:

```
ccm create cluster_name --scylla -v release:2025.3 --ssl ./scylla-ccm/ssl
ccm populate -n 1:0
ccm start --wait-for-binary-proto
```

Fails with the following error: 

```
Traceback (most recent call last):
  File "/usr/local/bin/ccm", line 4, in <module>
    __import__('pkg_resources').run_script('ccm==2.0.5', 'ccm')
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 722, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 1561, in run_script
    exec(code, namespace, namespace)
  File "/usr/local/lib/python3.12/dist-packages/ccm-2.0.5-py3.12.egg/EGG-INFO/scripts/ccm", line 74, in <module>
    cmd.run()
  File "/usr/local/lib/python3.12/dist-packages/ccm-2.0.5-py3.12.egg/ccmlib/cmds/cluster_cmds.py", line 676, in run
    if self.cluster.start(no_wait=self.options.no_wait,
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/ccm-2.0.5-py3.12.egg/ccmlib/scylla_cluster.py", line 175, in start
    started = self.start_nodes(**kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/ccm-2.0.5-py3.12.egg/ccmlib/scylla_cluster.py", line 140, in start_nodes
    p = node.start(update_pid=False, jvm_args=jvm_args,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/ccm-2.0.5-py3.12.egg/ccmlib/scylla_node.py", line 762, in start
    scylla_process = self._start_scylla(args=args, marks=marks, update_pid=update_pid,
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #682, opened 2025-11-12
Add scylla support of using --ssl from cli

fruch:
this was never implemented, and was mostly used from dtest that had the logic to set it up, and generate new key everytime

this change follow the cassandra path, and expectd specific files as in the ./ssl folder, and doesn't generate new keys.

Fix: #681


ssl/ccm_node.crl
```diff
@@ -0,0 +1,15 @@
...
+7ZqiFw0yNTEwMjMxMzM1MTJaoA4wDDAKBgNVHRQEAwIBATANBgkqhkiG9w0BAQsF
+AAOCAYEATPEFy0e1NB0heiFf4q/Y07hzfXZWFj6ouIegc1JqHWV7zcKw3zl+88cI
+8JRaeePj+CdyYAeDLHaIo5GqM+WPnWcFNtWLXiEpmehz1EvWyPZWoW3YQ/mt6p3R
+wP4pRM3G7oSX+KyholurbC1vpUZQu6pR4e+6PRy2AKtBSpXTzY4fHXlh6WPsMkfE
+LyVqN/2vAl/I33VGsl2ccNrIOV3aSATWtCNEhLwef75zIlLj4eSQMCCDJEk70NS0
+ofJMbEAErRQNeCBoxGTI8/k4+pAFHukU0SCjN01rvHWo+l6xirmWzcfPBQFp/zEj
+6K15GPcapSldbkD3G5OOKfN+ifN6H1y7HuqTfoJuQkhWpZNwDdOvL9Ut2NeWaP5p
+o5Xx0xAcAU8RR4fnZDZO+sJz8Rhc7MHarigqfpkRVoYzy9JlJaViBJrRxVBtvaKb
+m5VbE6xdWGDGvrm6NuWSmXTWIpABv7kN732xcjWjThFmmXwDBmOveUpZf6kBtRux
+ExfGZm5V
```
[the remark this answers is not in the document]
fruch: no need

ccmlib/scylla_cluster.py
```diff
@@ -221,6 +221,48 @@ def get_scylla_mode(self):
...
+        shutil.copyfile(os.path.join(ssl_path, 'ccm_node.key'), os.path.join(self.get_path(), 'ccm_node.key'))
+        ssl_options = {
+            'enabled': True,
+            'certificate': os.path.join(self.get_path(), "ccm_node.pem"),
+            'keyfile': os.path.join(self.get_path(), "ccm_node.key")
+        }
+
+        if require_client_auth:
+            cer_src = os.path.join(ssl_path, 'ccm_node.cer')
+            cer_dst = os.path.join(self.get_path(), "ccm_node.cer")
```
adespawn: Nit: shouldn't there be consistency between usage of `"` and `'`?

fruch: Cause I'm used to auto formatters, that we don't have by default in this repo

PR #727, opened 2026-03-23
ccmlib: fix wait_for_compactions premature return on zero active tasks

denesb:
Rename idle_timeout to timeout (it only guards against stuck compactions, not idle ones). Add quiesce_timeout (default 1s): the function now requires zero active tasks to be sustained for this long before returning, preventing a false 'done' signal during the scheduling gap between consecutive LCS compaction tasks. Poll at 100ms instead of 1s for faster response.

Fixes: SCYLLADB-1109


bhalevy (approved):

fruch (approved):
LGTM

PR #732, opened 2026-04-05
scylla_repository: fix cache invalidation for env var packages

avikivity:
Packages provided via environment variables (e.g. SCYLLA_UNIFIED_PACKAGE) were only read inside download_packages(), after the hash-based cache validation block. Since packages was None at validation time, a stale cached install was reused forever even when the local tarball changed.

Read env var packages earlier in setup() so they participate in hash validation, and extract the reading into a packages_from_env() helper to avoid duplication.

### DIFF
--- a/ccmlib/scylla_repository.py
+++ b/ccmlib/scylla_repository.py
@@ -338,6 +338,10 @@ def setup(version, verbose=True, skip_downloads=False):
     if skip_downloads:
         return directory_name(version), packages
 
+    # Packages may also come from environment variables (e.g. SCYLLA_UNIFIED_PACKAGE).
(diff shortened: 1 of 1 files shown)


avikivity:
(P0 label was added by mistake)

fruch (approved):
LGTM

avikivity:
@fruch please consider merging

fruch:
> @fruch please consider merging

I thought we are not looking for reason to break the code freeze :)

fruch:
> > @fruch please consider merging
> 
> I thought we are not looking for reason to break the code freeze :)

Still need to pull it in dtest

PR #740, opened 2026-06-24
fix(test): handle long tmp_path in test_short_path_returns_unchanged

mykaul:
In CI environments (e.g. Nix), `tmp_path` can be long enough that even a \short\ test path exceeds `UNIX_SOCKET_PATH_MAX` (107 on Linux), causing the symlink logic to kick in and the assertion to fail.  Fall back to a `tempfile.mkdtemp()` dir when `tmp_path` is too long.


fruch:
https://github.com/scylladb/scylla-ccm/pull/738 is already covering it

PR #779, opened 2026-08-19
check_socket_available: retry bind before giving up

yaronkaikov:
A bind failure right after another worker's cluster teardown often just
means that worker's scylla process hasn't fully released the port yet
(kernel hasn't reclaimed it), not that the address is genuinely stuck in
use. With many parallel dtest workers on a single node/container, a
worker can be handed a just-freed loopback address before the previous
occupant's process has actually released the port, causing an immediate
and permanent UnavailableSocketError.

Retry the bind briefly with backoff (5s, 0.5s steps) using the existing
wait_for() helper before giving up, instead of failing on the first
attempt.

Fixes: RELENG-819

### DIFF
--- a/ccmlib/common.py
+++ b/ccmlib/common.py
@@ -2,6 +2,7 @@
 # Cassandra Cluster Management lib
 #
 
+import errno
 import fnmatch
 import fcntl
 import os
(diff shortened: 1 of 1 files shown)


yaronkaikov:
@roydahan @kreuzerkrieg seems like those failure are not related to my change, please review and merge

ccmlib/common.py
```diff
@@ -766,35 +766,46 @@ def pids_listening_on(addr: str, port: int) -> List[int]:
...
+        for pid in pids:
+            p = psutil.Process(pid)
+            secs_ago = p.create_time()
+            fmt = '%H:%M:%S' if secs_ago < 60 * 60 * 24 else '%Y-%m-%d %H:%M:%S'
+            since = datetime.fromtimestamp(secs_ago).strftime(fmt)
+            logger.error('Address %s:%d used by: "%s" '
+                         '(pid: %d, since: %s, parent: "%s")',
+                         addr, port, ' '.join(p.cmdline()),
+                         p.pid, since, ' '.join(p.parent().cmdline()))
+    raise UnavailableSocketError(f'Inet address {addr}:{port} is not available after {timeout}s of retries')
```
kreuzerkrieg: you are loosing the error message here, maybe you want to return tuple from _try_bind

ccmlib/common.py
```diff
@@ -766,35 +766,46 @@ def pids_listening_on(addr: str, port: int) -> List[int]:
...
-                             p.pid, since, ' '.join(p.parent().cmdline()))
-        raise UnavailableSocketError(f'Inet address {addr}:{port} is not available: {msg}')
+    def _try_bind():
+        s = socket.socket(family, socktype)
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        try:
+            s.bind(sockaddr)
+            return True
+        except socket.error:
+            return False
```
kreuzerkrieg, line 790: you are going to return False and keep retrying on every error, even to retryable
maybe something like `except OSError as e: with if e.errno != errno.EADDRINUSE: raise`?

ccmlib/common.py
```diff
@@ -766,35 +766,46 @@ def pids_listening_on(addr: str, port: int) -> List[int]:
...
+    # mid-teardown and the kernel hasn't reclaimed its port yet, not that the
+    # address is genuinely stuck in use. Retry briefly with backoff before
+    # giving up, instead of failing outright on the first attempt.
+    if wait_for(_try_bind, timeout=timeout, step=step):
+        return
+
+    addr, port = itf
+    pids = pids_listening_on(addr, port)
+    if not pids:
+        logger.error('Address %s:%d used unknown process')
```
kreuzerkrieg: what are these %s and %d?

ccmlib/common.py
yaronkaikov: Good catch — reworked to capture the actual OSError as `last_error` and include it in the final `UnavailableSocketError` message instead of dropping it.

ccmlib/common.py
yaronkaikov: Right — fixed. `_try_bind` now catches `OSError` and re-raises immediately when `e.errno != errno.EADDRINUSE`, so only the actual retryable condition gets retried; anything else propagates out of `wait_for` on the first attempt.

ccmlib/common.py
yaronkaikov: They were leftover placeholders with no args passed (would have thrown a TypeError if that branch were ever hit). Fixed to `logger.error('Address %s:%d used by unknown process', addr, port)`.

kreuzerkrieg (approved):

yaronkaikov:
@kreuzerkrieg Can we merge this , who can update dtest submodule ?

kreuzerkrieg:
> @kreuzerkrieg Can we merge this , who can update dtest submodule ?

no idea how it works, but I can assume the same way as seastar, merge this, open another PR which updates the submodule for scylladb. Try to ping any of maintainers for more info

roydahan (approved):

yaronkaikov:
Re-checked the failing checks after the amended push — still unrelated to this change:

- `build (3.11/3.12/3.13/3.14)` and `(using Nix)`: all fail identically in `test_sctool[ccm_reloc_with_manager_cluster]` with `botocore.exceptions.ClientError: 403 Forbidden` on an S3 `HeadObject` call during test **setup** — the relocatable/manager package can't be fetched. This never reaches `check_socket_available()`.
- `Docker Integration Tests`: fails with `subprocess.TimeoutExpired: podman version ... timed out after 5 seconds` — podman unresponsive on the runner, unrelated to Python/`ccmlib.common`.

Same S3 403 signature also shows up on an unrelated branch/PR (`fix-777-manager-test-latest`) in this same CI workflow, so this looks like pre-existing runner-level flakiness (S3 access + podman availability) rather than something introduced here.

PR #780, opened 2026-08-19
tests: stop relying on ephemeral unstable-build S3 snapshots

pehala:
## Summary
CI integration tests were failing on every PR. Root cause: several tests pin specific historical S3 build snapshots (unstable master/enterprise dates, a manager branch) that get pruned by S3's retention lifecycle over time, so the objects eventually stop existing.

- `test_setup_unstable_master_new_url` / `test_setup_unstable_enterprise_new_url`: mock the S3 bucket listing and `00-Build.txt` manifest fetch instead of depending on a live historical build, since these tests only verify URL-construction logic.
- `test_setup_unified_package_reinstall`: swapped the hardcoded 2025-01 master snapshot for a currently valid one and re-verified the resulting version string against a real download.
- `TestGetManagerFunctions::test_get_manager_latest_reloc_url`: `branch-3.1` no longer exists in the manager unstable bucket (only `branch-3.11`/`branch-3.12` remain); updated to `branch-3.12`.

The reloc-manager hardcoded-URL fix (`test_scylla_cmds.py`) was already merged upstream in 7f35180, so this PR only contains the remaining fixes.

## Test plan
- [x] `pytest tests/test_scylla_repository.py -m "not docker and not reloc"` passes locally
- [x] Full `pytest ./tests -m "not docker"` passes locally (aside from cases requiring Java 8, which CI provisions but my local env doesn't)

### DIFF
--- a/tests/test_scylla_repository.py
+++ b/tests/test_scylla_repository.py
@@ -5,7 +5,7 @@ import tempfile
 import time
 import typing
 from pathlib import Path
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


pehala:
Podman failed on a timeout for some reason, but the actual test failures are fixed here

fruch (approved):

PR #784, opened 2026-09-09
ccmlib/node.py: adapt Node.data() to base-2 sizes

denesb:
Nodetool divides sizes by 1024 but uses base-10 units to represents these in command outputs (KB, MB, etc.). This will be corrected soon. Adapt Node.data() so it can parse both the old and the new corrected output.
This code is unused in dtest, but it is public API so worth fixing to avoid unpleasant surprises down the line.

Refs: scylladb/scylladb#31183


fruch (approved):
LGTM

denesb:
Thanks @fruch, who can merge this?

